PR #34, opened 2021-10-12
indexers: Fix BTN IRC Server

KyleSanderson:


ludviglundgren:
Hi! 

Thanks for your PR :) 

Do note that BTN (and some others) is not currently working because it's using multiple lines for announcements which there isn't yet support for. But it is on the todo list, and will try to get that worked out soon.

KyleSanderson:
Good to know. I noticed reconnecting is also not implemented (or if it fails to get into the channel - retrying) so I setup sonarr alongside.

PR #43, opened 2021-12-21
Indexer: New definition Milkie

GregTroar:
adding MK definition which has been tested and is working


internal/indexer/definitions/milkie.yaml
```diff
@@ -0,0 +1,58 @@
...
+supports:
+  - irc
+  - rss
+source: custom
+settings:
+  - name: apikey
+    type: secret
+    label: Api key
+    help: "Go to your profile and copy your Api key, make sure to replace <+> by <%2B>"
+    regex: /([a-zA-Z0-9\+])
```
ludviglundgren, line 20: This seem to break at least syntax highlight, but did it load fine when you tested? How does the key actually look?

The regex for settings isn't yet used, but could could be useful.

internal/indexer/definitions/milkie.yaml
```diff
@@ -0,0 +1,58 @@
...
+urls:
+  - https://www.milkie.cc
+privacy: private
+protocol: torrent
+supports:
+  - irc
+  - rss
+source: custom
+settings:
+  - name: apikey
```
ludviglundgren, line 16: Are you're sure it's an api key and not passkey? Doesn't matter more than it could be confusing for the user to get it if they differ.

internal/indexer/definitions/milkie.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+irc:
+  network: p2p-network.net
+  server: irc.p2p-network.net
+  port: 6697
+  tls: true
+  channels:
+    - "#milkie-announce"
+  announcers:
+    - the_cow
```
ludviglundgren, line 30: In the `autodl-irssi` .tracker file there's multiple announcers - are any of those used as well or actually only this one?

internal/indexer/definitions/milkie.yaml
```diff
@@ -0,0 +1,58 @@
...
+  type: single
+  lines:
+    - test:
+        - "{tv} :: The.Great.Xmas.Show.S00E00.1080p.WEB.H264-TEST :: [1080p, web] :: https://milkie.cc/browse/xxxxxxxxxxxx"
+      pattern: '{(.*)}\s::\s(.*)\s::\s.*\s::\s(https?\:\/\/[^\/]+\/)browse\/(.*)'
+      vars:
+        - category
+        - torrentName
+        - baseUrl
+        - torrentId
```
ludviglundgren, line 53: No freeleech flags etc?

ludviglundgren (changes requested):
Hi! 

Thanks for you first contribution :)

Have some minor things to note, but looks overall fine.

internal/indexer/definitions/milkie.yaml
GregTroar: I tested this one completely and it worked well yes.
The api key looks like this: Cpi1Xt9fKYGtTln+6ilkuzGsWiJ7cSHq, this one is a fake one but same patent as mine.
Regex not updated in other files for the time being

internal/indexer/definitions/milkie.yaml
GregTroar: sure 100% I mean this is what is explicitly mentioned in the indexer page.
In the past it seems it was called Download key but is now showing apikey.

internal/indexer/definitions/milkie.yaml
GregTroar: the_cow1 and the_cow2 are not in the irc chan nope, I checked the .tracker file anyway already.

internal/indexer/definitions/milkie.yaml
GregTroar: Could not find any for the time being unfortunately (checked around 1,500 torrents)

internal/indexer/definitions/milkie.yaml
ludviglundgren: I see, so it's thats only to replace it for an url encoded value. Thats fine with the regex, can update that later on.

internal/indexer/definitions/milkie.yaml
ludviglundgren: Great, thank you for double checking that.

internal/indexer/definitions/milkie.yaml
ludviglundgren: Then it's probably not used. Thanks for checking!

PR #57, opened 2021-12-30
Indexers: Update TorrentSeeds for new site

GregTroar:
Regex was not working with wrong URL.
Updated it and tested, it is working


ludviglundgren (approved):
🚀

PR #58, opened 2021-12-31
Indexer: Update MTV

GregTroar:
Dunno why so many commits not verified, I presume I cannot just commit only one ? it seems i cannot chose only one commit to be a PR.

Anyway NickServ pass is required for MTV


ludviglundgren:
You should make a new feature branch from your fork's develop, so not everything in your develop gets pushed in a PR. 

Can you fix that? Let me know if I should fix it or scrap the pr.

GregTroar:
Scrap the PR I will try to see how to get only this update

Issue #59, opened 2022-01-01
Cannot add IRC / Clients (403 error)


GitHubMilo:
This is setup using the downloaded binary, and installed as a service on Linux.

I keep hitting these 403 errors when using the service and trying either add an IRC network, or a client (like radarr / sonarr).

Any suggestions?

```
    at APIClient.ts:32
    at s (runtime.js:63)
    at Generator._invoke (runtime.js:294)
    at Generator.next (runtime.js:119)
    at r (asyncToGenerator.js:3)
    at u (asyncToGenerator.js:25)
    at asyncToGenerator.js:32
    at new Promise (<anonymous>)
    at asyncToGenerator.js:21
    at APIClient.ts:21
```

DEBUG logs don't appear to provide anything useful for this.

GregTroar:
Could you try to update your config.toml with TRACE log instead and relaunch ??

GitHubMilo:
> Could you try to update your config.toml with TRACE log instead and relaunch ??

Unfortunately still nothing more in the log.
This is all I get with "TRACE" on:
`{"level":"debug","time":"2022-01-01T18:37:48Z","message":"Starting autobrr"}
{"level":"debug","time":"2022-01-01T18:37:48Z","message":"Log-level: TRACE"}
{"level":"info","time":"2022-01-01T18:37:48Z","message":"Starting server. Listening on 10.20.20.140:9999"}
{"level":"trace","time":"2022-01-01T18:37:48Z","message":"parsing: definitions/alpharatio.yaml"}
{"level":"trace","time":"2022-01-01T18:37:48Z","message":"parsing: definitions/beyondhd.yaml"}
{"level":"trace","time":"2022-01-01T18:37:48Z","message":"parsing: definitions/btn.yaml"}
{"level":"trace","time":"2022-01-01T18:37:48Z","message":"parsing: definitions/emp.yaml"}
{"level":"trace","time":"2022-01-01T18:37:48Z","message":"parsing: definitions/filelist.yaml"}
{"level":"trace","time":"2022-01-01T18:37:48Z","message":"parsing: definitions/gazellegames.yaml"}
{"level":"trace","time":"2022-01-01T18:37:48Z","message":"parsing: definitions/hd-torrents.yaml"}
{"level":"trace","time":"2022-01-01T18:37:48Z","message":"parsing: definitions/iptorrents.yaml"}
[...]

PR #67, opened 2022-01-06
Feature: Improve config for http server

ludviglundgren:
Simplify config for http server and make some changes to how session cookies are set.

It will now check if the app is behind a reverse-proxy with a simple check for the header `X-Forwarded-Proto` which should be available if behind a RP like nginx etc.

If this header is equal to `https` it will set `Secure` and `SameSite = Strict` for the cookie. If not then skip that and `SameSite = Lax` will be used, as per the spec https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite

This closes issue #59


brettpetch:
Review: maybe disable secure cookies for those hosting on port. non ssl users can enable it via proxy_cookie_flags argument.

PR #97, opened 2022-01-23
Indexers: Add TorrentDb

GregTroar:
Have been testing and working both with Freeleech and without.


ludviglundgren (approved):
Did some minor fixes, changed `size` to `torrentSize` and added to supported indexers in README.

GregTroar:
Well done

PR #105, opened 2022-02-03
Indexer: Add DigitalCore

GregTroar:
tested and working

### DIFF
--- a/internal/indexer/definitions/digitalcore.yaml
+++ b/internal/indexer/definitions/digitalcore.yaml
@@ -1,0 +1,61 @@
+---
+#id: digitalcore
+name: DigitalCore
+identifier: digitalcore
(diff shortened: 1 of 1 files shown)


internal/indexer/definitions/digitalcore.yaml
```diff
@@ -0,0 +1,61 @@
...
+  port: 7000
+  tls: true
+  channels:
+    - "#announce"
+  announcers:
+    - ENDOR
+  settings:
+    - name: nickserv.account
+      type: text
+      required: false
```
ludviglundgren, line 33: Change to `true` to require a username, even if it's not necessarily a nickserv account.

ludviglundgren (approved):
Looks good! Only one minor thing :rocket:

Issue #119, opened 2022-02-06
Error processing deluge action


zmiguel:
Hi all, I set this up today and I'm getting an error when it grabs a release from IRC and tries to send it to Deluge v2

My setup:
* Docker container for Deluge V2
* Docker container for autobrr
* Deluge V2.0.5
* autobrr: ghcr.io/autobrr/autobrr:latest (v0.13.0)

Error log (redacted):
```
2022-02-06T18:22:01Z DEBUG irc.site.org: #announces _Bot_: New Torrent: <TV> Name:'torrent_name' uploaded by 'Anonymous' freeleech - https://www.some_website.com/torrent/99999999
2022-02-06T18:22:01Z DEBUG filter-service.find_and_check_filters: found and matched filter: site Freeleech
2022-02-06T18:22:01Z DEBUG filter-service.find_and_check_filters: (site Freeleech) additional size check required
2022-02-06T18:22:02Z DEBUG successfully downloaded file: /tmp/autobrr-772268386
2022-02-06T18:22:02Z INFO Matched 'torrent_name' (site Freeleech) for some_website
2022-02-06T18:22:02Z DEBUG process action: Deluge Download for 'torrent_name'
2022-02-06T18:22:02Z DEBUG action Deluge: Deluge Download
2022-02-06T18:22:02Z ERROR could not read torrent file:
2022-02-06T18:22:02Z ERROR error sending torrent to Deluge
2022-02-06T18:22:02Z ERROR process action failed: Deluge Download for 'torrent_name'
```

The containers don't share any folders between them, do they need to?

Any help would be much appreciated,
Thanks!

ludviglundgren:
Hi! What image are you using for Deluge? We had another user report this as well, and he solved it by doing this:

---

### Autobrr not connecting to Deluge (linuxserver/deluge)
Solution:
### Host
Host should be the deluge container you're trying to reach, for my setup it was simply `deluge`.
Make sure that your docker containers are on the same network, so they can reach each other. If you're using a single compose file, it should be by default. Make sure `container_name` is set as well.

### Port
This is the daemon port, not the webui port. Find the correct one by logging into deluge webui, and checking `preferences` or under `connection manager` (default: 58846).

### Authentication 
The Username/Password required is for the daemon, not the webui.
It can be found at `/docker/appdata/deluge/auth`, the default one looks like: `localclient:password:10`
You can add your own if you wish, as, `username:password:powerlevel`

[...]

PR #121, opened 2022-02-07
fix: empty tmpFile var passed to download client's action

zmiguel:
Related to #119

After some debugging I figured out that when `release.TorrentTmpFile` was not empty the `tmpFile` variable would not be set. 
Resulting in sending an empty variable to the download client's action function. That ultimately resulted in a file not found error when trying to send the torrent file to the download client.

Fixed by adding an else condition setting `tmpFile` to `release.TorrentTmpFile` (when it is not empty).

Even though this only happened to me on deluge, I added the else condition to all instances where this could possibly happen.

---

edit: I've tested it by building the container and deluge is now correctly getting the torrent added

### DIFF
--- a/internal/action/run.go
+++ b/internal/action/run.go
@@ -64,6 +64,8 @@ func (s *service) runAction(action domain.Action, release domain.Release) error 
 			}
 
 			tmpFile = t.TmpFileName
+		} else {
+			tmpFile = release.TorrentTmpFile
 		}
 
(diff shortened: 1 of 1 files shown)


ludviglundgren (approved):
Thank you for debugging and fixing this yourself! I haven't seen this issue before but always good to do an additional check so I'll merge this :)

Fixes #119

PR #141, opened 2022-02-17
fix(Base): Added missing transition-colors property

stacksmash76:
fix(Base): Added missing transition-colors property to the user dropdown pill in the header component.


ludviglundgren (approved):
:+1:

Issue #158, opened 2022-03-01
IRC Network cannot be edited if no channels exist


SavageCore:
Upon clicking on an IRC network with no channels you receive the error:

```
react-dom.production.min.js:216 TypeError: Cannot read properties of null (reading 'length')
    at children (IrcForms.tsx:25:52)
    at n.r.render (FieldArray.tsx:367:12)
    at Qi (react-dom.production.min.js:187:188)
    at Wi (react-dom.production.min.js:186:173)
    at Zs (react-dom.production.min.js:269:427)
    at Pl (react-dom.production.min.js:250:347)
    at Rl (react-dom.production.min.js:250:278)
    at Nl (react-dom.production.min.js:250:138)
    at bl (react-dom.production.min.js:243:163)
    at react-dom.production.min.js:123:115
cs @ react-dom.production.min.js:216
n.callback @ react-dom.production.min.js:216
ho @ react-dom.production.min.js:131
gs @ react-dom.production.min.js:220
Al @ react-dom.production.min.js:259
t.unstable_runWithPriority @ scheduler.production.min.js:18
Qa @ react-dom.production.min.js:122
Il @ react-dom.production.min.js:252
bl @ react-dom.production.min.js:243
(anonymous) @ react-dom.production.min.js:123
t.unstable_runWithPriority @ scheduler.production.min.js:18
Qa @ react-dom.production.min.js:122
Ka @ react-dom.production.min.js:123
Za @ react-dom.production.min.js:122
Ll @ react-dom.production.min.js:263
t.unstable_runWithPriority @ scheduler.production.min.js:18
Qa @ react-dom.production.min.js:122
Fl @ react-dom.production.min.js:261
(anonymous) @ react-dom.production.min.js:261
D @ scheduler.production.min.js:16
x.port1.onmessage @ scheduler.production.min.js:12
IrcForms.tsx:25 Uncaught TypeError: Cannot read properties of null (reading 'length')
    at children (IrcForms.tsx:25:52)
    at n.r.render (FieldArray.tsx:367:12)
    at Qi (react-dom.production.min.js:187:188)
    at Wi (react-dom.production.min.js:186:173)
    at Zs (react-dom.production.min.js:269:427)
    at Pl (react-dom.production.min.js:250:347)
    at Rl (react-dom.production.min.js:250:278)
    at Nl (react-dom.production.min.js:250:138)
    at bl (react-dom.production.min.js:243:163)
    at react-dom.production.min.js:123:115
[...]

PR #159, opened 2022-03-01
Refactor: release table-related code and fix for #158

stacksmash76:
refactor: moved duplicate dashboard/release code to a separate folder: components/data-table. [most of the PR]
refactor(APIClient): updated the newly added findQuery function to use URLSearchParams instead of manually crafting the URI string itself.
refactor(SlideOver): added proper typings to the SlideOver component and added a sanity check to prevent passing of null/undefined values to the child component before rendering.
refactor: removed the redundant Network and Channel typings and updated relevant typings to match the backend. adapted relevant code to match these changes.
fix(ChannelsFieldArray): fixed a bug where it was unable to add a new irc network due to the validation object being initialized as non-empty (formik requires that successful validated entries return empty objects)
refactor(screens/settings/Irc): replaced incorrect typings, sanitized potentially null values and cleaned up the surrounding code.
fix: included changes should fix issue #158 as well.

PR passes linting, building and also manual testing.


web/src/api/APIClient.ts
```diff
@@ -87,38 +87,35 @@ export const APIClient = {
...
+                    return;
+
+                if (filter.id == "indexer")
+                    params.append("indexer", filter.value);
+                else if (filter.id === "action_status")
+                    params.append("push_status", filter.value);
+            });
 
-            return appClient.Get<ReleaseFindResponse>(`api/release${queryString}`)
+            return appClient.Get<ReleaseFindResponse>(`api/release?${params.toString()}`)
```
ludviglundgren, line 118: This is of course the correct way and much better! :smile:

ludviglundgren (approved):
Great improvements :+1: 

Changed this:
* api: initialize an empty array for channels to not send `null`.
* web: check in the irc network list if channels exist, if not add empty message

PR #166, opened 2022-03-05
feat(filters): add priority

ludviglundgren:
Add support for priority on filters.

When making the filter check it will fetch filters and order and run them by `priority`, highest to lowest.

Supports positive and negative numbers. Default `priority` is set to `0`.


KyleSanderson:
Can we get a "continue" flag on this one - to keep running if it's successful?

PR #173, opened 2022-03-09
docker: Add curl to base image

KyleSanderson:
cURL is a very useful utility to have in the image - I'm using it with exec to notify other applications of the event.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,7 +42,7 @@ ENV HOME="/config" \
 XDG_CONFIG_HOME="/config" \
 XDG_DATA_HOME="/config"
 
-RUN apk --no-cache add ca-certificates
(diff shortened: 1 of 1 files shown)


ludviglundgren:
What do you use for notifications? It's being worked on but not ready yet, so would be good to know if it's something that can be supported natively. Do you have actions that run notifications on the filters?

Was thinking about adding a http action as well, so this can be done in-app.

KyleSanderson:
Webhooks primarily. The one that makes this a bit more interesting is cross-seed.

ludviglundgren:
Okay! I'll merge this now and probably make a release later this week, but it will build the develop image on merge.

And lets see what we can do in the future with a proper webhook action :slightly_smiling_face:

PR #178, opened 2022-03-16
Indexer: Adding Torrent-Syndikat

F4n4t:
Tested it for a while. Should work now.

I used origin to mark the release source. (P2P, SCENE, O-SCENE)
Scene is for unpacked Scene Releases and O-Scene for untouched Scene Releases.


ludviglundgren (approved):
Hello! Thanks for the PR 🙂 

Currently there's no way to filter for origin, but I'm planning on adding support for it in the near future.

PR #184, opened 2022-03-19
feat(actions): add webhook support

### LINKED ISSUE #171: exec expansion splits args based on space after variable expansion
Discussed on discord - just a quick summary here because I'm likely going to forget.

brr Exec currently does not support spaces within quotes, and passes them as individual args. Because of the ordering of the macro expansion https://github.com/autobrr/autobrr/blob/5a4585167726d49fd6f8ffc6d6c2fd474369c343/internal/action/exec.go#L26

into the arg string split https://github.com/autobrr/autobrr/blob/5a4585167726d49fd6f8ffc6d6c2fd474369c343/internal/action/exec.go#L33 

it is impossible to pass arguments and macros containing a space to programs. Invoking a script is a current workaround, because $SHELL can repair the damage by bringing the arg string together again by wrapping in quotes. Changing this ordering can help, but the implementation will still be incorrect as expected spaces (even if escaped) will be split.

STR: "{{ .TorrentName }}" > `Sally Goes to the Mall S04E29` > curl -XPOST

There's a couple ways to tackle this, an out-of-tree one being https://github.com/mattn/go-shellwords which could add the ability to have environment variables as well on execution.

ludviglundgren:
Add support for webhook action. This will allow for basic webhooks for now to integrate external programs easier like [mmgoodnow/cross-seed](https://github.com/mmgoodnow/cross-seed).

Currently only `POST` and json data.

Future improvements:
* set method (GET, POST, PUT, PATCH, DELETE)
* set data type (URL encoded, JSON)
* set headers

Resolves #171

### DIFF
--- a/internal/action/run.go
+++ b/internal/action/run.go
@@ -270,17 +270,18 @@ func (s *service) webhook(action domain.Action, release domain.Release) {
 	m := NewMacro(release)
 
 	// parse and replace values in argument string before continuing
-	dataArgs, err := m.Parse(action.Data)
+	dataArgs, err := m.Parse(action.WebhookData)
 	if err != nil {
-		log.Error().Stack().Err(err).Msgf("could not parse macro: %v", action.Data)
+		log.Error().Stack().Err(err).Msgf("could not parse macro: %v", action.WebhookData)
+		return
 	}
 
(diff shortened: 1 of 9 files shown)


jorge123255:
I wanted to know if its possible with API keys to post a download to offcloud.com[ ](https://github.com/Offcloud/offcloud-api and if so how i cant find the post method

PR #185, opened 2022-03-20
qBittorrent: add resume, recheck, and forceStart

KyleSanderson:
I didn't realize you already put in all this work to create golang bindings for qBittorrent - this alone is immense value (beyond thankful for you sharing all this). I've added three methods that I'm interested in to fix stalls, and other nuances.

Outside of these additions, there was a typo on forceDL which breaks the stable API. Based on the fact this didn't work with TorrentState, the hope is there were no consumers of this.

### DIFF
--- a/pkg/qbittorrent/domain.go
+++ b/pkg/qbittorrent/domain.go
@@ -109,7 +109,7 @@ const (
 	TorrentStateCheckingDl TorrentState = "checkingDL"
 
 	// Torrent is forced to downloading to ignore queue limit
-	TorrentStateForceDl TorrentState = "forceDL"
+	TorrentStateForcedDl TorrentState = "forcedDL"
 
 	// Checking resume data on qBt startup
 	TorrentStateCheckingResumeData TorrentState = "checkingResumeData"
--- a/pkg/qbittorrent/go.mod
+++ b/pkg/qbittorrent/go.mod
@@ -1,0 +1,2 @@
+module qbittorrent
+go 1.17
--- a/pkg/qbittorrent/methods.go
+++ b/pkg/qbittorrent/methods.go
@@ -286,3 +286,74 @@ func (c *Client) GetTransferInfo() (*TransferInfo, error) {
 
 	return &info, nil
 }
+
+func (c *Client) Resume(hashes []string) error {
+	v := url.Values{}
+
+	// Add hashes together with | separator
+	hv := strings.Join(hashes, "|")
+	v.Add("hashes", hv)
+
+	encodedHashes := v.Encode()
+
+	resp, err := c.get("torrents/resume?"+encodedHashes, nil)
+	if err != nil {
+		log.Error().Err(err).Msgf("resume error: %v", hashes)
+		return err
+	} else if resp.StatusCode != http.StatusOK {
+		log.Error().Err(err).Msgf("resume error bad status: %v", hashes)
+		return err
+	}
+
+	defer resp.Body.Close()
+
+	return nil
+}
+
+func (c *Client) SetForceStart(hashes []string, value bool) error {
+	v := url.Values{}
+
+	// Add hashes together with | separator
+	hv := strings.Join(hashes, "|")
+	v.Add("hashes", hv)
+	v.Add("value", strconv.FormatBool(value))
+
+	encodedHashes := v.Encode()
+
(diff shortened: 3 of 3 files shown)


KyleSanderson:
It's an aside, but this is the code I have.

```

func fixStalled(c *qbittorrent.Client) {
	torrents, err := c.GetTorrentsFilter(qbittorrent.TorrentFilterStalledDownloading)
	if err != nil {
		fmt.Printf("Bad filter: %q\n", err)
		return
	}

	hashes := make([]string, 0, len(torrents))
	for _, t := range torrents {
		if t.TotalSize == 0 {
			continue
		}

		if (float64(t.AmountLeft) / float64(t.TotalSize)) > 0.1 {
			fmt.Printf("Skipping %q: %f\n", t.Name, (float64(t.AmountLeft) / float64(t.TotalSize)))
			continue
		}

		fmt.Printf("Stalled Torrent: %q\n", t.Name)
		hashes = append(hashes, t.Hash)
	}

	if err = c.SetForceStart(hashes, true); err != nil {
		fmt.Printf("Bad forcestart: %q\n", err)
		return
	}
	
	if err = c.Recheck(hashes); err != nil {
		fmt.Printf("Bad recheck: %q\n", err)
		return
	}
}

func fixForcedDown(c *qbittorrent.Client) {
	torrents, err := c.GetTorrentsFilter(qbittorrent.TorrentFilterDownloading)
	if err != nil {
		fmt.Printf("Bad filter: %q\n", err)
		return
	}

	hashes := make([]string, 0, len(torrents))
	for _, t := range torrents {
		if t.State != qbittorrent.TorrentStateForcedDl {
			continue
		}

		fmt.Printf("Forced Down Torrent: %q\n", t.Name)
		hashes = append(hashes, t.Hash)
	}

	if err = c.Resume(hashes); err != nil {
		fmt.Printf("Bad resume: %q\n", err)
		return
	}
}

func fixForcedUp(c *qbittorrent.Client) {
	torrents, err := c.GetTorrentsFilter(qbittorrent.TorrentFilterUploading)
	if err != nil {
		fmt.Printf("Bad filter: %q\n", err)
		return
	}

	hashes := make([]string, 0, len(torrents))
	for _, t := range torrents {
		if t.State != qbittorrent.TorrentStateForcedUp {
			continue
		}

		fmt.Printf("Forced Torrent: %q\n", t.Name)
		hashes = append(hashes, t.Hash)
	}

	if err = c.Resume(hashes); err != nil {
		fmt.Printf("Bad resume: %q\n", err)
		return
	}
}

func fixErrored(c *qbittorrent.Client) {
	torrents, err := c.GetTorrentsFilter(qbittorrent.TorrentFilterInactive)
	if err != nil {
		fmt.Printf("Bad filter: %q\n", err)
		return
	}

	hashes := make([]string, 0, len(torrents))
	for _, t := range torrents {
		if t.State != qbittorrent.TorrentStateError {
			continue
		}

		fmt.Printf("Errored Torrent: %q\n", t.Name)
		hashes = append(hashes, t.Hash)
	}

	if err = c.SetForceStart(hashes, true); err != nil {
		fmt.Printf("Bad resume: %q\n", err)
		return
	}

	if err = c.Resume(hashes); err != nil {
		fmt.Printf("Bad resume: %q\n", err)
		return
	}
}
```

KyleSanderson:
added Pause, ForceTMM, and Category functions.

ludviglundgren:
Are these to be used internally or for external projects? There's some more methods implemented in my other tool [qbittorrent-cli/pkg/qbittorrent/methods.go](https://github.com/ludviglundgren/qbittorrent-cli/blob/develop/pkg/qbittorrent/methods.go) that might be of interest as well if it's for external use.

The `go.mod` file is not needed since it's using the root project one, and you can simply use the import `github.com/autobrr/autobrr/pkg/qbittorrent` in external projects.

KyleSanderson:
This is for external usage.

There's a decently significant bug with the current tool, where the hard coded address is assumed to be localhost:8080. https://github.com/ludviglundgren/qbittorrent-cli/blob/3c4c6b4ea38fd7c434c3fd56f976d9368c9daf69/pkg/qbittorrent/methods.go#L29

When I tried to use it with another host ages ago, things did not work well.

The bad script I have is here: https://github.com/KyleSanderson/qbittorrent-bugfixer/blob/main/main.go . It's a shame to exist, but seems to fix the decade long issues that have plagued the client.

PR #186, opened 2022-03-20
feat(indexers): load custom definitions

ludviglundgren:
Add support to load custom definitions from location on disk. Makes local development easier.


GregTroar:
So cool

PR #189, opened 2022-03-21
fix(actions): webhook double escaping with json.marshal

KyleSanderson:
json.marshal escapes quotes unfortunately, which generates invalid json ironically.


KyleSanderson:
This fellow describes the exact issue here (it's from 2016 - so there may be a better way now 6 years later): https://goinbigdata.com/how-to-correctly-serialize-json-string-in-golang/

The proposed PR should match my setup presently, which now works.

PR #203, opened 2022-03-25
feat(filters): add origin filter

F4n4t:
I've made a few adjustments to filter releases by origin.
First i used checkFilterStrings, but I think it would make more sense to get only unique matches for the origin filter.
e.g. a tracker where i use i filtered for "SCENE" which is the origin for unpacked scene, but they also use O-SCENE for untouched scene releases. So i made a new function to check for exact and wildcard matches.

### DIFF
--- a/internal/database/filter.go
+++ b/internal/database/filter.go
@@ -58,17 +58,17 @@ func (r *FilterRepo) FindByID(ctx context.Context, filterID int) (*domain.Filter
 	//r.db.lock.RLock()
 	//defer r.db.lock.RUnlock()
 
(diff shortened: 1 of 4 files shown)


ludviglundgren:
Thanks for the PR! 

Looks good, but there's some limitations to this solution even if it's flexible with wildcard matching etc.

My own plan for this was to add mapping for the definitions/vars to a set of variables that could be universally used like `scene, p2p, internal`. And then possibly more variables to cover o-scene like maybe `scene-unpacked` or something even if they are the only one using it.

I'll let this sit for a few days and get back to it. :slightly_smiling_face:

F4n4t:
Of course, that would be even better and much easier to set up. :)

ludviglundgren:
This has been fixed in https://github.com/autobrr/autobrr/pull/257 

Available values: `SCENE`, `O-SCENE`, `P2P` and `Internal` :smiley: 

The Filter origins is now an array so you'll unfortunately have to update your db with these changes somehow.

PR #208, opened 2022-03-27
Indexers add tnetwork

F4n4t:
Added another german tracker.


ludviglundgren (approved):
Nice! Thanks for your work :smiley: 

And great that you updated the README as well!

PR #223, opened 2022-04-08
feat: add ability to create an account via the webui

stacksmash76:
This PR makes it easier for new users to create an account without the need to use autobrrctl.


ludviglundgren (approved):
Nicely done! :rocket:

PR #240, opened 2022-04-12
feat(http): add healthcheck endpoints liveness and readiness

ludviglundgren:
Add healthcheck endpoints - liveness and readiness.

Useful for container deployments.

New endpoints. Both answers with appropiate status codes and body text as `text/plain`

- `/api/healtz/liveness` - 200 OK and OK in body
- `/api/healtz/readiness` - 200 OK and OK in body if databse ping is ok. If not, 500 and `Unhealthy. Database unreachable`

### DIFF
--- a/cmd/autobrr/main.go
+++ b/cmd/autobrr/main.go
@@ -100,7 +100,7 @@ func main() {
 	errorChannel := make(chan error)
 
 	go func() {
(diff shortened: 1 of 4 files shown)


KyleSanderson:
`api/healthz/readiness` and `api/healthz/liveness` :P the typo got me on the desc.

ludviglundgren:
Didn't even see it :D But the code is right so will edit description!

PR #245, opened 2022-04-13
feat(qbit): refactor url parse and add basic auth

ludviglundgren:
Rework url/port handling for qbit to be easier and more flexible to use.

Now supports most variants of urls like:

- `https://qbit.domain.me`
- `https://qbit.domain.me/qbittorrent`
- `domain.me:10200`
- `http://domain.me:10200`

Also added proper support for Basic Auth so no need to put it in the host field anymore.


pkg/qbittorrent/client.go
```diff
@@ -242,3 +263,50 @@ func (c *Client) setCookies(cookies []*http.Cookie) {
...
+
+func buildUrl(settings Settings, endpoint string) string {
+	// parse url
+	u, _ := url.Parse(settings.Hostname)
+
+	// reset Opaque
+	u.Opaque = ""
+
+	// set scheme
+	scheme := "http"
```
KyleSanderson, line 275: What's going on here?

ludviglundgren: It's all unnecessarily complex because it needs to support old setups with port in the separate field, as well as only using the `host`. There's some testcases that cover most of it and with one of them like `qbit.domain.ltd` it would parse it incorrectly and set `u.Opaque` to http or qbit which was a bit weird.

The background of the change is I got some reports of it being broken when using qbit behind the reverse proxy Traefik.

KyleSanderson: I had an issue I think with sonarr not working on port 80. Unrelated to this issue, but still something

KyleSanderson: The branch below doesn't make sense because it's the same code

pkg/qbittorrent/client.go
```diff
@@ -242,3 +263,50 @@ func (c *Client) setCookies(cookies []*http.Cookie) {
...
+	}
+
+	// reset Path
+	if u.Host == u.Path {
+		u.Path = ""
+	}
+
+	// handle ports
+	if settings.Port > 0 {
+		if settings.Port == 80 || settings.Port == 443 {
```
KyleSanderson, line 300: Why?

ludviglundgren: For the clients that still has a port we want to keep that. But in the case of 80 and 443 they're not needed and are stripped.

Again for compatibility, but maybe a bad idea?

KyleSanderson: I haven't looked at the golang http client to see if there's a fast path or something like that for this. Adding branches like this not assigning a port just makes it more brittle long term. It's all minor, just sharing an opinion.

PR #252, opened 2022-04-19
feat: add ability for the user to choose the log dir during onboarding

stacksmash76:
refactor(config): change default logPath parent directory from `log` to `logs` for better wording.
refactor(domain): added a custom CreateUserRequest structure to use for the onboarding POST endpoint and refactored related code
feat(onboarding): added possibility to configure log directory from frontend onboarding page. added /onboard/preferences GET endpoint which is currently used to recommend a log directory. maybe the endpoint should account for an already non-standard logPath provided in config.toml
refactor(http/auth): removed cookie/is user authenticated checks which are redundant and are already implied with the userCount > 0 checks.
refactor(frontend/App): moved toaster a bit up in the hierarchy and made it available to unauthorized routes, so now for example, the Login page displays a toast on error, ditto for the Onboarding page.
refactor(frontend/HttpClient): removed specialized 403/404 error handling case.
refactor(frontend/Onboard): adapted for above-mentioned changes and fixed a few styling-related things.

remarks:
- GetPreferredLogDir() keeps tracks of errors it finds and passes them to frontend. The idea behind this is to report potential misconfiguration to the end-user.
- This PR only helps solve at which directory should the log file be located at, and the idea is for non-advanced users for the log file name to be "autobrr.log" and leave custom configuring to more advanced users.

TODO: Doesn't actually save the logPath to config.toml

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/autobrr/autobrr
 go 1.17
 
 require (
+	github.com/BurntSushi/toml v0.3.1
 	github.com/Masterminds/squirrel v1.5.1
 	github.com/anacrolix/torrent v1.11.0
 	github.com/asaskevich/EventBus v0.0.0-20200907212545-49d423059eef
--- a/internal/auth/service.go
+++ b/internal/auth/service.go
@@ -3,11 +3,10 @@ package auth
 import (
 	"context"
 	"errors"
+
 	"github.com/autobrr/autobrr/internal/domain"
 	"github.com/autobrr/autobrr/internal/user"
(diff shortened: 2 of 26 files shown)


ludviglundgren:
Refactored how the onboarding is done.
- move handlers from auth handler to new onboard handler
- move some logic to the config
- minor ui changes and update api calls

But a problem came up. Writing the toml config removes comments and the original layout. Not ideal. :exclamation: 

This other library https://github.com/pelletier/go-toml might be used with some extended functionality to write/edit toml configs.

**Lets revisit later**

PR #255, opened 2022-04-25
feat(actions): wait for the filter delay before performing actions

ghostserverd:
IDK if you have plans already for how to handle filter delay, but I need it for one of my trackers, and this is one low-touch way to implement it.

### DIFF
--- a/internal/announce/service.go
+++ b/internal/announce/service.go
@@ -3,6 +3,7 @@ package announce
 import (
 	"context"
 	"strings"
+	"time"
 
 	"github.com/autobrr/autobrr/internal/action"
 	"github.com/autobrr/autobrr/internal/domain"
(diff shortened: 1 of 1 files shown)


internal/announce/service.go
```diff
@@ -86,6 +87,11 @@ func (s *service) Process(release *domain.Release) {
 
 		var rejections []string
 
+		// sleep for the delay period specified in the filter before running actions
+		delay := time.Duration(release.Filter.Delay)
+		log.Info().Msgf("Delaying processing of '%v' (%v) for %v by %d seconds as specified in the filter", release.TorrentName, release.Filter.Name, release.Indexer, delay)
+		time.Sleep(delay * time.Second)
+
```
ludviglundgren, line 94: The default value of Delay will be 0, so a small enhancement could be to wrap this block in an `if ...Delay > 0`, and change log level to `Debug` instead of `Info`.

ludviglundgren (changes requested):
Hello!

Thanks for the PR. It was in the plans but had low priority so I appreciate you adding it!

This solution will work good and if there's more special needs later on we can fix it then.

Just two minor things I'd like to see changed 👍

internal/announce/service.go
ghostserverd: I made it info because I thought it might be confusing to see the `Matched` log and then no indicator of any action taking place, but it looks like the action logs are debug anyway, so that makes sense.

ludviglundgren:
Have made some major changes and this logic needs to go in `internal/release/service.go` instead. Could you update your branch with latest from develop, move logic and push again?

ghostserverd:
> Have made some major changes and this logic needs to go in `internal/release/service.go` instead. Could you update your branch with latest from develop, move logic and push again?

Will do

PR #261, opened 2022-05-01
feat(download-clients): qbt set re-announce opts

ludviglundgren:
Make the re-announce options settable in the action to customize intervals or disable it completely.

image


mouzzampk2014:
Nice work my friend

PR #285, opened 2022-05-28
refactor: Started refactoring codebase for React 18.

stacksmash76:
- Migrated to react-router v6 fully
- Formik broke and won't cooperate (TypeError: formik.getFieldProps is not a function, for one)
- Removed useless test setup along with relevant packages.
- Updated other dependencies, as well (see issue #234)

The PR is in very early phase. Major changes expected, especially w.r.t. Formik and how input forms are handled.

PR #296, opened 2022-06-10
fix: mobile ux improvements

stacksmash76:
fix(ListboxFilter): Added z-index in order to properly render dropdown box.
fix(ReleaseTable): Added flex-direction: column to wrap on mobile devices (otherwise it's set to row)
feat(DownloadClientSettingsListItem): Made it possible to toggle client state directly, without opening the update form.

### DIFF
--- a/web/src/components/data-table/Cells.tsx
+++ b/web/src/components/data-table/Cells.tsx
@@ -17,7 +17,7 @@ export const AgeCell = ({ value }: CellProps) => (
 
 export const TitleCell = ({ value }: CellProps) => (
   <div
-    className="text-sm font-medium box-content text-gray-900 dark:text-gray-300 max-w-[128px] sm:max-w-none overflow-auto py-4"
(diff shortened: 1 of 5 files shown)


ludviglundgren (commented):
Looks good! But I found one ui bug on filter details that probably came with the react 18 refactor.

Screen Shot 2022-06-10 at 23 08 06

Looks like it's [this change](https://github.com/autobrr/autobrr/pull/285/files#diff-3e21b97dc9b916aed60fb39277bef67216f3746fa5c541dbbcc3c8096360a367L232-L242). Before it would just scroll the filter tabs and worked good. Would you mind fixing it?

stacksmash76:
> Looks good! But I found one ui bug on filter details that probably came with the react 18 refactor.
> 
> Looks like it's [this change](https://github.com/autobrr/autobrr/pull/285/files#diff-3e21b97dc9b916aed60fb39277bef67216f3746fa5c541dbbcc3c8096360a367L232-L242). Before it would just scroll the filter tabs and worked good. Would you mind fixing it?

Don't know how I didn't spot that one! Should be fixed now. :+1:

ludviglundgren (approved):
:ship:

PR #299, opened 2022-06-11
feat(notifications): add telegram support

ludviglundgren:
Add Telegram support for notifications.


ludviglundgren:
Hi @mkgeeky and @daomengRen

I started work on adding Telegram support. What info would you like to see in the message?

For Discord it looks something like this.
image

And it will add rejections if there's any of those 
image

---

We can do some basic formatting but need your feedback :smiley:

mkgeeky:
Hi @ludviglundgren 

I think basic (rls name, status, index, filter & action) is good enough 💪

daomengRen:
Hi @ludviglundgren 
good enough,thanks

PR #300, opened 2022-06-11
fix(download-clients): qBit transform speed limit from B to KB

KyleSanderson:
qBittorrent transparently moves this from B/s to KB/s in the UI and POST request. Presently if you specify 500 in the autobrr KB/s field for qBittorrent you get a staggering 500B/s.


ludviglundgren (approved):
Nice find @KyleSanderson!

PR #303, opened 2022-06-14
fix(indexers): don't die on failed load of custom-definitions

KyleSanderson:
not tested yet.

### DIFF
--- a/internal/indexer/service.go
+++ b/internal/indexer/service.go
@@ -399,7 +399,7 @@ func (s *service) LoadCustomIndexerDefinitions() error {
 	outputDirRead, err := os.Open(s.config.CustomDefinitions)
 
 	if err != nil {
-		s.Log.Warn().Stack().Msgf("failed opening custom definitions directory %q: %s", s.config.CustomDefinitions, err)
+		s.log.Warn().Stack().Msgf("failed opening custom definitions directory %q: %s", s.config.CustomDefinitions, err)
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


internal/indexer/service.go
```diff
@@ -396,24 +396,27 @@ func (s *service) LoadCustomIndexerDefinitions() error {
...
+		return nil
 	}
 
-	if len(entries) == 0 {
+	defer outputDirRead.Close()
+
+	//entries, err := fs.ReadDir(Definitions, "definitions")
+	entries, err := outputDirRead.ReadDir(0)
+	if err != nil {
 		s.log.Fatal().Stack().Msgf("failed reading directory: %s", err)
```
ludviglundgren, line 411: This `log.Fatal` will make it crash if reached, but it likely won't since it does the check above first and returns if it can't be opened.

internal/indexer/service.go
```diff
@@ -396,24 +396,27 @@ func (s *service) LoadCustomIndexerDefinitions() error {
...
-		return err
 	}
 
 	customCount := 0
 
 	for _, f := range entries {
 		fileExtension := filepath.Ext(f.Name())
-		if fileExtension != ".yaml" {
+		if fileExtension != ".yaml" && fileExtension != ".yml" {
+			s.log.Warn().Stack().Msgf("skipping unknown extension definition file: %s", f.Name())
```
ludviglundgren, line 419: Great addition and a basic oversight from me not thinking of the short version.

ludviglundgren (approved):
Great find and fix! :+1:

internal/indexer/service.go
KyleSanderson: Permissions I think would be the only thing, which should be a fatal error.

PR #304, opened 2022-06-14
speculative fix: add freeleech identifier to AB

KyleSanderson:
speculative fix for Kaiser


KyleSanderson:
> This shouldn't be needed since `releaseTags` are parsed and can contain freeleech tags.
> 
> I'll leave it open for a bit to see if it makes any difference.

do you mean we should be parsing these tags to support the freeleech slider?

ludviglundgren:
Release tags are parsed here https://github.com/autobrr/autobrr/blob/develop/internal/domain/releasetags.go#L347

And looking at it again it seems the filter doesn't check the `bonus` field so that might be why it isn't working as expected!!

PR #313, opened 2022-06-16
web: better UX error interaction, updated dependencies

stacksmash76:
fix: remove react-cookie since we can't delete the user_session cookie using JS (due to httponly being set)
chore: update dependencies and set a global react dependency override (react-ridge-state creates a problem, this fixes it)
chore: tidy up APIClient, login and logout pages
fix: catch canOnboard() error in login.tsx
enhancement: add toast notify on intentional logout and nicely reset the context state; make sure screen-blinking is down to a minimum by having min-h-screen present.
fix: let onboarding redirect to / instead of /login (/login isn't used anymore)
fix: use normal <input /> caret cursor for SearchColumnFilter, instead of pointer


ludviglundgren (approved):
Great! I removed the `react-cookie` package as well :rocket:

PR #315, opened 2022-06-17
feat(irc): improve reconnect and add notifications

ludviglundgren:
Improve irc handlers

- nickserv auth
- nick handling (no more name change loop)
- reconnect handling
- notifications on unexpected disconnect and reconnect

## Tested

- [x] No nickserv
- [x] NickServ
- [x] NickServ and Invite command


internal/irc/handler.go
```diff
@@ -253,41 +256,113 @@ func (h *Handler) Restart() error {
...
+				h.log.Error().Stack().Err(err).Msgf("error sending connect command %v", h.network.InviteCommand)
+				return
+			}
+		}
+
+		//[test-bot Invalid parameters. For usage, do /msg NickServ HELP IDENTIFY]
+		if contains(msg.Params[1], "invalid", "help") {
+			h.log.Debug().Msgf("NOTICE nickserv invalid: %v", msg.Params)
+
+			h.client.Send("PRIVMSG", "NickServ", fmt.Sprintf("IDENTIFY %v %v", h.network.NickServ.Account, h.network.NickServ.Password))
```
KyleSanderson, line 343: I'm under the impression that this is not universal functionality. It's definitely silly but this should likely old style if the nicks match, and if they don't then pass the nick.

ludviglundgren: This is not totally set in stone yet as it still needs more testing, but this is a backup for if the regular nickserv with just password fails. On Ergo it seem to need both username and password, while I've seen the opposite on some networks where it's not supported to identify with both nick and pass. Fun inconsistencies.

This is a bit different but in [ZNC there's this ](https://github.com/znc/znc/blob/znc-1.8.2/modules/nickserv.cpp) which was referenced by the go-irc maintainer.

PR #320, opened 2022-06-22
fix(web): react query error handling

stacksmash76:
fix(App): fix incorrect ordering of ErrorBoundary and QueryClientProvider (ErrorBoundary is now the parent). this should hopefully fix the "we caught an unrecoverable error" on a 401 request, where only a notification should be shown (along with the user getting deauthed and sent to the login page)
fix(ErrorPage): add padding to the page for mobile devices
chore(react-query): wrap APIClient calls in anonymous functions to avoid passing react-query context variables by accident

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -32,17 +32,17 @@ export function App() {
   const settings = SettingsContext.useValue();
 
   return (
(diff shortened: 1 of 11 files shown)


ludviglundgren (approved):
:rocket:

PR #323, opened 2022-06-29
refactor(discord): remove hardcoded username

varoOP:
Let the Discord Webhook use it's set name. The webhook cannot be created without a username so it should not create issues.

This way the webhook is not always named `brr`.


ludviglundgren (approved):
🚀

PR #324, opened 2022-06-29
feat(config): add container detection capabilities

KyleSanderson:
untested.


KyleSanderson:
> How will this behave on windows? The err from os.Open is ignored and it won't find that file so will that cause a panic? Haven't got a windows machine or vm to test on currently.

It checks if the ptr is valid before proceeding, and if it isn't it will skip the branch.

ludviglundgren (approved):
🚢

PR #325, opened 2022-06-29
feat(AB-compatibility): Enable autobrr to grab anime,LNs,& manga from AB

varoOP:
- Gets release group from AB correctly.
- Adds epub as a container.


varoOP:
Found a better way to do the same thing. Will create another PR.

PR #333, opened 2022-07-03
feat(irc): attempt SASL login with fallback to nickserv

slingamn:
I also tried updating the integration test config to be closer to Ergo's `default.yaml`. I couldn't figure out how to actually run the integration tests.

Anyway, if this works out for you then I think I'll release the current irc-go master branch as v0.2.0.


ludviglundgren:
Appreciate the PR a lot! Did some additional work on it to cleanup and refactor a bit how the code flows and listen to the proper events then do next thing if needed etc.

ludviglundgren (approved):

PR #334, opened 2022-07-03
feat(macros): expose parsed dates from rls string

KyleSanderson:


ludviglundgren:
Like we discussed before it would make more sense to change the variables for Year, Month and Day to CurrentYear, CurrentMonth and CurrentDay but that would be a breaking change. But other than you I'm not sure how many actually use it.

v1.0.0 is getting near so technically following semver that would be fine.

What do you think?

ludviglundgren (approved):
I made the change to add the `Current` prefix to `Year, Month, Day, Hour. Minute, Second`.

This is a breaking change. v1.0.0 soon 🚀

Issue #349, opened 2022-07-10
watch dir action dont create folders


DerUntote:
Iam using rutorrent for downloading my files and want to use the autolabel feature, which works with subdirs in the watchfolder

Currently, i get errors, cause of "watch dir" action doesn't create folders on its way to place the torrent file.

Can we have this please as an option or forced?

ludviglundgren:
Hi! That should likely be taken care of. Does that plugin create deeply nested structures or is it just subdirs in the main download dir?

Does it save to say `movies`? If so just create the directories yourself to start with.

DerUntote:
https://github.com/DerUntote/autobrr
I added what I meant to my own fork. Thanks for sharing your work so we can extend our needs ourselves :)

ludviglundgren:
Oh nice, well in that case let's get that into here as well! 😄

DerUntote:
made a pull request ;)

PR #351, opened 2022-07-10
feat(actions): watch dir create folder if not exists

DerUntote:
https://github.com/autobrr/autobrr/issues/349

Iam no Go developer, please review

### DIFF
--- a/internal/action/run.go
+++ b/internal/action/run.go
@@ -153,6 +153,12 @@ func (s *service) watchFolder(action domain.Action, release domain.Release) erro
 	_, tmpFileName := path.Split(release.TorrentTmpFile)
 	fullFileName := path.Join(watchFolderArgs, tmpFileName+".torrent")
 
+	// Create folder
+	err := os.MkdirAll(watchFolderArgs, os.ModePerm);
+	if err != nil {
+		return errors.Wrap(err, "could not create new folders %v", fullFileName)
+	}
+	
 	// Create new file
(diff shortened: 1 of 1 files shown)


ludviglundgren:
Have you tested it yourself? Looks simple enough 👍

ludviglundgren (approved):
We have similar code already for config dir. It will not do anything if already exists so that's great.

Thanks for your contribution 🥇

internal/action/run.go
```diff
@@ -153,6 +153,12 @@ func (s *service) watchFolder(action domain.Action, release domain.Release) erro
 	_, tmpFileName := path.Split(release.TorrentTmpFile)
 	fullFileName := path.Join(watchFolderArgs, tmpFileName+".torrent")
 
+	// Create folder
+	err := os.MkdirAll(watchFolderArgs, os.ModePerm);
```
ludviglundgren, line 157: You need to change it to `err =` to not reassign the err.

DerUntote: Can i add my fix to this pull request or do we need a new one?

DerUntote: why is it ok to use := in line 147 and also again in 163?

ludviglundgren: You can make a commit and push to your develop branch it this pr will update.

As to why - it's already been declared, and down on 163 we create two new variables so that's different. Not a great answer but that's how it is.

PR #378, opened 2022-07-27
Add multiarch build for ARM, ARM64 and AMD64

shapechecker:
Updated workflow to support mutilarch and configured ARM, ARM64 and AMD64 builds
Updated dockerfile to fix yarn timeout issue that was introduced with multiarch build


zze0s:
Hi! Thank you a lot for this!!

Just a shame it increases the build with more than 4x so will look into setting up my own runner and see if it makes a difference.

Will wait a bit with merge until tested things etc.

zze0s:
These buildx workflows with Node are stupid, considering it will build static files anyway. I'll try some more stuff like build the frontend directly with actions and then just inject that into a new Dockerfile.

zze0s (approved):
I think that's it. Last error is just the permissions because this is from a fork. Got it down from 30 minutes something, to now last run at 11 minutes. I'm fine with that.

I'll go ahead a merge this. 👍

PR #388, opened 2022-08-01
feat(torznab): add Jackett support

KyleSanderson:


KyleSanderson:
jackett now passes caps atleast, the other invocation I think is on get so it's patched there as well. Prowlarr (caps atleast) continues to function with the change.

zze0s (approved):
Made some changes
* Only add apikey if it exists (hopefully it's always there anyway)
* Remove the `X-API-Key` header since only Prowlarr support it
* Update test to reflect change

I left the code commented out if we ever go back to check.

Tested with both Prowlarr and Jackett with same working result 👍

Issue #389, opened 2022-08-01
Misleading feed refresh interval


KptCheeseWhiz:
Hello,

The refresh interval for the feed interface is misleading..

If we look at the code, it is adding the minutes to a ["*/%d * * * *" cron job interval](https://github.com/autobrr/autobrr/blob/develop/internal/feed/service.go#L277), so if you put a number that is not a multiple of 60, the job will also be executed at the start of the hour. 

As an example, if we put in 25 minutes, the job will be executed at [0, 25 and 50 minutes every hour.](https://crontab.guru/#*/25_*_*_*_*) which makes a 25-25-10 minutes interval.

Thank you :)

KyleSanderson:
hi - if you want to test you can try this image `ghcr.io/autobrr/autobrr:pr-390`. Based on feedback it will be merged in the coming days.

PR #390, opened 2022-08-01
fix(scheduler): use time as opposed to cron fields

KyleSanderson:
per https://github.com/autobrr/autobrr/issues/389 it seems to be a bit nutty right now where we're using cron fields to schedule work to be completed. This results not only in a thundering herd problem, but unexpected behaviour for people unfamiliar with the intricates of cron (and the specific flavour of implementation by the library owner).

This PR flips from fixed fields to use duration - it's completely untested and just from docs.

### DIFF
--- a/internal/feed/service.go
+++ b/internal/feed/service.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/dcarbone/zadapters/zstdlog"
 	"github.com/rs/zerolog"
+	"time"
 )
 
 type Service interface {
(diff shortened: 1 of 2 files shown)


KyleSanderson:
Testing now - we're now enforcing a minimum of 15m for feeds and I backed off the update check to 36 hours (day and a half) from every 6 hours.

zze0s (approved):
Great improvement and quickly done, thank you Kyle!

Closes #389

PR #395, opened 2022-08-02
fix(indexers): fix ptm regexp to make preGAP optional

ghostserverd:
`preGAP` is not always present in the feed for `pretome` so update the match regexp to make it optional.

### DIFF
--- a/internal/indexer/definitions/ptm.yaml
+++ b/internal/indexer/definitions/ptm.yaml
@@ -46,7 +46,9 @@ parse:
     - test:
         - "tehFire: [Applications|Windows] Chen went to the Mall :: preGAP: 1m and 32s :: https://pretome.info/details.php?id=696969"
         - "[Movies|x264] Orlando.Bloom.Had.A.Cow-ze0s :: preGAP: P2P source :: https://pretome.info/details.php?id=646321"
-      pattern: '\[([^\]]+)\] ([^:]+) :: [^:]+.*(https?\:\/\/[^\/]+).*id=(\d+)'
+        - "tehFIRE: [TV|XviD] Royal.Institution.Christmas.Lectures.2009.Part2.WS.PDTV.XviD-WATERS :: preGAP: 1m and 9s :: https://pretome.info/details.php?id=127107"
+        - "tehFIRE: [TV|x264] Newsreaders.S01E05.HDTV.x264-2HD https://pretome.info/details.php?id=333951"
+      pattern: '\[([^\]]+)\] ([^:]+)(?: :: [^:]+:.* :: )?(https?\:\/\/[^\/]+).*id=(\d+)'
       vars:
         - category
         - torrentName


KyleSanderson (approved):
thank you very much

ghostserverd:
Is there anything I can do here to fix the Docker build failure? Kind of looks like it might just be a permission issue.

KyleSanderson:
> Is there anything I can do here to fix the Docker build failure? Kind of looks like it might just be a permission issue.

Yeah it's just permissions - nothing to do with your commit.

zze0s (approved):
Looks good 👍

I saw you had some edits on the parsing as well, like this https://github.com/autobrr/autobrr/commit/335f47e418a9697a95091e1c01b4a57b2d8e902f

Did the regex change fix so that wasn't necessary?

ghostserverd:
> Looks good 👍
> 
> I saw you had some edits on the parsing as well, like this [335f47e](https://github.com/autobrr/autobrr/commit/335f47e418a9697a95091e1c01b4a57b2d8e902f)
> 
> Did the regex change fix so that wasn't necessary?

I had originally implemented this as an additional `line` in the yaml file so it was necessary to support multiple lines (which that change should do). However, I cleaned up the regexp to remove that requirement.

I also wanted to make sure I understood why it had been implemented as it was (the commented out `//patternParsed := false`) before opening a PR, but I've been running with this change for a while now, and the logic is sound. Might as well have support for multiple `line`s since the `yaml` has it defined as a list.

I can add that change to this PR or open another one. Whatever's easier.

zze0s:
Thanks for the explanation 😄 

We actually already support multi-line but the syntax is a bit different. Take a look at [btn](https://github.com/autobrr/autobrr/blob/develop/internal/indexer/definitions/btn.yaml) and [tt](https://github.com/autobrr/autobrr/blob/develop/internal/indexer/definitions/trancetraffic.yaml)

As for the commented code, at this point I've forgotten why it's there 🤷 

Then I'll go ahead and merge this. Thanks for your contribution! ❤️

ghostserverd:
> Thanks for the explanation 😄
> 
> We actually already support multi-line but the syntax is a bit different. Take a look at [btn](https://github.com/autobrr/autobrr/blob/develop/internal/indexer/definitions/btn.yaml) and [tt](https://github.com/autobrr/autobrr/blob/develop/internal/indexer/definitions/trancetraffic.yaml)
> 
> As for the commented code, at this point I've forgotten why it's there 🤷
> 
> Then I'll go ahead and merge this. Thanks for your contribution! ❤️

I think the issue with the current implementation is that if the first pattern fails to match, none of the other patterns will be checked because it `break`s if `!match`. 

https://github.com/autobrr/autobrr/blob/develop/internal/announce/announce.go#L90

I believe we would want it to continue so it checks subsequent `lines` and only break once it finds a successful match. Which is what my other change does. Unless maybe designating it as `multi` sends it down a different code path somewhere? I'll open a PR for discussion.

Thanks for this fantastic tool! It's so much better than a autodl-irssi setup, and I think even better than trackarr. 🎉

ghostserverd:
Oh wait, is multi-line support specifically to support a release message that's multiple lines, not to support multiple different possible regexp's for a given indexer's release messages? I think I didn't fully grasp what it's meant to be used for.

If that's the case, it's clearer now to me why it would be implemented to break if the first line does not match.

Although it does seem like it might be nice to support multiple different regexp's even if the indexer's release message is only a single line. It would allow for more targeted (and I'd argue probably more readable and less error-prone) regexp's instead of having to have one that supports all possible formats.

Anyway that's probably a discussion for a different thread. Sorry for cluttering this PR!

KyleSanderson:
> Oh wait, is multi-line support specifically to support a release message that's multiple lines, not to support multiple different possible regexp's for a given indexer's release messages? I think I didn't fully grasp what it's meant to be used for.
> 
> If that's the case, it's clearer now to me why it would be implemented to break if the first line does not match.
> 
> Although it does seem like it might be nice to support multiple different regexp's even if the indexer's release message is only a single line. It would allow for more targeted (and I'd argue probably more readable and less error-prone) regexp's instead of having to have one that supports all possible formats.
> 
> Anyway that's probably a discussion for a different thread. Sorry for cluttering this PR!

All conversation is welcome. Feel free to hit me or others on Discord if it can be improved.

PR #398, opened 2022-08-04
fix(irc): locks + disconnect + nickserv

KyleSanderson:


zze0s:
Closed in favor for #399

PR #399, opened 2022-08-06
fix(irc): rewrite handler pipeline

KyleSanderson:
This might be overkill but the pipeline has been made event driven. I've tested on a couple networks, and bouncing nicks / connections brings it back every time. creating the a different branch from https://github.com/autobrr/autobrr/pull/398 because it's pretty intrusive (and I didn't apply the GUI changes).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -60,6 +60,7 @@ require (
 	github.com/mitchellh/mapstructure v1.5.0 // indirect
 	github.com/pelletier/go-toml v1.9.5 // indirect
 	github.com/pelletier/go-toml/v2 v2.0.2 // indirect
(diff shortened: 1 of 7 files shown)


zze0s:
Running with `-race` and the new deadlock mutex library it showed a lot of non working code with nested locks/unlocks etc.

Only tested locally and works fine so far. Still needs to be tested with more real networks and simulated network issues.

Planning on bringing in the UI changes. Added a new field to the network struct, `healthy` which is a bool and checks a couple of things to determine if the network is healthy.

- Enabled and connected
- If any channel that should be monitored isn't

This is to be considered a WIP still.

PR #406, opened 2022-08-11
feat(indexers): add Fuzer

SweetMNM:


internal/indexer/definitions/fuzer.yaml
```diff
@@ -0,0 +1,61 @@
...
+        - name1
+        - category
+        - uploader
+        - torrentSize
+        - baseUrl
+        - torrentId
+        - torrentName
+
+  match:
+    torrenturl: "https://www.{{ .baseUrl }}/rss/torrent.php/{{ .torrentId }}/{{ .uid }}/{{ .passkey }}/{{ .torrentName }}.torrent"
```
zze0s, line 61: It needs to encode the torrentName, like seen in the [autodl-irssi file](https://github.com/mkgeeky/autodl-trackers/blob/master/trackers/Fuzer.tracker#L87)

It's enough to add this on line 62. Like [here for tl](https://github.com/autobrr/autobrr/blob/9aaf79e7bb5153f545ea0fe056c9f5a2c370934d/internal/indexer/definitions/torrentleech.yaml#L60-L61)

```yaml
encode:
      - torrentName
```

zze0s (approved):
Thanks for your contribution! 🥇

PR #408, opened 2022-08-14
feat(api): add apikey support

zze0s:
Add apikey support in addition to current cookie support.

Supported types:

* Header: `X-API-Key`
* Query param: `?apikey=KEY`


zze0s:
image
image

PR #414, opened 2022-08-21
feat(docker): expose default port

agneevX:


zze0s (approved):
Hi! Thanks for your contribution! This might be useful for those doing `docker run`.

PR #449, opened 2022-09-02
fix(web): fix artist placeholder typo in music filters

xoaaC:
First PR ever - be gentle if I made a mistake in creating a pull request!

### DIFF
--- a/web/src/screens/filters/details.tsx
+++ b/web/src/screens/filters/details.tsx
@@ -401,7 +401,7 @@ export function Music() {
   return (
     <div>
       <div className="mt-6 grid grid-cols-12 gap-6">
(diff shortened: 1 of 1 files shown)


KyleSanderson (approved):

zze0s (approved):
Hi! Thank you! Looks good!

We are about to introduce a license on the project, but have yet to decide which.

We are leaning towards GPLv2/v3, MIT, BSD or similar. Are you ok with that? Those are commonly used in open source projects.

The core contributors have all agreed to follow my decision within reason, so if you don't know a whole lot about licenses then take that as a sign of good faith and confirm with a `Yes I agree`.

xoaaC:
Yes I agree.

PR #460, opened 2022-09-10
feat(irc): handle nickserv without mode change

zze0s:
Force NickServ auth for networks not setting new `MODE` reply.


KyleSanderson (approved):
my gut says this is the wrong thing to do, but there's buggy IRCd's out there... so. :shipit:

zze0s:
After a discussion we came to the conclusion that this isn't the correct solution for the core problem.

The networks where this is an issue allows for access without NickServ/SASL auth.

Planned thing to try out is to implement a way to set auth protocol.

PR #470, opened 2022-09-23
fix(web): TS typings and always active navbar link

stacksmash76:
- fix TS typings since `yarn build` was failing on my machine.
- also fixed the dashboard navlink (it was always active on my end)


zze0s (approved):
:+1:

PR #473, opened 2022-09-24
fix(web): indexer options stale caching

stacksmash76:
- de-duplicate incompatible queries which led to a stale type-unsafe caching bug


zze0s (approved):
:rocket:

PR #490, opened 2022-10-08
feat(clients): add Readarr support

voltron4lyfe:
Initial Readarr support.  Working on hack to create a title that MaM will actually parse.  Of course MaM isn't really having many releases right now


voltron4lyfe:
I haven't yet had a successful push but I'm having successful rejects and the Readarr parsing looks right.  May only work with MaM but that's probably what most people want.

Issue #362, opened 2022-07-20
Implement retries on failed download actions


martyluky:
Currently if an error occurs during downloading a torrent file 
autobrr will not attempt to retry downloading the torrent file resulting in an PUSH_ERROR / rejection.

This could be prevented in implementing retries for following error:
`Status: PUSH_ERROR`
`...`
`error downloading torrent file for release: [redacted]: error writing downloaded file: /tmp/autobrr-3550693884: read tcp [redacted]:58372->[redacted]:443: read: connection reset by peer`

Keep up the amazing work! Appreciate it big time!

PR #491, opened 2022-10-11
feat(releases): retry failed downloads

KyleSanderson:
Fixes https://github.com/autobrr/autobrr/issues/362

There's been an incredible influx of bad advice floating around where implementing arbitrary delays is the suggested method to resolve this. This implements a retry, and quick validity check if the file is valid before proceeding.


stacksmash76:
IMO, I wouldn't retry on 401 (Unauthorized), 403 (Forbidden), 405 (Method Not Allowed) and 410 (Gone).
Those will pretty much always be pure spam.

zze0s (approved):
Made some changes, added tests and did manual testing as well with everything looking OK :+1:

PR #522, opened 2022-11-05
fix(web): download client table overflow

xoaaC:
Alternative solution for a part of https://github.com/autobrr/autobrr/pull/521

@stacksmash76 solution contains the table to the div now but part of the table will be cut off if the values in the table are too long.

chrome_EOhTRVNPJ0

This PR has following changes:
• Converted DownloadClient table into a responsive grid.
• Overflowing content will now truncate and hovering over it will show you the full value.

chrome_TxyWEuyVE7

In case someone has concerns about this PR, feel free to close it.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -7,6 +7,9 @@ assignees: ''
 
 ---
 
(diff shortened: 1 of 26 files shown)


zze0s (approved):
Great improvement! :+1:

PR #528, opened 2022-11-06
fix(web): infinitely retry queries, remove "ago" from age cells

### LINKED ISSUE #497: enhancement: shorten time column data format
Since the column is usually named `age`, it makes sense to remove the ` ago` suffix, and it might be beneficial to use shorthand notation for days, hours, minutes, seconds. This change would save some screen estate.

stacksmash76:
- infinitely retry web queries so we avoid the "failed to fetch" error when the web server is unavailable
- remove the "ago" suffix from age cells (closes #497)


zze0s (approved):
Good improvement!

Issue #499, opened 2022-10-16
Cryptic rejection reason for not selecting a client in the qBittorrent action for a filter.


xoaaC:
**Describe the bug**
Currently - if you don't select a client in a qBittorrent action for a filter you get this cryptic rejection reason:
`error finding client: 0: error scanning row: sql: no rows in result set` 

**To Reproduce**
1. Create a filter with a qBittorrent action
2. In the clients dropdown menu **do not** choose a client
3. Save and enable the filter
4. In case this filter now matches a release you get the above error as a rejection reason.

**Expected behavior**
Since this is quite a cryptic rejection reason it could maybe be catched and replaced by a hint - something along the line of:
`No client selected in the qBittorrent client dropdown menu. Please select a client.`

**Additional context**
@KyleSanderson said this _should_ be a 4 line improvement and to file an issue as a reminder to improve this.

PR #529, opened 2022-11-06
fix(download_client): improve no-client error

KyleSanderson:
Fixes https://github.com/autobrr/autobrr/issues/499


zze0s (approved):
An alternative solution would have been to check the ID in this function or the caller if it was the default value of 0, and then return the error instead of checking the database.

PR #531, opened 2022-11-08
fix(indexers): immortalseed regex

xoaaC:
fix(indexers): immortalseed regex
Fixed issue where the regex for matching the announce line from immortalseed only had 1 whitespace in front of "Link:"
but the announce line had 2 whitespaces in front of "Link:"

Added "\s+" (instead of " ") in front of "Link:" to match one or more whitespaces and simplified regex.

Other changes:
change IRC port to 7000
set TLS to true
replace "IRCKEY" in placeholder and help test with "PASSKEY" for invite_command field.

Tested with a sample size of 50 of which 50 matched. Before none of these announces matched.
Will report back again when i tested this regex with a bigger sample size.


zze0s (approved):
:rocket:

PR #538, opened 2022-11-12
fix(feeds): indexer identifier issues

KyleSanderson:
Presently to avoid duplicate identifiers, we don't store the implementation name in the database. However, as luck may have it, humans also use these names (and may for instance, want to name their feed `DistroWatch - RSS`). This adds a strange complexity to the issue, because if another feed was then named `DistroWatch - RSSRSS` you'd encounter a duplicate error. While the aforementioned isn't likely, it still results in a bad situation never the less. Torznab also does this, so this backs out the string replacement.

There's another issue present, where if the indexer ends with - the releases page doesn't have the trailing - which results in a query against a non-existent indexer, returning 0 rows. This was found working through bzwarts instance on 1.10.

### DIFF
--- a/internal/database/feed.go
+++ b/internal/database/feed.go
@@ -27,22 +27,23 @@ type FeedRepo struct {
 func (r *FeedRepo) FindByID(ctx context.Context, id int) (*domain.Feed, error) {
 	queryBuilder := r.db.squirrel.
 		Select(
-			"id",
-			"indexer",
-			"name",
-			"type",
-			"enabled",
-			"url",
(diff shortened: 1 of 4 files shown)


KyleSanderson:
This may also be incomplete because this is done twice(?)

https://github.com/autobrr/autobrr/blob/4bf023d0306510dce12e6ab4e77085bc4ec81366/web/src/forms/settings/IndexerForms.tsx#L186

zze0s (approved):
:rocket:

PR #540, opened 2022-11-14
fix(actions): watch dir name check torrent suffix

TMD20:
Make conditional for generating random torrent name more specific


zze0s (approved):
What a fun edge-case I didn't think of :smile:

PR #543, opened 2022-11-17
feat: add usenet support

KyleSanderson:
This branch begins work on supporting Usenet announcements, and pushing them to your action of choice.

Fixes https://github.com/autobrr/autobrr/issues/544 (🐔 or the 🥚? 😆)


zze0s (approved):
I think that's it for now!

Tested with newznab feeds directly from indexers as well as from Prowlarr and it all works as intended.

Will add NzbGet support later on if people are interested.

PR #545, opened 2022-11-18
fix(web/filters): truncate too long name

xoaaC:
Fixed an issue where the div containing the filter name in the delete modal would overflow if the filter name is too long.
    Text inside the div is now contained and will word-break when too long.
Fixed an issue where the filter name headline would break into a second line if the filter name is too long.
    In case the filter name is too long the filter name headline will now truncate and show it's full value at mouseover.

chrome_2q3dibEFHY


zze0s (approved):
Great improvement!!

PR #548, opened 2022-11-19
fix(web): relative time timezone issues

zze0s:
Web ui shows different times in the UI.


zze0s:
This brought back errors with filter limits. Could be only during the time transition and for a couple of hours, but unclear.

After further investigation of timezone mismatch it could be related to Postgres.

PR #555, opened 2022-12-02
feat(tools): Add a simple MockIndexer helper for IRC announcing and indexing

vktr:
This adds a simple self indexer to autobrr. The idea is to have an easy way of testing IRC announces locally.

The self indexer implements just enough of the IRC protocol for autobrr to happily connect and listen for announces. It also hosts the torrent files that autobrr downloads. This makes the self indexer an easy way to have a contained testing setup for both IRC announces with subsequent file downloading.

There is almost no error handling, logging is crude, but it is tiny and it works.


zze0s (approved):
This is great!! Highly appreciated! :clap:

PR #556, opened 2022-12-02
feat(web): When adding the first action, display it in edit mode

vktr:
When the first action is added to a filter, show it in edit mode for increased discoverability.


zze0s (approved):
Nice! :+1:

PR #558, opened 2022-12-04
feat(clients): add support for qBittorrent 4.4.0+

zze0s:
Add support for qBittorrent 4.4.0 and up.

* replace with shared library from [autobrr/go-qbittorrent](https://github.com/autobrr/go-qbittorrent)
* cache download client


internal/download_client/service.go
```diff
@@ -112,10 +125,52 @@ func (s *service) Test(client domain.DownloadClient) error {
...
 		s.log.Error().Err(err).Msg("client connection test error")
 		return err
 	}
 
 	return nil
 }
+
+func (s *service) GetCachedClient(ctx context.Context, clientId int32) *domain.DownloadClientCached {
+	// check if client exists in cache
+	cached, ok := s.qbitClients[clientId]
```
KyleSanderson, line 138: Is this safe for concurrent accesses?

internal/domain/client.go
```diff
@@ -57,3 +68,54 @@ const (
...
+func (c DownloadClient) qbitBuildLegacyHost() string {
+	// parse url
+	u, _ := url.Parse(c.Host)
+
+	// reset Opaque
+	u.Opaque = ""
+
+	// set scheme
+	scheme := "http"
+	if u.Scheme == "http" || u.Scheme == "https" {
```
KyleSanderson, line 88: bad branch

internal/action/qbittorrent.go
```diff
@@ -83,38 +39,41 @@ func (s *service) qbittorrent(action domain.Action, release domain.Release) ([]s
...
+	s.log.Info().Msgf("torrent with hash %v successfully added to client: '%v'", release.TorrentHash, c.Dc.Name)
 
 	return nil, nil
 }
 
 func (s *service) prepareQbitOptions(action domain.Action, m domain.Macro) (map[string]string, error) {
-
 	opts := &qbittorrent.TorrentAddOptions{}
 
+	opts.Paused = false
```
KyleSanderson, line 65: Can these be set directly in the struct initialization?

KyleSanderson (changes requested):

internal/action/qbittorrent.go
zze0s: The zero value of this is already false so maybe unnecessary.

internal/download_client/service.go
zze0s: Good catch, fixed with mutex.

internal/domain/client.go
zze0s: It's apparent how it would be that. This is here for legacy reasons when we constructed the `host` for qbit where it was valid to not have the scheme in the host field.

KyleSanderson: understood, but the else branch is the exact same code... so this doesn't need to be in a conditional statement (unless if that's the bug here 😉)

zze0s: Sometimes it's just right in front of you. HA! Will resolve.

PR #565, opened 2022-12-12
fix(releases): improve load time by 47x

KyleSanderson:
Presently we scan the entire DB on each line item returned by the database engine, in an effort to populate the single "total releases" counter on the bottom of the page. This results in a 47x slowdown on my system as the calculation is done for every line item returned. While it's more appropriate to break this out into another query entirely, changing the query to work on the static set of data in memory results in a massive speedup (47x on a classic SSD, much more on rotational systems).

EDIT: Removed one of the subqueries on stats, the speedup is only 40% (100ms savings), but I'm sure that's better than it was. Using Group By as opposed to renaming the fields like this is also faster, but would break the webui.

https://www.youtube.com/watch?v=yK0P1Bk8Cx4


zze0s (approved):
Need some testing on Postgres before merge, but the code looks like generic SQL so should be good :+1:

PR #569, opened 2022-12-18
refactor(database): clean-up queries

KyleSanderson:
SQLite defaults to sql.LevelSerializable, which takes a full atomic lock during the transaction. MSSQL defaults to sql.LevelReadCommitted, which seems like a reasonable compromise. Presently untested.

Documentation: https://www.sqlite.org/isolation.html
https://pkg.go.dev/database/sql#IsolationLevel

### DIFF
--- a/internal/database/filter.go
+++ b/internal/database/filter.go
@@ -40,7 +40,7 @@ func (r *FilterRepo) Find(ctx context.Context, params domain.FilterQueryParams) 
 		return nil, err
 	}
 
-	if err = tx.Commit(); err != nil {
+	if err := tx.Commit(); err != nil {
 		return nil, errors.Wrap(err, "error commit transaction find releases")
 	}
 
@@ -319,6 +319,10 @@ func (r *FilterRepo) FindByIndexerIdentifier(indexer string) ([]domain.Filter, e
 			continue
 		}
 		filters[i].Downloads = downloads
+	}
+
+	if err := tx.Commit(); err != nil {
+		return nil, errors.Wrap(err, "error finding filter by identifier")
 	}
 
 	return filters, nil
@@ -949,8 +953,7 @@ func (r *FilterRepo) StoreIndexerConnections(ctx context.Context, filterID int, 
 		r.log.Debug().Msgf("filter.StoreIndexerConnections: store '%v' on filter: %v", indexer.Name, filterID)
 	}
 
-	err = tx.Commit()
-	if err != nil {
+	if err := tx.Commit(); err != nil {
 		return errors.Wrap(err, "error store indexers for filter: %v", filterID)
 	}
 
--- a/internal/database/filter.go
+++ b/internal/database/filter.go
@@ -29,7 +29,7 @@ func NewFilterRepo(log logger.Logger, db *DB) domain.FilterRepo {
 }
 
 func (r *FilterRepo) Find(ctx context.Context, params domain.FilterQueryParams) ([]domain.Filter, error) {
-	tx, err := r.db.BeginTx(ctx, &sql.TxOptions{})
+	tx, err := r.db.BeginTx(ctx, &sql.TxOptions{Isolation: sql.LevelReadCommitted})
 	if err != nil {
 		return nil, errors.Wrap(err, "error begin transaction")
 	}
--- a/internal/database/filter.go
+++ b/internal/database/filter.go
@@ -233,7 +233,7 @@ func (r *FilterRepo) FindByID(ctx context.Context, filterID int) (*domain.Filter
 			"updated_at",
 		).
 		From("filter").
-		Where("id = ?", filterID)
+		Where(sq.Eq{"id": filterID})
 
 	query, args, err := queryBuilder.ToSql()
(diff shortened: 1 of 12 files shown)


KyleSanderson:
So presently when this runs https://github.com/autobrr/autobrr/blob/0f3ce7573e4b18a93b5090b5b5e9285f049095bc/internal/database/release.go#L388 it runs 11 queries (minimum) in a single transaction while holding the table lock. This is because commit is not called until the end to release the transaction. I'm seeing a massive improvement here by using something like 
```
	queryBuilder := repo.db.squirrel.
		Select("r.id", "r.filter_status", "r.rejections", "r.indexer", "r.filter", "r.protocol", "r.title", "r.torrent_name", "r.size", "r.timestamp", "ra.client", "ra.filter").
		FromSelect(repo.db.squirrel.
			Select("*").
			From("release").
			OrderBy("r.id DESC").
			Limit(10), "r").
		LeftJoin("release_action_status as ra ON r.id = ra.release_id")
```

@zze0s what do you think?

internal/database/filter.go
```diff
@@ -70,7 +70,7 @@ func (r *FilterRepo) find(ctx context.Context, tx *Tx, params domain.FilterQuery
 		From("filter f")
 
 	if params.Search != "" {
-		queryBuilder = queryBuilder.Where("f.name LIKE ?", fmt.Sprint("%", params.Search, "%"))
+		queryBuilder = queryBuilder.Where(sq.Like{"f.name": params.Search + "%"})
```
zze0s, line 73: Is it intentional to skip the prefix `%`?

internal/database/release.go
```diff
@@ -511,7 +471,7 @@ func (repo *ReleaseRepo) CanDownloadShow(ctx context.Context, title string, seas
 	queryBuilder := repo.db.squirrel.
 		Select("COUNT(*)").
 		From("release").
-		Where("title LIKE ?", fmt.Sprint("%", title, "%"))
+		Where(sq.Like{"title": title + "%"})
```
zze0s, line 474: Same here as above. Previous `%title%` new `title%`

zze0s (approved):
Great improvements Kyle!

Will do some testing of this and see how it handles.

internal/database/release.go
```diff
@@ -126,7 +118,7 @@ func (repo *ReleaseRepo) Find(ctx context.Context, params domain.ReleaseQueryPar
 
 func (repo *ReleaseRepo) findReleases(ctx context.Context, tx *Tx, params domain.ReleaseQueryParams) ([]*domain.Release, int64, int64, error) {
 	queryBuilder := repo.db.squirrel.
-		Select("r.id", "r.filter_status", "r.rejections", "r.indexer", "r.filter", "r.protocol", "r.title", "r.torrent_name", "r.size", "r.timestamp", "(SELECT COUNT(*) FROM release) AS total_count").
+		Select("r.id", "r.filter_status", "r.rejections", "r.indexer", "r.filter", "r.protocol", "r.title", "r.torrent_name", "r.size", "r.timestamp", "ras.id", "ras.status", "ras.action", "ras.type", "ras.client", "ras.filter", "ras.rejections", "ras.timestamp", "(SELECT COUNT(*) FROM release) AS total_count").
 		From("release r").
 		OrderBy("r.id DESC")
```
zze0s, line 123: Found an issue with this 😄 since it fetches the actions and some releases have multiple actions, the limit becomes the total limit. So if we have `LIMIT 10` and two releases have two actions, we'll only get 8 actual releases.

Well described in this article along with some alternative fixes. Will try one of them!

internal/database/filter.go
KyleSanderson: Yeah. Right now if you search for `Dog` you'll get Dog.S02E04 and `Mary.Poppins.S01E04.Clifford.The.Dog`.

internal/database/release.go
KyleSanderson: There's inherently less operations with yours (by a far margin because there's no copies), time seems to be consistent with both.
```
	subQueryBuilder := repo.db.squirrel.
		Select("*").
		From("release").
		OrderBy("id DESC")

	if params.Limit > 0 {
		subQueryBuilder = subQueryBuilder.Limit(params.Limit)
	} else {
		subQueryBuilder = subQueryBuilder.Limit(20)
	}

	if params.Offset > 0 {
		subQueryBuilder = subQueryBuilder.Offset(params.Offset)
	}

	if params.Cursor > 0 {
		subQueryBuilder = subQueryBuilder.Where(sq.Lt{"r.id": params.Cursor})
	}

	queryBuilder := repo.db.squirrel.
		Select("r.id", "r.filter_status", "r.rejections", "r.indexer", "r.filter", "r.protocol", "r.title", "r.torrent_name", "r.size", "r.timestamp",
		"ras.id", "ras.status", "ras.action", "ras.type", "ras.client", "ras.filter", "ras.rejections", "ras.timestamp",
		"(SELECT COUNT(*) FROM release) AS total_count").
		FromSelect(subQueryBuilder, "r")
```

PR #579, opened 2022-12-24
fix(indexers): TS regex

martylukyy:
Adjusted our regex with capturing the ID after `/` instead of `id=`. 
Tested this the new expression with 1180 announces of which all resulted in a match.

EDIT:
Tested this PR with a snapshot build aswell. Successfully fetching files now again.


zze0s (approved):
🎅

PR #582, opened 2022-12-25
fix(indexers): fix AR and TS (GER) regex patterns

martylukyy:
The indexer definition for AR resulted in `line not matching expected regex pattern` for all announces for unknown reason.
Recreated the regex to a similar but much more rudimentary fashion.

The indexer definition for TS (GER) resulted in `line not matching expected regex pattern` when there were no tags existing in the announce. 

AR is successfully tested in a live env and in regex101 with 1069 multi-line announces and
TS (GER) got successfully tested through mock with an announce previously resulting in an error.


zze0s (approved):
Thanks as always!

Generally speaking `+` is enough to capture instead of `*` but no big deal for this.

PR #584, opened 2022-12-26
feat(web): added missing placeholders

s0up4200:
Added placeholders where there weren't any.

Some notable ones:

- Warn that label must exist in Deluge in order to work in `Deluge Label placeholder`.
- Higher number = higher prio in `Priority placeholder`.
- Emphasize that 0 is infinite in `Max downloads placeholder`.
- Emphasize the use of seconds in `Delay placeholder`.


zze0s (approved):
Great improvement!

PR #586, opened 2022-12-28
fix(indexers): bad ncore torrent URL

s0up4200:
Fixed the URL change initially made in the [baseurl override commit](https://github.com/autobrr/autobrr/commit/25a165b764ace7937ee8b0786ecd9bcd4b86a939#diff-8f417b2cd8914a5cdbdf14faaae4938e7f50de76992015ca2512204d3d0871a4)
An ncore user reported getting 404s after upgrading to v1.12.0

Untested, but should be good. The guy reporting it checked and confirmed that the URL is correct.


zze0s (approved):
This should work a lot better! 😄

PR #594, opened 2022-12-30
feat(indexers): Add DocsPedia.world

nelsonpecora:
This PR adds DocsPedia.world as an indexer

ref #1


nelsonpecora:
Manually testing this, but it seems like something wrong with the IRC config.

nelsonpecora:
The IRC config seems correct. I'll retest in case there was a ghost connection.

zze0s (approved):
Nice addition!

PR #597, opened 2022-12-30
feat(http): improve invalid login handling

zze0s:
Improve invalid login handling.

### DIFF
--- a/internal/http/auth.go
+++ b/internal/http/auth.go
@@ -75,7 +75,7 @@ func (h authHandler) login(w http.ResponseWriter, r *http.Request) {
 
 	_, err := h.service.Login(ctx, data.Username, data.Password)
 	if err != nil {
-		h.log.Error().Err(err).Msgf("invalid login [%s] from: %s", ReadUserIP(r))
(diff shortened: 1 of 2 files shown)


KyleSanderson (approved):

internal/http/auth.go
```diff
@@ -75,7 +75,7 @@ func (h authHandler) login(w http.ResponseWriter, r *http.Request) {
 
 	_, err := h.service.Login(ctx, data.Username, data.Password)
 	if err != nil {
-		h.log.Error().Err(err).Msgf("invalid login [%s] from: %s", ReadUserIP(r))
+		h.log.Error().Err(err).Msgf("Auth: Failed login attempt username: [%s] password: [%s] ip: %s", data.Username, data.Password, ReadUserIP(r))
```
vktr, line 78: I wouldn't recommend logging the password even for failed attempts - odds are that the user is providing a password used _somewhere_, and having that in the logs is almost as bad as having the autobrr password in it

zze0s: Yeah agree. Should've kept it without :D will update!

PR #599, opened 2022-12-31
fix(database): fix filter regression from clean-up

KyleSanderson:
So, as we're now painfully aware from https://github.com/autobrr/autobrr/pull/569 squirrel does not support subqueries as well as one would expect. There's some support, if you use an Alias on a column to provide a window, but this only supports params in either the window, another window, or a parent function, otherwise squirrel will misplace them. As squirrel is now in "maintenance" mode, I don't expect this situation to improve unfortunately.

So, lets revert this (hopefully) one regression.


zze0s (approved):
This seem to fix it!

Lets look at Squirrel alternatives..

PR #601, opened 2022-12-31
feat(web): make PWA compatible

s0up4200:
This should make it so when using "Add to Home Screen" on iOS it will show as a regular app with the autobrr icon. It will also show splash screens with the autobrr logo when clicking on the "app".

For future reference:
- Does not work when behind basic auth

~~- Does not currently load splash on IOS, but Chrome does it on Android.~~
~~- Currently directs you to autobrr.domain.tld/static when clicking the app icon.~~ fixed

image
image


zze0s:
Replaced by #793 since the move from CRA to Vite.

PR #606, opened 2023-01-01
feat(web): switch buttons on irc settings page

martylukyy:
Since people are often missing the part where they have to double-check their IRC settings and enable the network, 
i put switches in front of the IRC network name.
I know it looks a bit cluttered, but I couldn't find a better solution to this problem and I think implementing this
will be a bigger UX gain than it making the UI look cluttered.

image


zze0s (approved):
🐰

PR #608, opened 2023-01-01
enhancement(web): gray color for disabled checkboxes in dark mode

martylukyy:
Disabled checkboxes in dark mode are now the same color as disabled NumberFields in dark mode.

image


zze0s (approved):
🚀

PR #611, opened 2023-01-02
fix(web) reversed onChange for priority numberfield

s0up4200:
Unable to adjust priority within filters with the following code present, so removed it for now.
```js
            onChange={event => {
              // safeguard and validation if user removes the number
              // it will then set 0 by default. Formik can't handle this properly
              if (event.target.value == "") {
                form.setFieldValue(field.name, 0);
                return;
              }

              form.setFieldValue(field.name, event.target.value);
            }}
          />
          {meta.touched && meta.error && (
            <div className="error">{meta.error}</div>
          )}
```
image

### DIFF
--- a/web/src/components/inputs/input.tsx
+++ b/web/src/components/inputs/input.tsx
@@ -249,16 +249,6 @@ export const NumberField = ({
             )}
             placeholder={placeholder}
             disabled={disabled}
-            onChange={event => {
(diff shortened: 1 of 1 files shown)


zze0s (approved):
Lets see what we do in the future to fix this properly. Migrate from Formik or something.

Thx!

PR #624, opened 2023-01-06
fix(indexers): subsplease regex

martylukyy:
A user on discord found an issue with the SubsPlease regex not matching batches.
To fix the issue the releaseTags capture group is now optional.

Making the releaseTags group optional proposed another issue,
where the torrentName capture group would consume the content of the releaseTags group aswell.
This got fixed by setting the quantifier of the torrentName capture group lazy/ungreedy.

After setting this quantifier lazy/ungreedy the torrentName capture group got ended early by the torrentSize capture group.
That issue got fixed by creating a more specific regex for the torrentSize group.

Additionally I decided to move the releaseTags capture group out of the torrentName group in order to create more exact matches from the names of nyaa and the SubsPlease site itself.

Unfortunately i only had 4 announces to test this.
Should there be any concerns about the new regex you can leave it open
and I will check the regex from time to time after having gathered more announces.

https://regex101.com/r/M8nlhM/2

### DIFF
--- a/internal/indexer/definitions/subsplease.yaml
+++ b/internal/indexer/definitions/subsplease.yaml
@@ -44,7 +44,7 @@ irc:
   parse:
     type: single
     lines:
-      - pattern: '\[Release\] (.*(SubsPlease).*.(mkv)) \((.*)\) - .* - (.*)'
+      - pattern: '\[Release\] (.*(SubsPlease).*?)\.?(mkv)? \((\d+.?\d*[KMGTP]?B)\) - .* - (.*)'
         vars:
           - torrentName
           - releaseGroup


zze0s (approved):

OakyTree:
> Unfortunately i only had 4 announces to test this. Should there be any concerns about the new regex you can leave it open and I will check the regex from time to time after having gathered more announces.

@martylukyy 

I have an archive of all their announces going back at least 6 months you could use to test matching. Forward compatibility isn't guaranteed ofc, but it gives you a little more to test against than what you'll get same day watching IRC.

[subsplease_announces.txt](https://github.com/autobrr/autobrr/files/11357724/subsplease_announces.txt)

PR #625, opened 2023-01-06
feat(indexers): add SkipTheCommericals

zze0s:
Add STC.

Playground: https://regex101.com/r/wP53rX/2

Ref #1.


s0up4200 (approved):

martylukyy (approved):
Looks good.
For the next time i would probably do `.*` instead of `.+` in case for some unknown reason the type capture group is empty for example.
With `.*` it will not error in case the parenthesis are empty. `.+` will result in `line not matching expected regex pattern`.

PR #629, opened 2023-01-07
feat(indexers): improve RED and OPS artist parsing

zze0s:
This PR should improve the artist parsing for RED and OPS.

From `Artist - Album name` to now catch the whole `Artist - Albumname [2008] [Single] - FLAC / Lossless / Log / 100% / Cue / CD`. This should properly parse artist and album name. The other values were working good already.

And with this change we remove the workaround for Lidarr where we added year to the basic release name.

### DIFF
--- a/internal/domain/filter.go
+++ b/internal/domain/filter.go
@@ -376,11 +376,11 @@ func (f Filter) CheckFilter(r *Release) ([]string, bool) {
 	}
 
 	if len(f.Artists) > 0 && !contains(r.Artists, f.Artists) {
-		r.addRejectionF("artists not matching. got: %v want: %v", r.TorrentName, f.Artists)
+		r.addRejectionF("artists not matching. got: %v want: %v", r.Artists, f.Artists)
 	}
 
 	if len(f.Albums) > 0 && !contains(r.Title, f.Albums) {
-		r.addRejectionF("albums not matching. got: %v want: %v", r.TorrentName, f.Albums)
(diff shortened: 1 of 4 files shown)


internal/domain/filter.go
```diff
@@ -375,11 +375,11 @@ func (f Filter) CheckFilter(r *Release) ([]string, bool) {
 		r.addRejectionF("tags unwanted. got: %v want: %v", r.Tags, f.ExceptTags)
 	}
 
-	if len(f.Artists) > 0 && !containsFuzzy(r.TorrentName, f.Artists) {
+	if len(f.Artists) > 0 && !contains(r.Artists, f.Artists) {
 		r.addRejectionF("artists not matching. got: %v want: %v", r.TorrentName, f.Artists)
 	}
 
-	if len(f.Albums) > 0 && !containsFuzzy(r.TorrentName, f.Albums) {
+	if len(f.Albums) > 0 && !contains(r.Title, f.Albums) {
```
KyleSanderson, line 382: Does Subtitle ever contain anything for tunes?

martylukyy (approved):
Made the separator and the tags group for red optional in case there might be an announce without tags.
All the captures groups for OPS were already being made optional in this PR so looks good!
🚀

internal/domain/filter.go
zze0s: Not that I know!

PR #633, opened 2023-01-08
feat(downloadclients): qBit add rule UploadSpeedThreshold

### LINKED ISSUE #574: [FEATURE REQ] Upload speed check for filters.
A new option to send clients to qbit based on the network speed. 

if above 90mB/s do not send any new torrents. If below pass filter checks.

Scripting it on my own in python it no worky.

zze0s:
Add new rule check for qBittorrent to check if the total transfer speed is below set threshold.

It works the same way as the Download Speed Threshold but upload instead.

Example: Upload speed threshhold set to 90000 (90MB/s)

If total transfer speed of client is less than 90000 it will add a new torrent if "ignore slow torrents" is enabled in the download client settings in autobrr.

If it's above 90000 then it will reject and skip adding.

Resolves #574

### DIFF
--- a/internal/action/qbittorrent.go
+++ b/internal/action/qbittorrent.go
@@ -2,6 +2,7 @@ package action
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/autobrr/autobrr/internal/domain"
 	"github.com/autobrr/autobrr/pkg/errors"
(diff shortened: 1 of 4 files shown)


s0up4200 (approved):
Looks good 👍

PR #635, opened 2023-01-08
fix(indexers): IPT regex

s0up4200:
https://regex101.com/r/YtdL46/1


martylukyy (approved):
Quantifier for the first capture group is now lazy/ungreedy to prevent it from consuming too much characters if a release begins with `[SomeString]`.
Fixed the possibility of the regex failing at the optional `FREELEECH` group.
Looks good to send it.

zze0s (approved):

PR #639, opened 2023-01-09
feat(irc): add with network password

zze0s:
Add network from Indexer setup with network password. Used by Aither and custom indexers.


s0up4200 (approved):
Test was successful here 👍

PR #642, opened 2023-01-09
feat(indexers): add funfile

s0up4200:
https://regex101.com/r/LO2ET8/1- updated with fresh announces.

### DIFF
--- a/internal/indexer/definitions/funfile.yaml
+++ b/internal/indexer/definitions/funfile.yaml
@@ -5,7 +5,7 @@ identifier: funfile
 description: Welcome to funfile.org, where files are fun!
 language: en-us
 urls:
-  - https://funfile.org
+  - https://funfile.org/
 privacy: private
 protocol: torrent
 supports:
@@ -65,8 +65,10 @@ irc:
           - "[Scene] Uploaded 33 secs after pre"
         pattern: '\[(Scene)\]\s+Uploaded (.*) after pre'
         vars:
-          - scene
+          - origin
           - preTime
 
     match:
       torrenturl: "/download.php?{{ .torrentId }}/{{ .passkey }}/{{ .torrentName }}.torrent"
+      encode:
+        - torrentName
--- a/internal/indexer/definitions/funfile.yaml
+++ b/internal/indexer/definitions/funfile.yaml
@@ -51,7 +51,8 @@ irc:
     type: multi
     lines:
       - test:
-          - '[New Shit] S.W.A.T.Firefight.2011.PROPER.NTSC.MULTi.DVDR-BonG [Movies] @ https://www.funfile.org/details.php?id=131267 [size] [freeleech]'
+          - "[New Shit] Wiseguy.S04E01.720p.WEB.h264-SKYFiRE [TV/Web-HD] @ https://www.funfile.org/details.php?id=1662804 [2 GB] [FL]"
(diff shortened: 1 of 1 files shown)


internal/indexer/definitions/funfile.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+      - test:
+          - "[Scene] Uploaded 33 secs after pre"
+        pattern: '\[(Scene)\]\s+Uploaded (.*) after pre'
+        vars:
+          - scene
+          - preTime
+
+    match:
+      torrenturl: "/download.php?{{ .torrentId }}/{{ .passkey }}/{{ .torrentName }}.torrent"
```
zze0s, line 72: [This line in the .tracker file](https://github.com/autodl-community/autodl-trackers/blob/master/trackers/FunFile.tracker#L90) with `varenc` means it does a url encoding of the variable behind the scenes. They are using the torrent name in the url and it must be made sure to be safe for url use.

We support that by doing [it like here](https://github.com/autobrr/autobrr/blob/develop/internal/indexer/definitions/torrentleech.yaml#L72-L73).

```yml
torrenturl: "/download.php?{{ .torrentId }}/{{ .passkey }}/{{ .torrentName }}.torrent"
encode:
  - torrentName
```

internal/indexer/definitions/funfile.yaml
```diff
@@ -0,0 +1,72 @@
...
+          - baseUrl
+          - torrentId
+          - torrentSize
+          - freeleech
+
+      - test:
+          - "[Scene] Uploaded 33 secs after pre"
+        pattern: '\[(Scene)\]\s+Uploaded (.*) after pre'
+        vars:
+          - scene
```
zze0s, line 68: This can be set to `origin`.

internal/indexer/definitions/funfile.yaml
```diff
@@ -0,0 +1,72 @@
+---
+#id: funfile
+name: FunFile
+identifier: funfile
+description: Welcome to funfile.org, where files are fun!
+language: en-us
+urls:
+  - https://funfile.org
```
zze0s, line 8: End with slash to keep it uniform with other definitions.

internal/indexer/definitions/funfile.yaml
```diff
@@ -0,0 +1,72 @@
+---
+#id: funfile
+name: FunFile
+identifier: funfile
+description: Welcome to funfile.org, where files are fun!
```
zze0s, line 5: I usually grab the description from [this page](https://wiki.servarr.com/prowlarr/supported-indexers) but looks like there's no entry for them so leave as is.

zze0s (changes requested):
Some minor things :+1:

s0up4200:
> Some minor things 👍

Fixed 👍

internal/indexer/definitions/funfile.yaml
```diff
@@ -0,0 +1,76 @@
...
+language: en-us
+urls:
+  - https://funfile.org/
```
s0up4200, line 8: Should this be changed to https://www.funfile.org/ since they redirect to that? The announces use https://www.funfile.org/ as well.

zze0s: Yeah that is probably wise to change

zze0s (approved):

PR #645, opened 2023-01-10
fix(filters): music type don't disable when perfect flac is selected

zze0s:
Music type don't disable when perfect flac is selected.


s0up4200 (approved):
Tested, works as expected 👍 
image

PR #646, opened 2023-01-11
fix(indexers): funfile single line parsing

s0up4200:
Line 2 of their announce is only listed when there is a scene release.
~~This regex should solve that.~~

https://regex101.com/r/1JI657/1

### DIFF
--- a/internal/indexer/definitions/funfile.yaml
+++ b/internal/indexer/definitions/funfile.yaml
@@ -48,7 +48,7 @@ irc:
       help: NickServ password
 
   parse:
-    type: multi
+    type: single
     lines:
       - test:
           - "[New Shit] Wiseguy.S04E01.720p.WEB.h264-SKYFiRE [TV/Web-HD] @ https://www.funfile.org/details.php?id=1662804 [2 GB] [FL]"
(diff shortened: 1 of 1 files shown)


s0up4200:
Me and Marty have tested it with the mockindexer and it just hangs waiting for a 2nd line to come through every time.
Even setting the second one to `ignore: true` (https://github.com/autobrr/autobrr/pull/641/) does not seem to work.

After announcing the first line:

`{"level":"trace","module":"irc","network":"localhost","module":"announce_processor","time":"2023-01-11T21:03:12+01:00","message":"announce: queued line: New Torrent Announcement: <PC :: Iso>  Name: debian live 10 6 0 amd64 standard iso uploaded by 'Anonymous' -  http://localhost:3999/torrent/1"}
{"level":"trace","module":"irc","network":"localhost","module":"announce_processor","time":"2023-01-11T21:03:12+01:00","message":"announce: process line: New Torrent Announcement: <PC :: Iso>  Name: debian live 10 6 0 amd64 standard iso uploaded by 'Anonymous' -  http://localhost:3999/torrent/1"}`

Nothing happens until a second line is announced. The second line can be anything. It just waits for a second line since it's set to `type: multi` I guess?

The only solution I see right now is changing from `type: multi` to `type: single` and not capture `origin` and `preTime`.

martylukyy:
Quick summary about what was discussed:

The problem here is that we currently have no way of saying autobrr that there is a 2nd announce line which is optional.
Currently autobrr will always check pattern 1 2 1 2 1 2... regardless of the `ignore` flag.
The `ignore` flag only tells it to not parse anything from it if i understood this correctly.
Correct me if I am wrong here though.

In this example this results the announce processor to go out of sequence and/or missing single line announces.
This has been confirmed by testing with the mock indexer.

After looking through this with soup i can only think of two solutions this is happening:

The first and cleaner solution would be to implement a `optional` flag similar to the `ignore` flag.
In case the regex pattern is not matching the second announce line in this example here
it could go over and check it with the first pattern.
I don't know how much work it would be to implement this, nor if this even possible, or if this is a good idea.
Additionally i haven't thought thoroughly about what side effects this could have 
but I can't see any obvious problems with this approach.

The second and dirty solution would be to just wing it, set `type: single` and just ignore the second line completely.
This would result in us not being able to capture `origin` and `preTime` and 
we would be getting quite a substantial amount of `line not matching expected regex pattern` errors in the logs.

In case you wanna know my opinion about it or what i would do - i prefer to let ze0s decide what he wants to do here.

zze0s (approved):
Great! 

Lets keep this in mind for later with a proper fix for full optional announce lines. Funfile is the only one which uses it.

PR #647, opened 2023-01-13
enhancement(feeds): torznab freeleech percentage

michelheusschen:
The current implementation for Torznab feed uses the `downloadvolumefactor` attribute to set the freeleech status based on the attribute value being `"0"`. However, this does not follow the [specification](https://torznab.github.io/spec-1.3-draft/revisions/1.2-Torznab-Download-Upload-Factors.html) where it is outlined how `downloadvolumefactor` has a decimal value.

This PR aims to solve this problem and extend the functionality by parsing the value as a decimal number and setting the freeleech percentage accordingly. Invalid values are ignored, but a debug message is logged. For the freeleech percentages 25, 50, 75 and 100 an additional bonus is added, but it is unclear to me how bonuses are further used.

If any changes are required, please let me know.

### DIFF
--- a/internal/feed/torznab.go
+++ b/internal/feed/torznab.go
@@ -2,6 +2,7 @@ package feed
 
 import (
 	"context"
+	"math"
 	"sort"
 	"strconv"
 	"time"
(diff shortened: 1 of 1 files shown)


zze0s (approved):
Hi! I wasn't aware of it being an issue so thanks a lot for your PR fixing this! 

Looks simple and good. As for bonus, it's something that will be fully implemented in the future for the filters and irc indexers etc.

PR #648, opened 2023-01-13
enhancement(web): filters list mobile friendly

s0up4200:
## Make Filters page dynamically change based on screen width
I know this isn't what you ultimately want[^1], but I think it's a massive improvement to the current version. Even if it's just temporary.

This does not change how it looks on a regular computer screen except for 1p of vertical padding on the rows.

Demo of how it changes from wide to narrow: https://i.imgur.com/c02tqNo.mp4

### Current:
old

### Proposed change:
new

[^1]: I won't get hurt if this gets yeeted 😉


zze0s (approved):
Great interim solution until properly fixed! :clap:

Issue #650, opened 2023-01-14
Unable to clear log score once set


Audionut:
Pretty much the title.

You can leave the log score empty when creating a filter just fine, but if you save a filter with a log score, you cannot then clear the log score and save the filter.

Error
{"message":"json: cannot unmarshal string into Go struct field Filter.log_score of type int"} 

edit: Same problem when trying to clear 'max downloads'.

s0up4200:
What version are you on?

Audionut:
1.15.0

PR #651, opened 2023-01-14
fix(filters): Set NumberFields to required

s0up4200:
To avoid errors when user clears the field and tries to save. 

Ref https://github.com/autobrr/autobrr/issues/650

### DIFF
--- a/web/src/screens/filters/action.tsx
+++ b/web/src/screens/filters/action.tsx
@@ -199,10 +199,14 @@ const TypeForm = ({ action, idx, clients }: TypeFormProps) => {
               <NumberField
                 name={`actions.${idx}.limit_download_speed`}
                 label="Limit download speed (KiB/s)"
+                placeholder="Takes any number (0 is no limit)"
+                min={0} required={true}
               />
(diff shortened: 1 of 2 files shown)


Audionut:
I edited the issue to report that the same problem also occurs with 'max downloads'.

Haven't tested any other input boxes with a similar style, but it may be worth doing a sanity check on those, if there are any.

Cheers.

zze0s:
> I edited the issue to report that the same problem also occurs with 'max downloads'.
> 
> Haven't tested any other input boxes with a similar style, but it may be worth doing a sanity check on those, if there are any.
> 
> Cheers.

It's a known "issue". 0 Is default values for the number fields, which means it will only check if 1 or greater. 0 will be ignored. The issue is with the form library `Formik` doing something weird when you remove numbers and makes them into strings instead. The library isn't maintained anymore which is a shame so we don't know what we'll do about it. We can get around it with the fix as in this PR but it's not an actual fix for the core issue, just a workaround.

zze0s (approved):

PR #654, opened 2023-01-16
fix(indexers): animebytes regex

### LINKED ISSUE #653: animebytes pattern does not parse release group for translated printed media.
**Version**
1.13.0

**Describe the bug**
https://regex101.com/r/1XBznG/1
The release group of printed media is expected after Translated in the announce message. It is not captured by the 5th capture group, like it should be.

**To Reproduce**
1) Enable Trace Logs
2) Wait for english translated printed media announce.
3) Check trace logs.

**Expected behavior**
The release group of english translated printed media should be parsed by autobrr.

 - OS: Windows
 - Browser: Chrome
 - Version: 109.0.5414.74

martylukyy:
Fixes #653 

Regex will now successfully capture the release group for translated printed media.
Additionally you can now filter for Sub type in the releaseTags!

Fixed an edge case where miscapturing occurred when the torrentName contained the string `TV`

Disclaimer: To achieve this the Release group is now part of the releaseTags var which might lead to misparsed tags.
I have not run tests against this.

There is also the possibility of unwanted behaviour when using the `MATCH RELEASE TAGS` field under `Advanced` in autobrr.
But since we have the warning banner for releaseTags filtering i feel like this is justified.

In case you have any other concerns about this change feel free to point them out! :)

https://regex101.com/r/zmgPQu/4


varoOP:
Looks good, hope it's in the next release.

zze0s (approved):

PR #659, opened 2023-01-18
fix(web): filters list desktop and mobile friendly

martylukyy:
changed the resolution response from sm (640px) to md (768px) for the indexer column
compacted the filter lines while maintaining the padding to the border when a filter name spans over multiple lines 
span the filter name over the full width until the next div 
fixed a duplicate display property
adjusted the width of the indexer and sort listbox for show the full values. (widths for both are uniform) 
adjusted padding of indexer listbox entries


s0up4200 (approved):
Looks good Marty 👍

PR #661, opened 2023-01-19
enhancement(web): styled scrollbars for webkit browser

martylukyy:
This PR adds CSS properties for globally styled scrollbars with webkit enabled browsers.
Initially I wanted to do this with Tailwind but it turned out to be too much of an headache, so CSS it is now.

new

Same applies to horizontal scrollbar if they happen to appear somewhere.

Thanks to @s0up4200 for his help on implementing this enhancement!


s0up4200 (approved):
Glad you figured it out! Looks much better 👍

Issue #636, opened 2023-01-08
RSS Size Filter Issue


CarlitoGil:
**Version**
autobrr 1.14.0

**Describe the bug**
From the RSS feed, the torrent file size from `<enclosure ... length="39399"/>` is detected instead of `<size>26.34 GiB</size>`. 

**To Reproduce**
Some trackers use this format:
```
<item>
<title>Terminator Dark Fate 2019 2160p UHD BluRay SDR DD+ 5.1 HEVC</title>
<category>Peliculas</category>
<type>BDRip</type>
<resolution>2160p</resolution>
<size>26.34 GiB</size>
[...]
<pubDate>Sun, 08 Jan 2023 20:21:22 +0000</pubDate>
<enclosure url="https://lat-team.com/torrent/download/[...]" type="application/x-bittorrent" length="39399"/>
</item>
```
Using that example, JSON puts it inside `custom`:
```
"custom": {
	...
	"size": "26.34 GiB",
	...
}
```
**Expected behavior**
The feed is formatted clearly. The size should be detected from `<size>` instead of `<enclosure>`.

**Additional context**
The size from CinemaZ and other trackers is being detected quite well now (thanks!).

PR #667, opened 2023-01-22
fix(feeds): RSS improve torrent size parsing

ghoto:
When some RSS feeds have `Enclosure->Length` fixed size, autobrr fails to retrieve the size that's included in `Custom->Size`.

The previous logic used `custom->size` when `enclosure->length` was 0. But as stated in a previous issue, this `enclosure->length` can be a fixed length unrelated to the torrent. Therefore this value can be 0, fixed length, or the real one.

This modification will pick `custom->size` whenever is defined and it's bigger than `enclosures->length`.

https://github.com/autobrr/autobrr/issues/636

### DIFF
--- a/internal/feed/rss.go
+++ b/internal/feed/rss.go
@@ -154,11 +154,11 @@ func (j *RSSJob) processItem(item *gofeed.Item) *domain.Release {
 		rls.Uploader += v.Name
 	}
 
-	if rls.Size == 0 {
-		// parse size bytes string
-		if sz, ok := item.Custom["size"]; ok {
-			rls.ParseSizeBytesString(sz)
-		}
+	// Enclosure->length can be 0, a fixed size (see next additional parsing required), or the actual torrent size.
+	// Therefore when Custom->Size is defined it should be picked as the size. Otherwise we use Enclosure->Length
+	// This solves the issue of torrent size parsing of trackers like Lat-Team
+	if size, ok := item.Custom["size"]; ok {
+		rls.ParseSizeBytesString(size)
 	}
 
 	// additional size parsing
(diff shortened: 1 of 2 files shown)


internal/feed/rss.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/autobrr/autobrr/internal/release"
 	"github.com/autobrr/autobrr/pkg/errors"
 
+	"github.com/dustin/go-humanize"
```
KyleSanderson, line 15: Can you revert this?

internal/feed/rss.go
```diff
@@ -154,10 +155,12 @@ func (j *RSSJob) processItem(item *gofeed.Item) *domain.Release {
...
 	}
 
-	if rls.Size == 0 {
-		// parse size bytes string
-		if sz, ok := item.Custom["size"]; ok {
-			rls.ParseSizeBytesString(sz)
+	
+	// When custom->size and enclosures->size differ, pick the largest one.
+	if size, ok := item.Custom["size"]; ok {
+		s, err := humanize.ParseBytes(size)
```
KyleSanderson, line 161: Ditto.

KyleSanderson (changes requested):
Thank you very much for the contribution. Can you please revert the module changes (if there's a problem with the existing function, that should be fixed instead).

ghoto:
What do you mean with the existing function?

ghoto:
@KyleSanderson I moved the logic to `ParseSizeBytesString`. I fixed another potential bug on the function that would reset the size of the Release to 0 if the string passed to the function was malformed.

KyleSanderson (approved):

zze0s (commented):
If we are making changes then might be good to check for not only enclosure size difference but quite a bit of feeds set the enclosure size to a static `39399` on all items.

While it could happen that something actually have that exact size, it's imo unlikely.

zze0s (approved):

PR #669, opened 2023-01-23
fix(filters): allow negative priority again

martylukyy:
https://github.com/autobrr/autobrr/pull/605/files#diff-3e21b97dc9b916aed60fb39277bef67216f3746fa5c541dbbcc3c8096360a367L349
breaks the priority field when users decided to use negative values for priorities.
This PR reverts the `min={0}` attribute for the priority NumberField.


s0up4200 (approved):
I don't see a problem with letting people set negative values.
It currently throws an error for people who had a negative value set before the change was made in #605

PR #678, opened 2023-01-27
feat(filters): Double Upload

s0up4200:
### Add check for double upload

As of right now, Aither and Skipthecomericals announce double upload. There are a lot of indexers with RSS/Torznab support that has this as well.

Tested fine with mockindexer, waiting for real world examples to be announced with Aither. Will update this when that is confirmed working.

### DIFF
--- a/internal/database/sqlite_migrate.go
+++ b/internal/database/sqlite_migrate.go
@@ -69,6 +69,7 @@ CREATE TABLE filter
     min_size                       TEXT,
     max_size                       TEXT,
     delay                          INTEGER,
+    double_upload                  BOOLEAN,
     priority                       INTEGER   DEFAULT 0 NOT NULL,
     max_downloads                  INTEGER   DEFAULT 0,
     max_downloads_unit             TEXT,
@@ -384,6 +385,7 @@ var sqliteMigrations = []string{
 		tags              TEXT []   DEFAULT '{}' NOT NULL,
 		freeleech         BOOLEAN,
 		freeleech_percent INTEGER,
+        double_upload     BOOLEAN,
 		uploader          TEXT,
 		pre_time          TEXT
 	);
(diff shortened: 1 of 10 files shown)


internal/database/sqlite_migrate.go
```diff
@@ -384,6 +385,7 @@ var sqliteMigrations = []string{
 		tags              TEXT []   DEFAULT '{}' NOT NULL,
 		freeleech         BOOLEAN,
 		freeleech_percent INTEGER,
+        double_upload     BOOLEAN,
```
KyleSanderson, line 388: tab tab

KyleSanderson:
bit conflicted on this one, what about 2.5x and 3x upload?

s0up4200:
> bit conflicted on this one, what about 2.5x and 3x upload?

Right now there are only two indexers with IRC support that has double upload.

I'm not sure how to go about making it work for RSS/Torznab though, but to make it work for 2x, 3x, 4x and so on shouldn't be a problem I guess?

zze0s:
While this is nicely done, I'm not sure this is the right approach for this.

It's been on my mind for a while and partially done in code to move the freeleech to a new generic `bonus` model that can contain freeleech, 2x upload etc. But it's not yet set in stone.

Torznab does a similar thing but they do `downloadVolumeFactor` and `uploadVolumeFactor` which is a float/percent. In case of 2x upload that would be `uploadVolumeFactor: 2.0` or 200, or similar. But that only allows for one. 

For ultimate flexibility we would want to be able to set multiple values with operators like **less than**, **greater than**, **equals**. `< = >` and different values.

There's only a handful irc announcers that announce freeleech percent. 2 that do 2x upload like you mention here. Many do just freeleech. And for RSS feeds there's a lot of variants.

So something needs to be decided :smile:

s0up4200:
> While this is nicely done, I'm not sure this is the right approach for this.
> ...

Totally understandable. It was first and foremost a challenge, and I got it working which is good enough for me 😄 I'll close this one.

PR #680, opened 2023-01-29
enhancement(web): disable freeleech percent field if freeleech is enabled

s0up4200:
This should hopefully help avoid confusion on the users part since toggling the switch locks freeleech percent to 100. Seen it happen a few times where they enter 50-100 and then toggles the switch.

### DIFF
--- a/web/src/screens/filters/details.tsx
+++ b/web/src/screens/filters/details.tsx
@@ -546,7 +546,7 @@ export function Advanced({ values }: AdvancedProps) {
           <SwitchGroup name="freeleech" label="Freeleech" />
         </div>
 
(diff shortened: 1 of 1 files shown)


Audionut:
Some other suggestion.

"Match only freeleech and freeleech percent." > "Match only 100% freeleech or a freeleech percentage".

"eg. 50,75-100" > "eg. 50,75-99"

zze0s:
> Some other suggestion.
> 
> "Match only freeleech and freeleech percent." > "Match only 100% freeleech or a freeleech percentage".
> 
> "eg. 50,75-100" > "eg. 50,75-99" < edit that doesn't work.

Something like that is planned but there's no ETA.

zze0s (approved):

PR #681, opened 2023-01-29
fix(releases): check for duplicate tags

KyleSanderson:
Presently Audio, Codec, and Other can contain duplicate information if it's present in the announced tags. This results in a not great situation where we're not only storing duplicate information, but filtering on it as well.


zze0s (approved):
Creative solution :D

PR #683, opened 2023-01-29
feat(releases): add info url link to releases list

zze0s:
Save info url for release and show in releases list.

### DIFF
--- a/internal/indexer/definitions/animebytes.yaml
+++ b/internal/indexer/definitions/animebytes.yaml
@@ -67,7 +67,7 @@ irc:
           - "Best Visual Novel - Visual Novel  [2006] :: Game / PC / Unarchived / Hentai (Censored) || https://animebytes.tv/torrents.php?id=00000&torrentid=00000 || nukige || Uploaded by: Uploader"
           - "Artist Name - Album of awesome Music  [1991] :: MP3 / V0 (VBR) / CD || https://animebytes.tv/torrents2.php?id=00000&torrentid=000000 || ambient, folk || Uploaded by: Uploader"
           - "Awesome Series - TV Series  [2022] :: Web / MKV / h264 / 1080p / AAC 2.0 / Softsubs (Sub Group) / Episode 1 / Freeleech || https://animebytes.tv/torrents.php?id=00000&torrentid=000000 || || Uploaded by: Uploader"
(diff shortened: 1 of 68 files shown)


internal/indexer/definitions/animebytes.yaml
```diff
@@ -82,5 +82,6 @@ irc:
           - uploader
 
     match:
+      infourl: "/torrents.php?id={{ .torrentId }}"
```
martylukyy, line 85: This is wrong...
For non music torrents it is: https://animebytes.tv/torrents.php?id=00000&torrentid=0000000
For music torrents it is: https://animebytes.tv/torrents2.php?id=00000&torrentid=0000000
Where `id` is the ID for the torrent group and `torrentid` is for the torrent itself..

`torrentid` alone won't do it here unfortunatly..

internal/indexer/definitions/beyondhd.yaml
```diff
@@ -70,4 +70,5 @@ irc:
           - torrentId
 
     match:
+      infourl: "/details.php?id={{ .torrentId }}"
```
martylukyy, line 73: https://beyond-hd.me/details.php?id=000000 gives me 404..

zze0s: Ah yeah, it uses `/torrents/url-encoded-torrent-name.torrentID` so we'll have to encode that to work.

internal/indexer/definitions/btn.yaml
```diff
@@ -102,4 +102,5 @@ irc:
           - torrentId
 
     match:
+      infourl: "/torrents.php?id={{ .torrentId }}"
```
martylukyy, line 105: We capture the ID of the torrent itself here...
But `/torrents.php?id={{ .torrentId }}` takes you to the torrent group which would be the respective episode for example..
Correct would be https://broadcasthe.net/torrents.php?id=000000&torrentid=0000000

Hence this resulting in 404 aswell.

internal/indexer/definitions/btn.yaml
```diff
@@ -96,10 +96,12 @@ irc:
...
           - "[ https://XXXXXXXXX/torrents.php?id=7338 / https://XXXXXXXXX/torrents.php?action=download&id=9116 ]"
-        pattern: \[ .* \/ (https?:\/\/.*\/).+id=(\d+) \]
+        pattern: \[ (https?:\/\/.*\/).+id=(\d+) \/ https?:\/\/.*\/.+id=(\d+) \]
         vars:
           - baseUrl
+          - groupId
           - torrentId
 
     match:
+      infourl: "/torrents.php?id={{ .groupId }}"
```
martylukyy, line 106: We could extend the infourl here to make it open the info of the actual torrent instead of just opening the group here...
`/torrents.php?id={{ .groupId }}&torrentid={{ .torrentId }}`
Don't know if you care about that...

martylukyy (approved):
Looks good! Head of brr putting in the hours! 🚀

PR #684, opened 2023-02-01
feat(sqlite): commit WAL on startup

KyleSanderson:
Right now this needs to be manually run to actually have the pragma changed, the reason why is WAL prevents schema changes like this. I was getting terrible vacuum performance over vfs, then I started looking at the pragmas available for sqlite and stumbled onto this one.

https://blogs.gnome.org/jnelson/2015/01/06/sqlite-vacuum-and-auto_vacuum/
https://stackoverflow.com/questions/46010813/how-to-increase-the-size-limit-of-a-sqlite-database-to-140tb
https://phiresky.github.io/blog/2020/sqlite-performance-tuning/
https://sqlite.org/forum/forumpost/35a0c8cf32f5296a


zze0s:
The page size should probably be set here somewhere [/internal/database/sqlite.go#L21](https://github.com/autobrr/autobrr/blob/develop/internal/database/sqlite.go#L21).

And the vacuum should probably run as a task since that is useful for both sqlite and postgres.

KyleSanderson:
I've reverted the page_size change for now, committing the WAL at startup should be helpful for "less than ideal" environments.

zze0s (approved):

PR #687, opened 2023-02-02
fix(indexers): add default http baseurl for PTN

s0up4200:


zze0s (changes requested):
For these cases I think it's better to have both of them since the https might still be used for rss or other web things.

So add the http one as first, and then add back the https one as second. That way it will default to http.

The way it is currently you can also set whatever url you want in the indexer settings, but if there is none set it will default to the first url set above here.

zze0s (approved):

Issue #267, opened 2022-05-06
Possibility to communicate breaking changes to users between releases


stacksmash76:
Title, basically. If a user upgrades from version A to a later version B we should communicate possibly breaking changes via a sort of "one-time" modal which could be managed by the database.

PR #690, opened 2023-02-04
feat: show new updates in dashboard

### LINKED ISSUE #319: Notify on updates
Notify of available updates, with banner and in settings. Fetch on schedule, or manually.

## Future goals

- Auto update
- Update "in place" for non docker env.

zze0s:
Show new updates in dashboard.

image

Other work:

- add request logger
- minor api restructure

Resolves #319 and #267

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ web/build
 bin/
 log/
 dist/
+.run/
 # If needed, package-lock.json shall be added
 # manually using an explicit git add command.
 package-lock.json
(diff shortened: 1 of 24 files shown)


s0up4200 (commented):
Small suggestion front end wise.
image
I've written the front end for it, but unsure on how to do the backend bit.

zze0s:
> Small suggestion front end wise. image I've written the front end for it, but unsure on how to do the backend bit.

That would be nice! But to make it work properly it needs to be able to update the `config.toml` file (like some other features) and we currently have no way to do that.

I'll see how fast I can get that going since it will open up for other config/settings like logLevel.

zze0s:
Added ability to toggle updates check on and off from the settings! These changes to edit the config now makes it possible to add more settings like `logLevel` etc.

s0up4200 (approved):

PR #691, opened 2023-02-04
feat(config): allow config environment variables

KyleSanderson:
Viper has environment variable support built-in and will be used as the first source of information, before falling back to other resources. This permits containerized environments to overload config file settings by setting them in a docker-compose file. As an example, the environment variable of AUTOBRR_PORT can be used to change the port to another to align with health checks in the orchestrator of choice.


zze0s (changes requested):
Probably something that came with from testing, but not a great idea to comment out `port` ;) `logLevel` is fine but should be left as is imo.

zze0s (approved):

PR #694, opened 2023-02-05
fix(indexer): Fixed bad url for PTN

s0up4200:


internal/indexer/definitions/ptn.yaml
```diff
@@ -5,19 +5,19 @@ identifier: ptn
...
   - irc
   - rss
 source: custom
 settings:
-  - name: torrent_pass
+  - name: passkey
     type: secret
     required: true
-    label: Torrent pass
+    label: Passkey
```
zze0s, line 20: Lets keep name as torrent_pass to not break existing users but keep the new label. Set it in the torrentUrl as well

zze0s (changes requested):
Minimal but needed change

zze0s (approved):

PR #708, opened 2023-02-12
feat(transmissionbt): implement reannounce and max active rules

KyleSanderson:
Couple frustrations with qBittorrent... As expected, Transmission has always had way more features than qBittorrent (free disk space based on path, as an example). Based on the (upstream) renewed interest in Transmissionbt, the goal of the PR is to hit feature parity with our presently most well supported client (qBitorrent).


zze0s (approved):
Did a bit of refactoring and added rules checks for Max Active Downloads.

PR #713, opened 2023-02-13
fix(web): fixed baseurl for log downloads

s0up4200:
Right now you can't download logs if you're on a subfolder reverse proxy setup. It only works on `/`.
Reason: the Link component does not require `${baseUrl()}`.

image


s0up4200:
Feed.tsx does not require this change since it's using the `<a>` tag instead of `<Link>` `<a>` tags does need the `${baseUrl()}` part:
https://github.com/autobrr/autobrr/blob/12d05f074899e364dfb054451a730e2c7a75134d/web/src/screens/settings/Feed.tsx#L253-L260

zze0s (approved):

PR #715, opened 2023-02-14
feat(downloadclients): make fields required

martylukyy:
This PR adds field validation for the DownloadClientForms (Add and Update).

Because of this feature it also prevents users from adding a client without a name 
causing the added client to not show up in the client dropdown menu when adding an action in a filter.

Unfortunately i ran into a bug when adding a Porla client.
When clicking the `Create` button during adding a new client the `Auth token` field errors silently,
but it should show that the field is required. On the Update form however it works.
I couldn't figure out why this is happening maybe someone can take a look over this bug.


s0up4200:
Removed `validate={validate}` as it does not really serve a purpose for the download clients when using `required={true}`. It also does not work for the API fields, while `required={true}` does. A bonus is that it displays a pop up asking you to fill the field, compared to adding a red border, and text below the fields. It looks cleaner.

The only place I can think of that `validate={validate}` serves a better purpose is for the IRC command field when setting up indexers as they require you to replace default values with your own.

Demonstration of the change: https://i.imgur.com/eNWP5H5.mp4

zze0s (approved):
:+1:

PR #719, opened 2023-02-17
fix(web): reverted freeleech lock mechanism

s0up4200:
Reverted locking mechanism for the Freeleech toggle as it doesn't set `freeleechPercent` to 100 when doing so.


zze0s (approved):
:+1:

PR #725, opened 2023-02-21
fix(filters): quality multi select not closing

s0up4200:


zze0s:
I'll do a round of testing and see if this has any unwanted side effects.

zze0s (approved):
The main thing the long complicated code does is to create a new array with custom values and it then shows up in the dropdown.

Removing it will still have the option but it won't show in the list to deselect.

The `creatable` nature of these fields where the component is used is undocumented anyway and might not be used at all as far as we know, and likely not supported by the underlying parser except in maybe cases for resolution.

I'll approve and merge this and if anyone notices in the future we can deal with it then. This fixes the annoying bug with the dropdowns not closing when clicking outside and that is more important.

PR #726, opened 2023-02-22
chore(deps): update Go to 1.20

KyleSanderson:
https://go.dev/blog/go1.20

`
The new function [errors.Join](https://go.dev/pkg/errors#Join) returns an error wrapping a list of errors which may be obtained again if the error type implements the Unwrap() []error method.
`

`When [building a Go release from source](https://go.dev/doc/install/source), Go 1.20 requires a Go 1.17.13 or newer release. In the future, we plan to move the bootstrap toolchain forward approximately once a year. Also, starting with Go 1.21, some older operating systems will no longer be supported: this includes Windows 7, 8, Server 2008 and Server 2012, macOS 10.13 High Sierra, and 10.14 Mojave. On the other hand, Go 1.20 adds experimental support for FreeBSD on RISC-V.`

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -6,7 +6,7 @@ node_modules/
 config.toml
 .goreleaser.yml
 Dockerfile
-Dockerfile.ci
+ci.Dockerfile
 docker-compose.yml
 README.md
 bin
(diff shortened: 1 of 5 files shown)


zze0s (changes requested):
Should update `Dockerfile` and `Dockerfile.ci` as well.

zze0s (approved):

PR #758, opened 2023-03-14
fix(filters): remove requirement for numberfields

KyleSanderson:


s0up4200:
We already talked about this on Discord, but gonna leave it here just for the sake of clarity.

The reason for the `required{true}` is to avoid the following error when a user clears the field and hits save (in a scenario where the user has set a value in a field that is blank by default):

You can test this with the Delay field in `v1.19.0`:
```json
{"message":"json: cannot unmarshal string into Go struct field Filter.log_score of type int"} 
```

There is a better approach than using `required{true}`, and that is to reimplement an onChange for the NumberField component. Ze0s introduced in a previous PR, but it made it impossible to change from 0 to a different number:
https://github.com/autobrr/autobrr/blob/5972d421d805ee6c854ad7680833ec47a4cfb1c7/web/src/components/inputs/input.tsx#L252-L261

The reason it did not work is because the frontend attempts to send it as a string, and the backend expects an integer.

The following will fix the error with the previous code by using `parseInt()` to convert the input value to an integer:
```ts
form.setFieldValue(field.name, parseInt(event.target.value));
```

I removed all of the `required={true}` that were present in `<NumberField>`  components in `/screens/filters/`, but kept `min={0}` and`max={100}` for`log_score` as they serve a good purpose for that field.

s0up4200 (approved):
Monitored my DB with TablePlus while testing and can't find anything wrong.

zze0s (approved):

PR #759, opened 2023-03-14
fix(filters): remove requirement for Max Downloads.

KyleSanderson:


s0up4200:
Will be handled in #758

PR #763, opened 2023-03-15
fix(feeds): implement max_age for torznab/newznab

KyleSanderson:
Speculative fix for Max Age not working for items retrieved from Torznab / Newznab.

### DIFF
--- a/internal/feed/newznab.go
+++ b/internal/feed/newznab.go
@@ -80,7 +80,7 @@ func (j *NewznabJob) process(ctx context.Context) error {
 		if j.Feed.MaxAge > 0 {
 			if item.PubDate.After(time.Date(1970, time.April, 1, 0, 0, 0, 0, time.UTC)) {
 				if !isNewerThanMaxAge(j.Feed.MaxAge, item.PubDate.Time, now) {
-					return nil
+					continue
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


internal/feed/newznab.go
```diff
@@ -75,8 +75,16 @@ func (j *NewznabJob) process(ctx context.Context) error {
 	}
 
 	releases := make([]*domain.Release, 0)
-
+	now := time.Now()
 	for _, item := range items {
+		if j.Feed.MaxAge > 0 {
+			if item.PubDate.After(time.Date(1970, time.April, 1, 0, 0, 0, 0, time.UTC)) {
+				if !isNewerThanMaxAge(j.Feed.MaxAge, item.PubDate.Time, now) {
+					return nil
```
zze0s, line 83: Shouldn't this just be `continue` rather than return?

internal/feed/torznab.go
```diff
@@ -76,8 +76,16 @@ func (j *TorznabJob) process(ctx context.Context) error {
 	}
 
 	releases := make([]*domain.Release, 0)
-
+	now := time.Now()
 	for _, item := range items {
+		if j.Feed.MaxAge > 0 {
+			if item.PubDate.After(time.Date(1970, time.April, 1, 0, 0, 0, 0, time.UTC)) {
+				if !isNewerThanMaxAge(j.Feed.MaxAge, item.PubDate.Time, now) {
+					return nil
```
zze0s, line 84: Same as above.

internal/feed/newznab.go
KyleSanderson: most certainly

zze0s (approved):

PR #764, opened 2023-03-16
build: Implement autogofmt

KyleSanderson:


KyleSanderson:
https://github.com/actions/runner/issues/2033
https://github.com/iamnotaturtle/auto-gofmt/issues/4

zze0s:
It does not use gofmt directly, and instead uses https://github.com/mvdan/gofumpt which say:

> Enforce a stricter format than gofmt, while being backwards compatible. That is, gofumpt is happy with a subset of the formats that gofmt is happy with.

I have gofmt run on save and some of the things like comments with space is not happening on my end. I have also never seen this syntax for permissions before: https://github.com/autobrr/autobrr/pull/764/files#diff-54c7c1af5fa8d5db4dc49f0e8e80e93ba2b1183ba4d5c9e2e5729e6deae6a3cdR275

```diff
-if err := os.WriteFile(file, []byte(output), 0644); err != nil {
+if err := os.WriteFile(file, []byte(output), 0o644); err != nil {
        return errors.Wrap(err, "could not write config file: %s", file)
}
```

All in all not a super big fan of this. I get the point of it helping to have one style and not having to format correctly before commit, but that's about it.

PR #765, opened 2023-03-16
chore(http): set log level to trace

KyleSanderson:


internal/http/middleware.go
```diff
@@ -61,22 +62,24 @@ func LoggerMiddleware(logger *zerolog.Logger) func(next http.Handler) http.Handl
...
-						"proto":      r.Proto,
-						"method":     r.Method,
-						"user_agent": r.Header.Get("User-Agent"),
-						"status":     ww.Status(),
-						"latency_ms": float64(t2.Sub(t1).Nanoseconds()) / 1000000.0,
-						"bytes_in":   r.Header.Get("Content-Length"),
-						"bytes_out":  ww.BytesWritten(),
-					}).
-					Msg("incoming_request")
+				if !strings.Contains("/api/healthz/liveness|/api/healthz/readiness", r.URL.Path) {
```
zze0s, line 65: Is there a good reason to ignore logging from these endpoints?

KyleSanderson: When they're run every couple seconds from an external probe, yeah. In the UI this log is very spammy (it literally just says incoming_request).

zze0s (approved):

PR #766, opened 2023-03-16
fix(actions): reset client_id on action swap

s0up4200:


web/src/screens/filters/action.tsx
```diff
@@ -106,14 +107,51 @@ interface TypeFormProps {
...
+  const [prevActionType, setPrevActionType] = useState<string | null>(null);
+  
   switch (action.type) {
   case "TEST":
+    useEffect(() => {
+      if (prevActionType !== null) {
+        resetClientField(action, idx, prevActionType);
+      }
+      setPrevActionType(action.type);
+    }, [action.type, idx, setFieldValue]);
```
zze0s, line 142: This can probably be unified below line `133` to not have to repeat for each case. I might be wrong, but that is before rendering.

s0up4200: I actually tried something like this, but I think I placed it below L135 instead. Seems to work fine when putting it above the switch!

zze0s (approved):

PR #768, opened 2023-03-17
chore(indexers): properly capitalize Aither network

KyleSanderson:


s0up4200 (approved):
Steps I took:
1. Setup aither in develop so it has lowercase
2. Switch to this branch
3. Network is still called aither.cc, but matches fine with a test action
4. Recreate indexer
5. Network is called Aither.cc now

PR #772, opened 2023-03-19
feat(web): preserve sort order on Filters list

s0up4200:
## Preserve sortOrder using localStorage on Filters page

Store the user's chosen sortOrder in the browser's localStorage to maintain their sorting preference even when the page is refreshed. Previously, the sortOrder would reset upon refreshing the Filters page.

### DIFF
--- a/web/src/screens/filters/list.tsx
+++ b/web/src/screens/filters/list.tsx
@@ -1,4 +1,4 @@
-import { Dispatch, FC, Fragment, MouseEventHandler, useReducer, useRef, useState } from "react";
+import { Dispatch, FC, Fragment, MouseEventHandler, useReducer, useRef, useState, useEffect } from "react";
 import { Link } from "react-router-dom";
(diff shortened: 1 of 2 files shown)


stacksmash76:
Why not use SettingsContext instead? https://github.com/autobrr/autobrr/blob/develop/web/src/utils/Context.ts#L46

Usage here https://github.com/autobrr/autobrr/blob/develop/web/src/screens/Logs.tsx#L308

s0up4200:
> Why not use SettingsContext instead? https://github.com/autobrr/autobrr/blob/develop/web/src/utils/Context.ts#L46
> 
> 
> 
> Usage here https://github.com/autobrr/autobrr/blob/develop/web/src/screens/Logs.tsx#L308

Will look into it, thanks for the pointers👍

zze0s (approved):
Nice!

Only remark is it could have been merged into SettingsContext to have a single one, but since this works it can be improved later on.

PR #773, opened 2023-03-19
feat(filters): regex validation for supported fields

s0up4200:
## Regex validation for regex supported fields

This PR adds regex validation to `Match releases`, `Except releases`, `Match release tags` and `Except release tags`.
Green and red checkmarks will display whenever the `Use regex`switches are toggled.

Added Formik Validation. You can no longer save when regex is invalid (if the regex switch is enabled)

Demonstration: https://cln.sh/QyLlr082

It _should_ check for Golang flavor, but would like someone to look that over.


zze0s (commented):
I think it makes more sense to create a new field like `RegexField` or similar that has the regex things in it, rather than adding very specific functionality to a common field.

s0up4200:
> I think it makes more sense to create a new field like `RegexField` or similar that has the regex things in it, rather than adding very specific functionality to a common field.

Done!

zze0s:
> > I think it makes more sense to create a new field like `RegexField` or similar that has the regex things in it, rather than adding very specific functionality to a common field.
> 
> Done!

Great! Further improvements could be to use the built in functionality in Formik Fields for validation. 

- https://formik.org/docs/guides/validation#field-level-validation
- https://www.digitalocean.com/community/tutorials/how-to-validate-a-login-form-with-react-and-formik

s0up4200:
> > > I think it makes more sense to create a new field like `RegexField` or similar that has the regex things in it, rather than adding very specific functionality to a common field.
> > 
> > 
> > Done!
> 
> Great! Further improvements could be to use the built in functionality in Formik Fields for validation.
> 
>     * https://formik.org/docs/guides/validation#field-level-validation
> 
>     * https://www.digitalocean.com/community/tutorials/how-to-validate-a-login-form-with-react-and-formik

Should be good now 👍

zze0s (approved):
Made some improvements like discussed!

PR #776, opened 2023-03-19
feat(filters): navigate to edit page on filter creation

metonyms:
i'm not certain if there is an explict reason for the current behavior, but personally I always want to edit a filter immediately after creation.

i can adjust the commit message if needed, i wasn't sure if this was a fix or a feature /shrug


zze0s (approved):
Thanks for your PR!

None of the maintainers have an issue with this so we'll accept it!

PR #777, opened 2023-03-20
fix(filters): hide import dropdown when clicking outside it

s0up4200:
Obvisouly this was discovered 5 minutes after 1.20 was released.
Should be good now.

https://cln.sh/tB1jK60q

### DIFF
--- a/web/src/screens/filters/list.tsx
+++ b/web/src/screens/filters/list.tsx
@@ -4,7 +4,7 @@ import { toast } from "react-hot-toast";
 import { Listbox, Menu, Switch, Transition } from "@headlessui/react";
 import { useMutation, useQuery, useQueryClient } from "react-query";
 import { FormikValues } from "formik";
-import { useCallback } from "react";
+import { useCallback, useEffect } from "react";
 
 import { Tooltip } from "react-tooltip";
 
@@ -17,8 +17,7 @@ import {
   DocumentDuplicateIcon,
   EllipsisHorizontalIcon,
   PencilSquareIcon,
-  TrashIcon,
-  ExclamationTriangleIcon
+  TrashIcon
 } from "@heroicons/react/24/outline";
 
 import { queryClient } from "../../App";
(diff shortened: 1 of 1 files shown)


web/src/screens/filters/list.tsx
```diff
@@ -141,24 +140,49 @@ export default function Filters({}: FilterProps){
...
   const [showDropdown, setShowDropdown] = useState(false);
+  const dropdownRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    function handleClickOutside(event: { target: any; }) {
+      if (
+        dropdownRef.current &&
+        !(dropdownRef.current as any).contains(event.target) &&
+        event.target.parentElement.className !== "relative" &&
+        event.target.className !== "relative inline-flex items-center px-4 py-2 shadow-sm text-sm font-medium rounded-l-md text-white bg-blue-600 dark:bg-blue-600 hover:bg-blue-700 dark:hover:bg-blue-700 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-blue-500 dark:focus:ring-blue-500"
```
stacksmash76: Why not use the ID attribute instead? Ditto below.

web/src/screens/filters/list.tsx
```diff
@@ -141,6 +144,28 @@ export default function Filters({}: FilterProps){
...
   const [showDropdown, setShowDropdown] = useState(false);
+  const dropdownRef = useRef(null);
+  const addFilterRef = useRef(null);
+
+  useEffect(() => {
+    function handleClickOutside(event: { target: any }) {
+      if (
+        dropdownRef.current &&
+        !(dropdownRef.current as any).contains(event.target) &&
+        !(addFilterRef.current as any).contains(event.target)
```
stacksmash76, line 155: You forgot to check whether addFilterRef.current is null here.

web/src/screens/filters/list.tsx
```diff
@@ -141,6 +144,28 @@ export default function Filters({}: FilterProps){
...
+        setShowDropdown(false);
+      }
+    }
+
+    document.addEventListener("click", handleClickOutside);
+
+    return () => {
+      document.removeEventListener("click", handleClickOutside);
+    };
+  }, [dropdownRef, addFilterRef]);
```
stacksmash76, line 166: Pretty sure you don't have to add refs as dependencies in useEffect. If you're relying on VSCode warnings/hints, I'll tell you that the useEffect plugin is lying a good amount of time. Might be wrong in this case however.

stacksmash76 (changes requested):
Overall fine, could use a few changes.

PR #779, opened 2023-03-21
refactor(logs): improve log sanitization performance

s0up4200:
Improved regex handling, should redact everything properly without otherwise altering the log file.
Big improvements to the time it takes to sanitize the log file.


internal/http/logs.go
```diff
@@ -87,48 +92,139 @@ func (h logsHandler) files(w http.ResponseWriter, r *http.Request) {
...
-	inviteRegex   = regexp.MustCompile(`(Voyager autobot [\p{L}0-9]+ |Satsuki enter #announce [\p{L}0-9]+ |Millie announce |DBBot announce |ENDOR !invite [\p{L}0-9]+ |Vertigo ENTER #GGn-Announce [\p{L}0-9]+ |midgards announce |HeBoT !invite |NBOT !invite |Muffit bot #nbl-announce [\p{L}0-9]+ |hermes enter #announce [\p{L}0-9]+ |LiMEY_ !invite |PS-Info pass |PT-BOT invite |Hummingbird ENTER [\p{L}0-9]+ |Drone enter #red-announce [\p{L}0-9]+ |SceneHD \.invite |erica letmeinannounce [\p{L}0-9]+ |Synd1c4t3 invite |UHDBot invite |Sauron bot #ant-announce [\p{L}0-9]+ |RevoTT !invite [\p{L}0-9]+ |Cerberus identify [\p{L}0-9]+ )([\p{L}0-9]+)`)
 	nickservRegex = regexp.MustCompile(`(NickServ IDENTIFY )([\p{L}0-9!#%&*+/:;<=>?@^_` + "`" + `{|}~]+)`)
-	saslRegex     = regexp.MustCompile(`(--> AUTHENTICATE )([\p{L}0-9!#%&*+/:;<=>?@^_` + "`" + `{|}~]+)`)
+	saslRegex     = regexp.MustCompile(`(AUTHENTICATE )([\p{L}0-9!#%&*+/:;<=>?@^_` + "`" + `{|}~]+)`)
+
+	limeyInviteRegex       = regexp.MustCompile(`(LiMEY_ !invite\s+)([a-zA-Z0-9]+)(\s+\w+)`)
```
KyleSanderson, line 102: this should be an array, but even then the original match was likely faster?

internal/http/logs.go
```diff
@@ -87,48 +92,139 @@ func (h logsHandler) files(w http.ResponseWriter, r *http.Request) {
...
+	// Start the worker goroutines
+	for i := 0; i < numWorkers; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+
+			for {
+				// Read the next line from the file
+				fileMutex.Lock()
+				line, err := reader.ReadString('\n')
```
KyleSanderson, line 163: I think this may not be doing what you would expect. Writes will still not be ordered correctly. Going back to the single thread is cleaner and may improve overall times on these machines. The alternative is to create a channel array for your workers by idx, then receive on them by ranging over them until they're all closed.

internal/http/logs_sanitize_test.go
```diff
@@ -28,142 +31,150 @@ func TestSanitizeLogFile(t *testing.T) {
...
+			tmpFile, err := ioutil.TempFile("", "test-log-*.log")
+			if err != nil {
+				t.Fatal(err)
+			}
+			defer os.Remove(tmpFile.Name())
+
+			// Write the test case input to the temporary file
+			_, err = tmpFile.WriteString(testCase.input + "\n")
+			if err != nil {
+				tmpFile.Close()
```
KyleSanderson, line 151: why not defer close above?

internal/http/logs_sanitize_test.go
```diff
@@ -28,142 +31,150 @@ func TestSanitizeLogFile(t *testing.T) {
...
-			t.Fatal(err)
-		}
+			sanitizedData := buf.String()
 
-		// Check if the sanitized data matches the expected content
-		if string(sanitizedData) != testCase.expected {
-			t.Errorf("Sanitized data does not match expected data for input: %s\nExpected:\n%s\nActual:\n%s", testCase.input, testCase.expected, sanitizedData)
-		}
+			// Check if the sanitized data matches the expected content
+			if !strings.Contains(sanitizedData, testCase.expected+"\n") {
```
KyleSanderson, line 175: Check the length, HasPrefix, then check the last character with another conditional to avoid the memory allocation.

internal/http/logs.go
stacksmash76: Yeah, pretty sure the regex state machine will find a way to better optimize a single large expression than trying to do it by hand.

internal/http/logs_sanitize_test.go
stacksmash76: > Check the length, HasPrefix, then check the last character with another conditional to avoid the memory allocation.

It's a test case, no need.

internal/http/logs_sanitize_test.go
stacksmash76: > why not defer close above?

What would be the benefit here of deferring the call?

internal/http/logs.go
s0up4200: The reasons for splitting them up is to make it easier to maintain, when adding new indexers for example. It's easier to read. Sure, I can merge them into a single, but for what? A .1 speed increase? The previous one failed in several cases so I just redid them all one by one.

internal/http/logs_sanitize_test.go
KyleSanderson: There's a couple of us on Discord if you'd like to chat in real time.

zze0s (approved):
Great improvement! 🚀

PR #782, opened 2023-03-23
feat(actions): add "don't start" option for rtorrent

metonyms:
Someone in the discord asked for it* so here it is

*they said paused, but you can't do that on rtorrent from my understanding so i went with stopped

image

this matches the option in rutorrent's add dialog
image

### DIFF
--- a/web/src/screens/filters/action.tsx
+++ b/web/src/screens/filters/action.tsx
@@ -400,6 +400,14 @@ const TypeForm = ({ action, idx, clients }: TypeFormProps) => {
               />
             </div>
           </div>
+          <div className="col-span-12 sm:col-span-6">
+            <div className="col-span-6">
+              <SwitchGroup
+                name={`actions.${idx}.paused`}
+                label="Don't start download automatically"
+              />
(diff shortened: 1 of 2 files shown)


web/src/screens/filters/action.tsx
```diff
@@ -400,6 +400,14 @@ const TypeForm = ({ action, idx, clients }: TypeFormProps) => {
               />
             </div>
           </div>
+          <div className="col-span-12 sm:col-span-6">
+            <div className="col-span-6">
+              <SwitchGroup
+                name={`actions.${idx}.paused`}
+                label="Don't start download automatically"
```
s0up4200, line 407:
```suggestion
                label="Add paused"
                description="Add torrent as paused"
```
To make it uniform with the rest of the clients.

KyleSanderson: I think it's fine - they're matching the ruTorrent name which is what users at this level would expect. Think of it like a translation string.

zze0s (approved):

PR #791, opened 2023-03-24
fix(web): propagation bug with slide panel focus

### LINKED ISSUE #117: Propagation bug with the SlideOver component
If a notification pops up, while a SlideOver panel is opened, the SlideOver panel will always close itself, regardless if the notification should have consumed the user click.

The cause for this is most likely in the underlying headlessui package that is being used. Stopping event propagation on the X button won't work.

https://user-images.githubusercontent.com/98354295/152649702-e237f337-10f8-4da4-9873-7d5ce6b7284b.mp4

KaiserBh:
This fixes #117 by using portal we should eliminate the element from being in the same component that way the dialogue doesn't close by clicking on the X button on the toast.


zze0s (approved):
👏

PR #800, opened 2023-03-30
enhancement(settings): make lists sortable

s0up4200:


zze0s (approved):
There's quite a bit of duplication going on but it looks to be needed. 

A future improvement could be to build a reusable hook.

🚢

PR #807, opened 2023-04-02
feat(filters): support raw announce filtering

zze0s:
Add support for raw announce filtering.


KyleSanderson:
This is a huge anti-feature. Yes, it can be used in limited stop-gap situations to get people around bugs in autobrr, but exposing everything about the announce means that our definitions are not good enough for usage with the rest of the filtering. The one exception to this is RSS (and potentially Torznab) where we don't know the strings beforehand, and require user-intervention for the feeds to extract the parts out that are desired.

I might be wrong, but this feels actively harmful and encourages bad behaviour.

PR #810, opened 2023-04-03
feat(filters): add match logic for tags and except tags

gdmachado:
Added logic to support any/all matching logic for tags and except tags.

I found some unused fields in the `Filter` struct (`TagsAny`, `ExceptTagsAny`) that seemed like a way to start implementing this, but are not used anywhere. I kept them there for now.

Until now tag evaluation was always evaluated using `any` from what I could tell. Created new fields to store the logic to be used. In the DB migrations I updated the values of these to "any", not sure if that's the correct way to go about it

### DIFF
--- a/internal/domain/filter.go
+++ b/internal/domain/filter.go
@@ -115,6 +115,8 @@ type Filter struct {
 	ExceptTags                  string                 `json:"except_tags,omitempty"`
 	TagsAny                     string                 `json:"tags_any,omitempty"`
 	ExceptTagsAny               string                 `json:"except_tags_any,omitempty"`
+	TagsMatchLogic              string                 `json:"tags_match_logic,omitempty"`
+	ExceptTagsMatchLogic        string                 `json:"except_tags_match_logic,omitempty"`
 	MatchReleaseTags            string                 `json:"match_release_tags,omitempty"`
 	ExceptReleaseTags           string                 `json:"except_release_tags,omitempty"`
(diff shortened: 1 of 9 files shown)


internal/domain/filter.go
```diff
@@ -686,6 +702,39 @@ func containsMatch(tags []string, filters []string) bool {
...
+			if tag == "" {
+				continue
+			}
+			tag = strings.ToLower(tag)
+
+			a := strings.ContainsAny(filter, "?|*")
+			if a {
+				match := wildcard.Match(filter, tag)
+				if match {
+					found = true
```
stacksmash76, line 724: Break from the inner loop when a match gets found. Ditto for below.

internal/domain/filter.go
```diff
@@ -686,6 +702,39 @@ func containsMatch(tags []string, filters []string) bool {
...
+		filter = strings.Trim(filter, " ")
+		found := false
+
+		for _, tag := range tags {
+			if tag == "" {
+				continue
+			}
+			tag = strings.ToLower(tag)
+
+			a := strings.ContainsAny(filter, "?|*")
```
stacksmash76, line 720: You could reorganize the checking logic. First check whether tag == filter and then wrap the alternative logic in an else statement and then check for "a".

zze0s (approved):
Hey! Thanks for the PR! This is something users have been asking about for a long time! 👏

PR #812, opened 2023-04-04
fix(web): RegexField add padding

s0up4200:
Added a conditional pr-10 if useRegex=true to make the right side of the field remain visible when the icons are present.


martylukyy (approved):
Looks clean! 🚀

Issue #501, opened 2022-10-16
[INDEXER REQ] Tokyo Toshokan support


Drapersniper:
**We currently only support indexers with IRC announces.**

# Indexer request

-  Do they have irc announce? Yes 
-  Link to existing `autodl-irssi .tracker` file:  Haven't found any

### If no existing `autodl-irssi .tracker` file

- Indexer name: TokyoTosho
- Short name: TT
- URL: https://www.tokyotosho.info/
- Download link examples:
- https://dl.dmhy.org/2022/10/17/78c8176aa3e2436d6131d673f6379333b7f7c17e.torrent
- https://ehtracker.org/get/2352897/d0b4ea6c7e0458e2c2184189c4007199a4b0a8d1.torrent

## Announce examples

Please provide 1-5 different announces. And change release titles for good measure. If they have tags like freeleech and internals etc make sure to get examples of those.

- Full definition on their documentation [tokyotosho.info/pico/?sub%2FIRC_API](https://www.tokyotosho.info/pico/?sub%2FIRC_API)
- � is ASCII char 30 - Record Separator

- Torrent1706318Raws7[UCCUSS] Papa Mama Bye Bye パパ ママ バイバイ (DVD 640x480p AVC AC3).mkvhttps://dl.dmhy.org/2022/10/17/78c8176aa3e2436d6131d673f6379333b7f7c17e.torrent1.75GB
- Torrent_Meta170631878c8176aa3e2436d6131d673f6379333b7f7c17evolta_john0
- Torrent1706319Hentai (Manga)13[Sakekan Memorial (SOLOPIPB)] Shiragasane Kire (Shiragasane Soushuuhen Koromo) [English] [Kafka]+[rampantserenity]+[Lazarus H]+[TSHH] [Digital] [FULL COMPILATION].ziphttps://ehtracker.org/get/2352897/d0b4ea6c7e0458e2c2184189c4007199a4b0a8d1.torrent297.66MB
- Torrent_Meta1706319d0b4ea6c7e0458e2c2184189c4007199a4b0a8d1sdgfjuyuire0
- Torrent1706320Hentai (Manga)13(成年コミック) [西川康] 友達の母は僕のモノ [DL版].ziphttps://ehtracker.org/get/2352975/b9391ded079430f0968a266b7044fb0072027c94.torrent137.7MB
- Torrent_Meta1706320b9391ded079430f0968a266b7044fb0072027c94milannews0
## IRC Network

- Network name: Rizon
[...]

Issue #817, opened 2023-04-06
Special characters in IRC messages are removed


paperclip-go-brr:
**Version**
v1.23.0

**Describe the bug**
In the process of creating a custom definition for Tokyo Toshokan, the special ASCII character with code `30` is used to delimit the different fields in the message. The problem is that autobrr sanitizes the message and removes the ASCII character.
Specifically here:
https://github.com/autobrr/autobrr/blob/4bf023d0306510dce12e6ab4e77085bc4ec81366/internal/irc/handler.go#L857-L860

Which calls the `Strip` function from the `irc-go` library:
https://github.com/ergochat/irc-go/blob/v0.2.0/ircfmt/ircfmt.go#L181-L202

This basically makes it impossible to match against the messages in the `#tokyotosho-api` channel as specified in the following issue: https://github.com/autobrr/autobrr/issues/501

**To Reproduce**

The overall steps to reproduce is to setup an irc server
```shell
> docker run --name ircd -p 6667:6667 inspircd/inspircd-docker
```
Use the following python script to send the special character in the channel.
```python
import pydle

class MyOwnBot(pydle.Client):
    async def on_connect(self):
         await self.join('#autobrr')

    async def on_message(self, target, source, message):
         sc = chr(30)
         # don't respond to our own messages, as this leads to a positive feedback loop
         if source != self.nickname:
            await self.message(target, f"Special char -->{sc}<-- -->{sc}<--")

client = MyOwnBot('pybot', realname='pybot')
client.run('localhost', tls=False, port=6667, tls_verify=False)
```

Setup a definition to listen to the irc server/channel (and `pybot` as announcer) and add a breakpoint [here](https://github.com/autobrr/autobrr/blob/4bf023d0306510dce12e6ab4e77085bc4ec81366/internal/irc/handler.go#L560), and send a message in the IRC channel, and the python script will then respond with the message containing the special character.

**Expected behavior**
To be able to match against the special characters in my custom definition's regex expression.

**Screenshots**
In my test scenario, the following screenshot is from a breakpoint here:
https://github.com/autobrr/autobrr/blob/4bf023d0306510dce12e6ab4e77085bc4ec81366/internal/irc/handler.go#L558-L560
![image](https://user-images.githubusercontent.com/112905434/230360171-a5f615f4-7ed8-4783-9c59-7f6cd4bba083.png)

paperclip-go-brr:
[...]

PR #818, opened 2023-04-06
feat(indexers): add Tokyo Toshokan

paperclip-go-brr:
Add support for Tokyo Toshokan indexer.
Closes https://github.com/autobrr/autobrr/issues/501

The only caveat is that this definition uses their main `#tokyotosho` channel on IRC instead of the `#tokyotosho-api` channel as specified in the issue (https://github.com/autobrr/autobrr/issues/501). This is because of the way they delimit their messages and is therefore not possible to use at the moment, as described in this related issue: https://github.com/autobrr/autobrr/issues/817.

### DIFF
--- a/internal/indexer/definitions/tokyotoshokan.yaml
+++ b/internal/indexer/definitions/tokyotoshokan.yaml
@@ -1,0 +1,63 @@
+---
+#id: tokyotosho
+name: Tokyo Toshokan
+identifier: tokyotosho
+description: A BitTorrent Library for Japanese Media
+language: en-us
+urls:
+  - https://
+privacy: public
+protocol: torrent
+supports:
+  - irc
+
+irc:
+  network: Rizon
+  server: irc.rizon.net
+  port: 6697
+  tls: true
+  channels:
+    - "#tokyotosho"
+  announcers:
+    - "TokyoTosho"
+  settings:
+    - name: nick
+      type: text
+      required: true
+      label: Nick
+      help: Bot nick. Eg. user_bot
+
+    - name: auth.account
+      type: text
+      required: false
+      label: NickServ Account
+      help: NickServ account. Make sure to group your main user and bot.
+
+    - name: auth.password
+      type: secret
+      required: false
+      label: NickServ Password
+      help: NickServ password
+  parse:
+    type: multi
+    lines:
+      - test:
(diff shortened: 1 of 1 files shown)


internal/indexer/definitions/tokyotoshokan.yaml
```diff
@@ -0,0 +1,63 @@
...
+    lines:
+      - test:
+          - "Release 1748839: [Anime] SOMETITLE-007"
+        pattern: '^Release [0-9]*: \[(.*?)\] (.*)'
+        vars:
+          - category
+          - torrentName
+      - test:
+          - "Torrent: https://example.com/torrents/sha123/torrentname-007.torrent"
+        pattern: '^Torrent: (https:\/\/)(.*\.torrent)'
```
zze0s, line 52: I've been monitoring the channel for a couple of days and some urls does not have the `.torrent` so changing pattern to `^Torrent: (https:\/\/)(.*)` would be better.

And it could be simplified more to not capture the scheme: `^Torrent: https:\/\/(.*)` and then remove `- baseUrl`.

zze0s (changes requested):
Hey! Thanks for the PR! Think most of us was only aware of the -api channel which had the weird characters. Thankfully we already support multi-line announces so then this is easy.

One minor thing that needs to be changed and another minor thing that could be changed.

internal/indexer/definitions/tokyotoshokan.yaml
paperclip-go-brr: Good point! I have updated the regex and removed `baseUrl` 👍

paperclip-go-brr:
It seems i forgot to mention that the main channel does not contain any information as to who the uploader is, as opposed to the api channel where the uploader is included in the message.
I would prefer to be able to filter based on who the uploader is, but because of the special characters in the api channel, this PR will have to do for now.

zze0s:
> It seems i forgot to mention that the main channel does not contain any information as to who the uploader is, as opposed to the api channel where the uploader is included in the message. I would prefer to be able to filter based on who the uploader is, but because of the special characters in the api channel, this PR will have to do for now.

Lets see what we can do about that in the future.

zze0s (approved):

zze0s:
Related #1

PR #820, opened 2023-04-07
fix(web): auth bad login flash

KaiserBh:
since it's 401 we should reject the promise that way onError actually works. I think this was causing login to succeed  for like a second then goes back to login but with this we reject it early shouldn't pass the login screen.


stacksmash76:
Just a note: this is a regression to old behavior.
The problem is described in PR #331 and it was z's hacky way of mitigating a login failed error. So, I presume you'll have to add an exception in onError, in case the boss still thinks it's appropriate not to tell the user they've made a mistake (but don't tell the boss that I've never told him that he made a mistake)

zze0s (approved):
Good change ;)

PR #823, opened 2023-04-07
refactor(web): migrate create-react-app to vite

KaiserBh:
This should fix all the issue with baseUrl hopefully and we can say goodbye to create-react.


web/src/utils/index.ts
```diff
@@ -9,17 +9,12 @@ export function sleep(ms: number) {
...
   let baseUrl = "";
   if (window.APP.baseUrl) {
-    if (window.APP.baseUrl === "/") {
-      baseUrl = "/";
-    } else if (window.APP.baseUrl === "{{.BaseUrl}}") {
-      baseUrl = "/";
-    } else if (window.APP.baseUrl === "/autobrr/") {
-      baseUrl = "/autobrr/";
+    if (window.APP.baseUrl === "{{.BaseUrl}}") {
+      baseUrl = ""; // Use an empty string for local development
```
zze0s, line 13: This might need to be `/`. Still need to further test but without it my local dev env seem to break on hard refresh on nested routes like `/filters/id/actions`.

For built binaries and baseUrl it does not seem to matter much.

KaiserBh: hmm okay, well let me know if we need it then I will add it.

stacksmash76: The easiest would be just to leave it as it was. Otherwise, you can fix the baseUrl quirk in APIClient.ts:
```
  let base = baseUrl();
  if (base.slice(-1) !== "/")
    base += "/";

  return window.fetch(`${base}${endpoint}`, config)
```
(which wouldn't be a bad thing by itself either)

stacksmash76:
Another thing, it would be nice if in case there's no dist/ directory, the go backend would build the frontend on demand instead of spitting out `web/build.go:30:13: pattern all:dist: no matching files found`

KaiserBh:
> Another thing, it would be nice if in case there's no dist/ directory, the go backend would build the frontend on demand instead of spitting out `web/build.go:30:13: pattern all:dist: no matching files found`

@stacksmash76 what would you propose then? just check for the error then just call exec and execute the yarn build command? unless there is a better solution?

zze0s:
> > Another thing, it would be nice if in case there's no dist/ directory, the go backend would build the frontend on demand instead of spitting out `web/build.go:30:13: pattern all:dist: no matching files found`
> 
> @stacksmash76 what would you propose then? just check for the error then just call exec and execute the yarn build command? unless there is a better solution?

Not sure we should make it build from go/backend but we could add an “empty”  dist dir with just the index.html so it can still build.

KaiserBh:
> > > Another thing, it would be nice if in case there's no dist/ directory, the go backend would build the frontend on demand instead of spitting out `web/build.go:30:13: pattern all:dist: no matching files found`
> > 
> > 
> > @stacksmash76 what would you propose then? just check for the error then just call exec and execute the yarn build command? unless there is a better solution?
> 
> Not sure we should make it build from go/backend but we could add an “empty” dist dir with just the index.html so it can still build.

Hmm yea that would be better and easy I suppose otherwise another approach is that we make a run.sh
untested script.
```sh
#!/bin/bash
if [ "$1" == "-rg" ]; then
  go clean
  cd web && yarn run build
  cd .. && go build -o autobrr cmd/autobrr/main.go
  ./autobrr
elif [ "$1" == "-h" ] || [ "$1" == "--help" ]; then

  echo "Usage: ./run.sh [OPTION]"
  echo "Run autobrr."
  echo "Options:"
  echo "  -rg, --run-generate  It will build the frontend and backend and run autobrr"
else
  echo "Invalid option. Use -h or --help to see the options."
fi
```
can make something similar for windows but yea that's the gist of it. Otherwise the simplest solution is just what you said we have a dist folder with an html file.

zze0s:
The old setup required the frontend being built before compiling so this is not a regression, but I added an empty `web/dist` dir which now allows it to compile without the need to build the frontend, which is useful for local development.

The full build with Actions take care of it anyway, and the full Dockerfile builds it as well, and same with `make build` which builds frontend before.

zze0s (approved):

PR #824, opened 2023-04-07
fix(web): PasswordField eye icon alignment

martylukyy:
Small cosmetic bug i came across during testing for the migration to vite.

eyeicon

### DIFF
--- a/web/src/components/inputs/input.tsx
+++ b/web/src/components/inputs/input.tsx
@@ -280,7 +280,7 @@ export const PasswordField = ({
           {label} {required && <span className="text-gray-500">*</span>}
         </label>
       )}
-      <div className="relative">
+      <div>
         <Field name={name} defaultValue={defaultValue}>
           {({
             field,
--- a/web/src/components/inputs/input.tsx
+++ b/web/src/components/inputs/input.tsx
@@ -295,7 +295,7 @@ export const PasswordField = ({
               placeholder={placeholder}
             />
 
-            <div className="absolute inset-y-0 right-0 px-3 flex items-center" onClick={toggleVisibility}>
+            <div className="absolute inset-y-0 right-0 px-3 py-2.5 flex" onClick={toggleVisibility}>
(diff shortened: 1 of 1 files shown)


stacksmash76:
The problem here is that the parent is `sm:col-span-2 relative` and then on error the parent gets added to it, which increases the parent's height (since the icon container div depends on the parent's height)

A solution would be to move the error message outside of the parent. Or add a faux parent and move the current one down the stack, so the icon container and error message don't share the same relative parent.

martylukyy:
Thanks for suggesting proper solutions - i added you to review if you want to review.
Feel free to suggest changes in case you dislike parts of the solution now. :)

web/src/components/inputs/input.tsx
```diff
@@ -280,35 +280,44 @@ export const PasswordField = ({
...
-              type={isVisible ? "text" : "password"}
-              autoComplete={autoComplete}
-              className={classNames(meta.touched && meta.error ? "focus:ring-red-500 focus:border-red-500 border-red-500" : "focus:ring-blue-500 dark:focus:ring-blue-500 focus:border-blue-500 dark:focus:border-blue-500 border-gray-300 dark:border-gray-700", "mt-2 block w-full dark:bg-gray-800 dark:text-gray-100 rounded-md")}
-              placeholder={placeholder}
-            />
-
-            <div className="absolute inset-y-0 right-0 px-3 flex items-center" onClick={toggleVisibility}>
-              {!isVisible ? <EyeIcon className="h-5 w-5 text-gray-400 hover:text-gray-500" aria-hidden="true" /> : <EyeSlashIcon className="h-5 w-5 text-gray-400 hover:text-gray-500" aria-hidden="true" />}
-            </div>
+      <div className="relative">
```
stacksmash76, line 283: Pretty sure you don't need this parent relative here.

stacksmash76 (approved):
Looks good to me.

web/src/components/inputs/input.tsx
martylukyy: True. fixed. Thanks for looking over it! Always a pleasure!

s0up4200 (approved):

PR #832, opened 2023-04-12
fix(indexers): OPS parsing

s0up4200:
https://regex101.com/r/pHwz9e/1


martylukyy (approved):

zze0s (approved):
Weird that they removed year and type, but oh well 🤷

PR #837, opened 2023-04-14
fix(indexers): polishsource downloads

zze0s:
Debug issues with polishsource downloads.


zze0s (commented):
Tested by multiple users with great success!

PR #840, opened 2023-04-16
feat(releases): keep torrent in memory and only write a file when needed

KyleSanderson:
# Description

Keep the downloaded `.torrent` on the release as bytes instead of writing it to a file in `/tmp`, and only write a file when something actually needs a path.

`DownloadRelease` now stores the torrent in `Release.TorrentDataRawBytes`. A file on disk is created on demand by `Release.WriteTemporaryFile()`, which is called only when a macro needs a path (`TorrentPathName`, `TorrentTmpFile`) for an external script or webhook. Everything else, including all five torrent clients and the watch folder action, works off the bytes.

This is mostly deleting work we were already doing twice. `downloadTorrentFile` already read the whole response into `bodyBytes` before writing it to the tmp file, and four of the five clients read that file straight back into memory (`os.ReadFile` for deluge and rtorrent, `File2Base64` for transmission, `io.ReadAll` for porla). For those the peak memory is unchanged and the write plus read round trip is gone. qBittorrent moves from `AddTorrentFromFileCtx` to `AddTorrentFromMemoryCtx`; both funnel into the same `postReaderCtx` in go-qbittorrent and the multipart filename was already a generated string, so there is no behavioural difference there.

The branch was based on develop from April 2023 and was 949 commits behind. Rather than replaying three year old conflict hunks, develop was merged in and the change rebuilt against current code. The old branch also carried a nil `*bytes.Buffer` that would have panicked on every download; none of that code survives.

Related https://github.com/autobrr/autobrr/pull/839

## Bugs found and fixed along the way

Chasing whether the change was worth doing turned up several real defects, all sharing one root cause: the action executors took `domain.Release` **by value**, so anything an executor wrote to the release went to a copy that was then discarded.

- **Temporary files were leaked.** A download started inside an executor set `TorrentTmpFile` on the copy. `CleanupTemporaryFiles` runs against the original, saw an empty path, and removed nothing. Every torrent added through a download client left a file in `/tmp`.
- **The cleanup job that was meant to catch this deleted nothing.** `TempDirCleanupJob` resolved the current user from `os.Getenv("UID")`, which is a shell variable that is never exported to a process, then fell back to `USER`, which is a name. `strconv.ParseUint` failed on both, so the ownership check returned false for every file and the sweep was a no op. Confirmed by running the new test against the current implementation, where the stale file survives.
- **One indexer fetch per action instead of per release.** The next action got a fresh copy with empty bytes, so the "already downloaded" guard never fired and the same torrent was fetched again. This counts against download limits on indexers that track them.
- **Replaying an action leaked its temporary file.** `Retry` goes through `retryAction`, not `processRelease`, and `processRelease` holds the only `CleanupTemporaryFiles` call.
- **External filter macros that never triggered a download.** `NeedTorrentDownloaded` looked for `TorrentDataRawBytes` in `WebhookData` but not `ExecArgs`, and `TorrentTmpFile` was exposed as a macro but checked nowhere. Both expanded to nothing.

The executors now take `*domain.Release`. Nothing relied on the copy isolating writes; the only writes were the `DownloadRelease` calls, and those propagating is the point.

The cleanup job was also hardened for shared systems, where several users run off one box with a shared `/tmp` and isolation comes from permissions. It now compares against the effective uid, reads the directory entry with `Info` rather than `os.Stat` so a symlink is judged by the link and not by whatever it points at, and skips anything that is not a regular file.

## Behaviour changes worth noting

- Watch folder files are named `autobrr-<infohash>.torrent` instead of `autobrr-<random>.torrent`. The `autobrr-` prefix is kept, so existing `autobrr-*.torrent` globs still match. The infohash is unique, needs no escaping, and anything richer can be read out of the file itself.
- Deluge receives `<infohash>.torrent` as the filename argument to `add_torrent_file` instead of the temporary file path. Deluge only uses it as a name hint.
- External filters using `{{.TorrentDataRawBytes}}` in exec args, or `{{.TorrentTmpFile}}` anywhere, now cause a torrent download where they previously silently did not. That is the macros working as documented, but it is a fetch that was not happening before.

## Type of change

- [x] Refactor / maintenance (no functional change)

## How has this been tested?

* `go test ./...` passes and `go build ./...` is clean.
* New tests cover the areas the change touches: `WriteTemporaryFile` round trip, idempotence and failure without contents, `CleanupTemporaryFiles` clearing both the file and the bytes and tolerating a file an exec script already consumed, `TorrentReader` being re readable, both macro checks on the action and external filter side, the watch folder write from memory including the magnet and missing torrent rejections, and the temp dir cleanup job.
* The two cleanup job tests were run against the current implementation first to confirm they fail there: the stale file is not removed, and a symlink named `autobrr-*` is followed and unlinked. Both pass after the fix.
* One indexer fetch across several download client actions is asserted with a fake download service that only counts a fetch when the release has no bytes, mirroring the real guard.
* **Not tested against live clients.** qBittorrent, Deluge, rTorrent, Transmission and Porla were verified by reading their client libraries, not by adding a torrent to a running instance. The qBittorrent and Deluge call changes in particular deserve a real smoke test before merge.

## Checklist

- [x] If this changes the database schema, I have added migrations for both SQLite and PostgreSQL (no schema change)


zze0s:
This has been untouched for too long. Please fix the conflicts to get develop merged.

zze0s:
Will be superseded by a new pr.

zze0s (approved):

PR #841, opened 2023-04-16
fix(notifications): test handle unsupported type

KaiserBh:
If the agent variable is nil (This can happen if the `notification.Type` doesn't match any of the cases in the switch statement.) the server will crash.

On the frontend  when user test the notification without providing the type it will crash. Maybe we need to change the frontend as well, to make sure that the test button is disabled unless all the required field is filled, but this change at least will prevent the server from crashing.

### DIFF
--- a/internal/notification/service.go
+++ b/internal/notification/service.go
@@ -240,13 +240,14 @@ func (s *service) Test(ctx context.Context, notification domain.Notification) er
 
 	g, _ := errgroup.WithContext(ctx)
 
+	if agent == nil {
+		s.log.Error().Msgf("unsupported notification type: %v", notification.Type)
+		return errors.New("unsupported notification type")
+	}
+
 	for _, event := range events {
 		e := event
 		g.Go(func() error {
-			if agent == nil {
-				s.log.Error().Msgf("unsupported notification type: %v", notification.Type)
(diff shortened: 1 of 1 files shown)


internal/notification/service.go
```diff
@@ -237,7 +238,12 @@ func (s *service) Test(ctx context.Context, notification domain.Notification) er
 		agent = NewTelegramSender(s.log, notification)
 	}
 
-	g, ctx := errgroup.WithContext(ctx)
+	g, _ := errgroup.WithContext(ctx)
+
+	if agent == nil {
+		s.log.Error().Msgf("unsupported notification type: %v", notification.Type)
+		return errors.New("unsupported notification type")
+	}
```
zze0s, line 246: The idiomatic solution is to add a `default` case that return an error if it's an unknown or missing type.

zze0s (changes requested):

internal/notification/service.go
KaiserBh: ouhh okay will update it

KaiserBh: updated it @zze0s let me know if that's what you wanted. 
ngl forgot about default in switch statement.

zze0s (approved):

PR #842, opened 2023-04-16
fix(web): reset global state after logout

### LINKED ISSUE #360: web: focus (?) bug breaks global state
See attachment below for a video demonstration.

Steps to reproduce:
- be logged in
- do a full page refresh
- logout
- don't click anything and hover over the username/password input components
- the cursor will start flicking (only on input/button components for some reason?)

After the cursor starts flicking, interaction possibilities will be limited (e.g. you can log in/do an action, but it can *possibly* be reverted)

This occurred to me during my latest PR, and I can confirm it's present in 0.26.1 and latest commit and PR #359. Another thing to note is that initially I didn't trigger it the way I described it here. I was working on input components and it suddenly happened. I don't know what I did and can't reproduce it that way.

I'm thinking it has something to do with how some of the parameters get constructed/get passed to input component and further into node modules and some function gets directly ad absurdum.

https://user-images.githubusercontent.com/98354295/179424794-0c2bfa64-61ea-4421-b0ce-a6bd45e141be.mp4

(Note how react router cancels navigation, possibly due to the function's called side-effect)

I didn't find any console statements so far.

Edit: A full page refresh fixes this behavior.
Edit2: Might even have something to do with notifications and their state.

KaiserBh:
fixes  #360 I think.


stacksmash76 (approved):
So a syncing bug with navigation. How great.

KaiserBh:
> So a syncing bug with navigation. How great.

honestly I find out about it while going through the code base and was like hmm I am sure there was a hack somewhere.

martylukyy (approved):

zze0s (approved):
Fantastic find!!

PR #843, opened 2023-04-16
fix(web): unrecoverable unauthorized error

KaiserBh:
added check for failed login as well remove user when visiting login.

image


martylukyy (approved):
Looks good to me!
However I would have more piece of mind if @s0up4200 could look over this too,
since i had some difficulties replicating this bug from time to time.

s0up4200 (approved):

zze0s (approved):
A bit skeptic about this change but if it fixes the issue then by all means 😄

PR #855, opened 2023-04-18
build: try buildjet runners

zze0s:
Try BuildJet runners with ARM builder for Docker.


zze0s:
With the recent work from https://github.com/autobrr/autobrr/pull/859 this is not really needed any longer. Build times are acceptable now!

PR #859, opened 2023-04-19
build: make ci 2-3x faster

KyleSanderson:


zze0s (approved):
🚀 

Huge improvements in overall build time!! From about 25min down to 8-9 👏

PR #862, opened 2023-04-20
fix(indexers): OPS new announce format

### LINKED ISSUE #850: OPS not working
**Version**
v1.24.0

**Describe the bug**
The artist's name is returning as empty and isn't able to match the filter

from the logs(I've trimmed the comparison block as I have a lot of artist in it):
{
    "level": "debug",
    "module": "irc",
    "network": "irc.orpheus.network",
    "channel": "#announce",
    "user": "hermes",
    "time": "2023-04-18T08:22:22+01:00",
    "message": "TORRENT: Sanguisugabogg – Homicidal Ecstasy – CD/FLAC/Lossless – metal,rock – https://orpheus.network/... – https://orpheus.network/...?...&torrentid=....&action=download"
}

{
    "level": "debug",
    "module": "filter",
    "time": "2023-04-18T08:22:22+01:00",
    "message": "filter.Service.CheckFilter: (Orpheus) for release: Sanguisugabogg – Homicidal Ecstasy rejections: (artists not matching. got:  want: ..., Samia, Sanguisugabogg, Saweetie, ...)"
}

The regex was was changed recently is working correctly 
https://regex101.com/r/ATXZbp/1

martylukyy:
Fixes #850 

- Extend regex capture group for torrentName var to contain releaseTags var
- Replace all – in announces with a - for torrentName var

I don't think replacing all – with - will have any repercussions on other matching/parsing.
In case there are any concerns about this, feel free to express them.

NOTE: 
Testing this PR with release_test.go will not work, because it does not come across the newly added line.
This is why I did **not** write tests for this change.
When i tried to convert this into a function to use it in release_test.go
i ran into an import cycle which i don't know how to solve, except for duplicating the new line into the tests...
Maybe duplicating the line into all the tests in release_test.go is an okay thing to do? Is it needed? Thoughts?


internal/announce/announce.go
```diff
@@ -199,6 +199,9 @@ func (a *announceProcessor) onLinesMatched(def *domain.IndexerDefinition, vars m
 		return err
 	}
 
+	// replace all – with - in torrentName var (needed for proper parsing with rls)
+	rls.TorrentName = strings.ReplaceAll(rls.TorrentName, "–", "-")
```
zze0s, line 203: This feels like opening a can of worms with unexpected behavior.

Might be better to use the Sprig Replace function on `torrentName` for the indexer definition instead.

http://masterminds.github.io/sprig/strings.html#replace

martylukyy: If you have a solution in your head, i would appreciate you implementing that if that's okay for you!

zze0s (approved):

PR #863, opened 2023-04-20
feat(macros): add Category and Categories

cleesieboy:
Add "Category" and "Categories" to available filter variables, ie available as `{{ .Category }}`.


zze0s (approved):
Thanks for the PR!

Small and simple improvement 👍

Issue #865, opened 2023-04-21
Data from deleted filters is getting transferred into new filters


martylukyy:
**Version**
v1.24.1

**Describe the bug**
When deleting a filter and creating a new one, data from the old filter gets transferred into the new filters
which might lead to unwanted behavior with the new filter.
This does not only happen with the name field. It happens to other fields too.

**Video to Reproduce**

https://user-images.githubusercontent.com/35452459/233619609-e972f9aa-acd9-43d1-9a54-b1f43bc891f3.mp4

**Expected behavior**
When deleting a filter and creating a new one afterwards it should start without the data of the old filter.

stacksmash76:
Probably same thing as #864

PR #868, opened 2023-04-22
refactor(web): replace pkg `react-query` with `tanstack/react-query`

### LINKED ISSUE #864: Enabled state does not get refreshed when opening a filter.
**Version**
v1.24.1

**Describe the bug**
Enabled state does not get refreshed when opening a filter or when changing the enabled status in the filter list.
This causes a wrong status shown in the filter itself when enabled/disabled it before on the filter list.

**Video to Reproduce**

https://user-images.githubusercontent.com/35452459/233610836-7776c95f-8d42-4c2f-b161-124d8727a4e0.mp4

**Expected behavior**
When enabling/disabling a filter on the filter list and opening it afterwards it should show the correct status.

**Additional context**
If I read the code correctly, it should already invalidateQueries, but it seems like it does not?
https://github.com/autobrr/autobrr/blob/286f2f53f74c465f7e3d37857665a359c96b0f85/web/src/screens/filters/list.tsx#L636-L658
https://github.com/autobrr/autobrr/blob/286f2f53f74c465f7e3d37857665a359c96b0f85/web/src/screens/filters/list.tsx#L673-L675

Adding a invalidation to the filter.name fixes it, but feels wrong because it takes pretty long to open the filter then 
and it looks like a hard refresh...

This should be a simple queryClient.invalidateQueries somewhere but i don't know where to put it
since where i would put it, it is either already there or it takes pretty long to open the filter then and it looks like a hard refresh...

zze0s:
Lets move to the newer `@tanstack/react-query` where all new stuff will be.

This is a decently big refactor and in doing so fixes caching issues.

Fixes #864. #865


zze0s:
@stacksmash76 If you have time to look, any overall comments?

There's some queries and mutations for settings that can be more effective like removing from cache instead of invalidating, and updating in place and so on, but it's imo more a nice to have than strictly needed. 😄

Overall it makes it more unified and the pattern with separate queryKeys make things a lot easier to manage. There's a couple ones that are not done like `updates` but imo not really needed.

After this I think I'll play with tanstack router since it have some good QoL improvements especially for managing URL Params and Query Params in a typesafe way, and ties in with react-query nicely.

web/src/screens/filters/details.tsx
```diff
@@ -144,37 +145,49 @@ export default function FilterDetails() {
...
   const { filterId } = useParams<{ filterId: string }>();
 
-  const { isLoading, data: filter } = useQuery(
-    ["filters", filterId],
-    () => APIClient.filters.getByID(parseInt(filterId ?? "0")),
-    {
-      retry: false,
-      refetchOnWindowFocus: false,
-      onError: () => navigate("./")
+  if (filterId === "0" || undefined) {
```
stacksmash76, line 148: filterId should never be undefined. Since we match for "/*" after it.

Personally, I wouldn't hard-code navigate-back checks like this, because they will always miss on upper bound (if id >= max filter id), so I think it would be best if we made the HttpClient function respect relative redirects in those cases. And we also don't notify the user in this case, which is kinda eh (even though this case is extremely improbable).

stacksmash76:
Generally it's okay. Not what I would've done though.

Instead, I would've made a file for each separate "module" in the Api/ folder and moved 90% of the logic there. This would mean that useQuery/useMutation would be called there and you would only pass the other props (you can use Omit in TS to remove queryKey/queryFn/mutationFn so no collision happens) and then ultimately you'd just be passing the targeted input object with optional tanstack query args object.

Would make everything easier, IMO. But this is fine also.

Still, a few nitpicks:
- you're using arrays for keys, which isn't great, for the simplest reason: if you specify an out-of-bounds index, you end up with... undefined. Which might collide on some occasions. Unnamed/anonymous objects would've been better.
- we're still using isLoading/isError, when we should've switched to React Suspense + Error Boundary fully

PS. Don't know where you stand on unified backend errors (which I mentioned I think almost a year ago), but since tanstack query uses an error type as its 2nd template argument (unknown by default), your approach to handling them will hurt a bit. (With an overloaded function/type as I described above this is relatively easy)

stacksmash76:
By the way, you might consider on revamping the HttpClient function interface a bit. This is what I'm using currently:
```typescript
interface Config {
  method?: string;
  body?: RequestBody;
  queryString?: Record<string, string | number | boolean | symbol | undefined>;
}

export async function HttpClient<T = unknown>(
  endpoint: string,
  config: Config = {}
): Promise<T> {
  const init: RequestInit = {
    method: config.method,
    headers: { "Accept": "*/*" }
  };

  if (config.body) {
    init.body = JSON.stringify(config.body);

    if (typeof(config.body) === "object") {
      init.headers = {
        ...init.headers,
        "Content-Type": "application/json"
      };
    }
  }

  if (config.queryString) {
    const queryString = new URLSearchParams();
    for (const [key, value] of Object.entries(config.queryString)) {
      if (typeof(value) === undefined)
        continue;

      queryString.set(key, String(value));
    }
    endpoint += `?${queryString.toString()}`;
  }

  const response = await window.fetch(encodeURI(API_DOMAIN + endpoint), init);
  if (response.status === 204)
    return Promise.resolve<T>({} as T);
  
  const isJson = response.headers.get("Content-Type")?.includes("application/json");
  const json = isJson ? await response.json() : null;

  // Resolve on success
  if (response.status >= 200 && response.status < 300) {
    if (json) {
      return Promise.resolve(json);
    } else {
      return Promise.resolve<T>({} as T);
    }
  }

  // Otherwise reject
  return Promise.reject(json);
}

// ...
const Get = <T>(endpoint: string, config: Config = {}) => HttpClient<T>(endpoint, {
  ...config,
  method: "GET"
});

// ... e.g.
// all: (page: number, ordering: string, numItems: number) => 
//      Get<Pagination<Message>>("messages", {
//        queryString: { page, ordering, items: numItems }
//      }),
```
Should be easier to read and extend.

zze0s:
Thanks for the feedback! 😄 

> Generally it's okay. Not what I would've done though.
> 
> Instead, I would've made a file for each separate "module" in the Api/ folder and moved 90% of the logic there. This would mean that useQuery/useMutation would be called there and you would only pass the other props (you can use Omit in TS to remove queryKey/queryFn/mutationFn so no collision happens) and then ultimately you'd just be passing the targeted input object with optional tanstack query args object.
> 
> Would make everything easier, IMO. But this is fine also.

Yeah that might make sense to do.

> Still, a few nitpicks:
> 
>     * you're using arrays for keys, which isn't great, for the simplest reason: if you specify an out-of-bounds index, you end up with... undefined. Which might collide on some occasions. Unnamed/anonymous objects would've been better.

Array keys are the only supported way in tanstack query so not much to do about that. 

https://tanstack.com/query/v4/docs/react/guides/query-keys
https://tkdodo.eu/blog/effective-react-query-keys

>     * we're still using isLoading/isError, when we should've switched to React Suspense + Error Boundary fully

Also something that should be unified for sure.

> PS. Don't know where you stand on unified backend errors (which I mentioned I think almost a year ago), but since tanstack query uses an error type as its 2nd template argument (unknown by default), your approach to handling them will hurt a bit. (With an overloaded function/type as I described above this is relatively easy)

Been planning to do an overhaul of all api handlers so they are consistent. Have done some when I have touched related code but need to go over all of them.

I have waaay less time these days so have to put the efforts in right places..

web/src/screens/filters/details.tsx
zze0s: It's ugly indeed, but with the planned switch, or at least exploration of Tanstack router then that will be an issue of the past.

react-router have done some changes as well into similar direction but it's getting more and more tied in with Remix.

stacksmash76:
> Array keys are the only supported way in tanstack query so not much to do about that.

I know. But you can use anonymous objects inside the query key array. But this has the drawback that you're gonna need a custom evaluating function when you want to invalidate all queries that start with X... (https://tanstack.com/query/v4/docs/react/guides/query-invalidation down at the bottom)

> I have waaay less time these days so have to put the efforts in right places..

Thankfully you've got soup and marty, two full-time frontend devs to do the job for you ;) Oh, and Kaiser! (gl0ryus where you at mayne)

zze0s:
> > Array keys are the only supported way in tanstack query so not much to do about that.
> 
> I know. But you can use anonymous objects inside the query key array. But this has the drawback that you're gonna need a custom evaluating function when you want to invalidate all queries that start with X... (https://tanstack.com/query/v4/docs/react/guides/query-invalidation down at the bottom)

It's pretty flexible with multiple ways to do stuff from simple to more advanced when needed so think we'll try this way for a bit and extend upon it with time.

> > I have waaay less time these days so have to put the efforts in right places..
> 
> Thankfully you've got soup and marty, two full-time frontend devs to do the job for you ;) Oh, and Kaiser! (gl0ryus where you at mayne)

Yeah they are great! Thanks for still taking time to review ❤️

martylukyy (changes requested):
I hate to do this but #864 and #865 are still an issue after this refactor unfortunately..

https://user-images.githubusercontent.com/35452459/233858399-824c09b1-637e-45ca-a491-7f2d1e1cdf54.mp4

stacksmash76:
(This is not a query issue, heading title has proper value, textbox does not)
It's just Formik being an asshat. Try this: move the initialValues object to a separate function and then do initialValues={GetInitialValues(filter)}. This will force a Formik re-render properly.

martylukyy (commented):
enableReinitialize seems to have fixed the issues..
Will do more testing at a later point if still needed then.

web/src/screens/settings/Api.tsx
```diff
@@ -1,20 +1,31 @@
...
+import { DeleteModal } from "@components/modals";
+import APIKeyAddForm from "@forms/settings/APIKeyAddForm";
+import Toast from "@components/notifications/Toast";
+import { APIClient } from "@api/APIClient";
+import { useToggle } from "@hooks/hooks";
+import { classNames } from "@utils";
+import { EmptySimple } from "@components/emptystates";
+
+export const apiKeys = {
+  all: ["feeds"] as const,
```
zze0s, line 16: Naughy naughty, should be `api_keys`.

martylukyy:
Discovered one more thing...
The toast when enabling/disabling a feed says it got disabled when you enabled it and vice versa in this PR.
The toast is correct on my dedi with v1.24.1.
Need to look at it if just the toast is saying the wrong thing or if there is something else wrong..
Too tired to fix it now. Will look later this week at it. If someone wants to fix it in the meantime, feel free to do so.

martylukyy (approved):
Looks good to me now!
I would appreciate if someone could check the last 3 commits i made before merging.
Apart from that i'd say this is ready to be merged!

PR #885, opened 2023-04-27
feat(feeds): add allow_empty to ignore empty feed caches

fionera:
This is required to allow people to fetch bookmark feeds that dont contain anything except their bookmarks

### DIFF
--- a/internal/database/feed.go
+++ b/internal/database/feed.go
@@ -37,6 +37,7 @@ func (r *FeedRepo) FindByID(ctx context.Context, id int) (*domain.Feed, error) {
 			"f.interval",
 			"f.timeout",
 			"f.max_age",
+			"f.allow_empty",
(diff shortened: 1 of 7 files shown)


KyleSanderson (commented):
Nice 😸. I think you might be able to get the same behaviour if you disable the max age check. Is that ultimately what you're trying to accomplish here?

fionera:
No I also want to allow fetching releases even if the cache is empty. In my tests I had to add one fake entry into the cache table else it always ignored my bookmarks since it thought its the first fetch

zze0s:
Us maintainers discussed this and it is probably not a needed feature because the current way is to be considered a bug. We added the Max age check after the initial feeds implementation and did not change the initial cache check which probably should have been done.

We'll make a fix for it :smile:

PR #888, opened 2023-04-29
enhancement(web): specify error message when not choosing a client

martylukyy:
Every now an then we get questions about following error message:
`{"message":"json: cannot unmarshal string into Go struct field Action.actions.client_id of type int32"}`
So i went ahead and caught the error to replace it with a understandable one:

image

### DIFF
--- a/web/src/api/APIClient.ts
+++ b/web/src/api/APIClient.ts
@@ -38,6 +38,10 @@ export async function HttpClient<T>(
           return Promise.reject(new Error("Not found"));
         }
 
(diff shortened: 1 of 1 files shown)


s0up4200 (approved):
Good change 👍

stacksmash76:
Nope from me. Backend boys need to fix their shit and provide an interface to standardized error messages.

KaiserBh:
> Nope from me. Backend boys need to fix their shit and provide an interface to standardized error messages.

yea agreed, the client/frontend don't need to know what error exactly happened, because that's what logs are for, so we need to send proper error messages to the clients etc, that way the frontend don't need to do so much hacky stuff everything will be streamlined that way.

zze0s:
More proper fix with frontend validation using Zod. https://github.com/autobrr/autobrr/pull/888

PR #893, opened 2023-05-01
fix(notifications): always verify Telegram certificate

yegle:


zze0s:
Hi! I suppose that's a good change to verify it.

PR #598, opened 2022-12-31
feat(notifications): add Pushover

nelsonpecora:
This PR adds Pushover notifications support to Autobrr. Users can set their desired `api_key`, `user_key`, and `priority` in the settings.

PR #894, opened 2023-05-01
feat(notification): Telegram add support for topics in groups

yegle:


zze0s (commented):
Not sure if this is ready or not since it's **Open** not a **draft** and no body/description.

If you want topic to work you need to add new migrations to add it to the schema, and then in various repo functions for the database layer. Check [this PR](https://github.com/autobrr/autobrr/pull/598/files#diff-0c95a6d345a031f5d37558fc17644aa002aeae1ee65a0c008e677ef4913ce6f2) which added the new field `priority`.

zze0s (approved):
Thanks for the PR! Made some changes:

* Add the migrations so topic can be stored
* Update Notification repo to store/get topic
* Update forms for initial values

- [x] Tested and looks OK.

image

PR #898, opened 2023-05-01
fix(web): PWA asset and route handling

zze0s:


KaiserBh (approved):

KaiserBh:
It works fine on android and Iphone (xcode simulator) using reverse proxy subdomain. Haven't tested base url.
Also Iphone just says add to home screen as I haven't used Iphone in a hot minute am not sure if it's the same as Android where it says Install. 

https://user-images.githubusercontent.com/41852205/235599932-7b1bca1a-c592-4e9b-994f-16c2487ce382.mp4

s0up4200:
Its on sub dirs like`/autobrr` it'll fail. It doesn't find the icon and doesn't open in a separate full screen Safari window.
The "Add to Home Screen" part is correct though👍

s0up4200 (approved):

PR #900, opened 2023-05-01
fix(config): lxc container detection

fabricionaweb:
The current method to detect lxc and start the config file with host 0.0.0.0 does not work because linux kernel have moved to namespaces and also only root have permissions to access some proc files.

I have checked within all my containers and also with a friend that runs lxc on a different plataform, both of us have this file with same permissions 444.

I also have checked against [lxc source code](https://github.com/lxc/lxc/blob/master/src/lxc/conf.c#L4024).

I would assume its safe to assume this file should exists on every boot.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -107,7 +107,13 @@ func writeConfig(configPath string, configFile string) error {
 			// docker creates a .dockerenv file at the root
 			// of the directory tree inside the container.
 			// if this file exists then the viewer is running
-			// from inside a container so return true
+			// from inside a docker container so return true
+			host = "0.0.0.0"
+		} else if _, err := os.Stat("/dev/.lxc-boot-id"); err == nil {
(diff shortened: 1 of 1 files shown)


KyleSanderson (commented):
Can you restore the existing check as well? It doesn't look like it harms anything, and if the new method works for new, and the old method works for old, that covers more rogue installations out there.

fabricionaweb:
Yes I can, done.

zze0s (approved):

KyleSanderson:
> Yes I can. Done.
> 
> Btw the code that generates this lxc boot file is 4y old, so it maybe works for old containers as well - but I can not test it...
> 
> https://github.com/lxc/lxc/blame/master/src/lxc/conf.c#L4024

I wish we only had to support setups that are 4 years old :-). Docker has been around for 10 years, containers themselves longer. Thanks for the revert, this looks good to me.

PR #908, opened 2023-05-04
enhancement(filters): exclude `tags_match_logic` on export if default value

s0up4200:
Exclude `tags_match_logic` and `except_tags_match_logic` on filter export if default values.

```ts
if (["tags_match_logic", "except_tags_match_logic"].includes(key) && value === "ANY") {
    delete completeFilter[key as keyof CompleteFilterType];
  }
```

**EDIT:** Further testing indicates that its only added to the export if the user has interacted with that field.
Like setting it to `all`, save, then setting it back to `any` again. 

I still think we should remove it if set to `any` as its the default anyway. There is no need to export default values.

But that explains why some exports has it while others don't.


KyleSanderson (approved):
this is probably ok

zze0s (approved):

PR #910, opened 2023-05-05
fix(filters): handle empty `tag_match_logic` by defaulting to `ANY`

s0up4200:
#810 introduced `tags_match_logic` `ANY` and `ALL`

This PR addresses the issue where the `tag_match_logic` is set to an empty string. It will only get the `ANY` value if you change it to `ALL`, save, and then back to `ANY` again.

This PR solves that by simplifying the conditions to check only for the `ALL` case and treat everything else (including empty strings) as `ANY`.

Added trimming of tags in the `containsAllMatch` and `containsMatch` functions, ensuring that any leading or trailing whitespace is removed. This should improve the accuracy and consistency.


zze0s (approved):
Great find!

PR #911, opened 2023-05-05
feat(release): add torrent download link to release table

s0up4200:
Add torrent download link to Release Table. A good interim solution until a more proper function is implemented.

By using an extension like [Torrent Control](https://addons.mozilla.org/en-GB/firefox/addon/torrent-control/), users can easily add torrents to their client if they were rejected or for any other reason. This can be done by simply right-clicking the torrent link and selecting the option to add it to their client.

image


zze0s (approved):
:+1:

PR #915, opened 2023-05-06
fix(downloadclients): Deluge v1 save path

### LINKED ISSUE #727: Deluge v1 `save_path` not working
**Version**
1.18.0

**Describe the bug**
I setup filter, in Action tab, for `Save path` I have like: `/home/myuser/downloads/{{.TorrentName}}`

The path `/home/myuser/downloads` exists.

But in client, the {{.TorrentName}} seems to be empty and ignored.
**To Reproduce**
1. Setup Deluge client
2. Setup filter, then go to Action, set Deluge type, for `Save Path` set to `/home/myuser/downloads/{{.TorrentName}}`
3. Save and make sure that filter is enabled

**Expected behavior**
I expected that path with torrent name is appended at end.

**Screenshots**
image

image

image

 - OS: Windows
 - Browser no matter, I use often Chrome

zze0s:
Fix issue with Deluge v1 not setting `save path`.

Fixes #727

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ require (
 	github.com/Masterminds/squirrel v1.5.3
 	github.com/anacrolix/torrent v1.46.0
 	github.com/asaskevich/EventBus v0.0.0-20200907212545-49d423059eef
+	github.com/autobrr/go-deluge v1.0.0
 	github.com/autobrr/go-qbittorrent v1.3.1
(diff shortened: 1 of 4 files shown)


s0up4200 (approved):
1. Tested with Deluge 1.3.5, set a custom save path in filter: `/home/user/torrents/deluge/{{.TorrentName}}`
2. Torrent successfully added to correct path in the client 🚀

PR #919, opened 2023-05-08
chore(build): change to pnpm from yarn

KyleSanderson:
Autobrr web appears to take this long per product to resolve dependencies:
yarn: 70s
npm: 30s
pnpm: 15s

yarn doesn't result in a cache at all for subsequent runs, which to me is a bug but too much time has already been lost to something as silly as that.

both (p)npm take a cache, and the actual build is around 25s.


KaiserBh (changes requested):
Overall looks good and great change!
That being said need to remove leftovers for yarn since we are not using it anymore.

```prolog
remove .yarnrc.yml
clean up .dockerignore and .gitignore as well.
Fix Dockerfile it still uses yarn which I assume it will break.
clean up package.json: `  "packageManager": "yarn@3.2.2"`  (we don't need this anymore right?)
```

zze0s (approved):
That's a huge performance increase for the web build!

Will be beneficial for local development as well.

Tested local and works as expected :+1:

fabricionaweb:
Instead of remove the `packageManager`, why dont we keep it but set to `pnpm@someversion`?

This way we can just trust in corepack to handles it.

https://nodejs.org/api/packages.html#packagemanager
https://github.com/nodejs/corepack#usage

KyleSanderson:
> Instead of remove the `packageManager` from the package.json, why dont we keep it but set to `pnpm@someversion`?
> 
> This way we can just trust in corepack to handles it.
> 
> https://nodejs.org/api/packages.html#packagemanager https://github.com/nodejs/corepack#usage

In truth, I don't really understand any of this (not your links, but this node stuff in general haha). If there's a better way to accomplish something while still being technically correct a PR or steps would be appreciated. The one caveat being reducing work going forward (we pin versions in GitHub actions, so another pin would be more work, as an example).

PR #920, opened 2023-05-08
feat(logs): improve log search with regex

s0up4200:
Search is now case insensitive and way more flexible thanks to regex.

Searches like `this show`, `ThIs ShOw`, `s[0-9]{2}e[0-9]{2}.*(1080p|720p)` or `The.Company.*(1080p|720p).*-GRP` now works.

Show a red icon if pattern is invalid:
image


zze0s (approved):

martylukyy (approved):
At a later point it would be nice to be able to filter with stuff like network URL.
But for now it is looking fine AF so send it!

PR #924, opened 2023-05-09
feat(releases): delete older than x

s0up4200:
Delete releases older than X time.

Current frontend looks like this:

image
image

### Future improvements: 

- [ ] Add cron for automatic cleaning


web/src/screens/settings/Releases.tsx
```diff
@@ -33,6 +33,45 @@ function ReleaseSettings() {
 
   const cancelModalButtonRef = useRef(null);
 
+  const getDurationLabel = (durationValue: number): string => {
+    const durationOptions: Record<number, string> = {
```
stacksmash76, line 37: Should add a "DELETE EVERYTHING EVER" option (in caps) and get rid of the old "Delete all releases" button, as someone will surely get confused and fuck up. It could occupy the value 0, for example.

web/src/screens/settings/Releases.tsx
```diff
@@ -18,10 +18,10 @@ function ReleaseSettings() {
   const queryClient = useQueryClient();
 
   const deleteMutation = useMutation({
-    mutationFn:  APIClient.release.delete,
+    mutationFn: APIClient.release.delete,
```
stacksmash76, line 21: I saw you did this in previous PRs, but was too lazy to comment, but I'll write it just so you know: You should wrap your queryFn's/mutationFn's in a wrapper function because you will get argument tainting that /one/ time and you'll be scratching your head for a bit until you figure out what's the issue. TS will surely help, but don't rely solely on it.

Put in another way `mutationFn: (variables: TVariables) => Promise<TData>`, the `variables` parameter will get passed on `APIClient.release.delete` and if that function takes an argument, you might not get what you want.

Not an issue in this case, but definitely not a great practice.

web/src/screens/settings/Releases.tsx
```diff
@@ -66,7 +105,38 @@ function ReleaseSettings() {
...
+            </div>
+            <div className="mt-6">
+              <label htmlFor="duration" className="items-center block text-sm font-medium text-gray-700 dark:text-white text-center">
+                Delete releases older than:
+              </label>
+              <div className="flex justify-between items-center p-2 mt-2 max-w-sm m-auto mt-1 rounded-md shadow-sm">
+                <select
+                  name="duration"
+                  id="duration"
+                  className="focus:outline-none focus:ring-2 focus:ring-offset-0 focus:ring-blue-500 dark:focus:ring-blue-500 flex-1 block w-full rounded-none rounded-r-md sm:text-sm border-gray-300 dark:border-gray-600 dark:bg-gray-800 dark:text-white"
```
stacksmash76, line 118: Also forgot to comment this one, the dropdown/select element will have a rounded corners only on the right side, but not on the left.

web/src/screens/settings/Releases.tsx
```diff
@@ -66,7 +105,38 @@ function ReleaseSettings() {
               <h3 style={{ textAlign: "center" }} className="text-lg leading-6 font-medium text-gray-900 dark:text-white">
                 Danger Zone
               </h3>
-              <p style={{ textAlign: "center" }} className="mt-1 text-sm text-gray-900 dark:text-white">This will clear all release history in your database.</p>
+              <p style={{ textAlign: "center" }} className="mt-1 text-sm text-gray-900 dark:text-white">This will clear release history in your database.</p>
+            </div>
+            <div className="mt-6">
+              <label htmlFor="duration" className="items-center block text-sm font-medium text-gray-700 dark:text-white text-center">
```
stacksmash76, line 111: Another thing, I think it would be better if this text was left-aligned as opposed to centered.

s0up4200: Thanks for all the feedback and suggestions as always 👍 

Can you show me a screenshot of what you mean perhaps? I'm not sure if you want the text all the way to the left. It'd be great to confirm!

stacksmash76: Was referring to the label "Delete releases older than:", so it's in line with the dropdown element text. It's a personal preference, however.

Looking at your screenshot again, might also be nice to change the subtitle on the panel page from "Release settings. Reset state." to something like "Manage release history" or something similar.

s0up4200: Pushed the changes you suggested. New screenshot in the OP!

stacksmash76: Noice!

zze0s (approved):
Very nice! Well done! :rocket:

PR #931, opened 2023-05-13
chore(build): Use pnpm in Makefile

Shraymonks:
Update to use pnpm in Makefile. Looks like this was missed in #919.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,4 +39,5 @@ tmp/
 package-lock.json
 # Ditto for yarn, except we're using npm.
 yarn.lock
-dist/
+dist/*
+!dist/.gitkeep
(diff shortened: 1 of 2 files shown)


KyleSanderson (approved):

web/dist/.gitkeep
zze0s, line 1: You need to add this file back. The backend requires it to have the embedding of the frontend work, but it's fine being an empty dir for testing.

zze0s (changes requested):
You need to add back the `/web/dist/.gitkeep` file. It's why the tests are failing.

KyleSanderson (approved):

zze0s (approved):

PR #932, opened 2023-05-14
feat(releases): replay actions

### LINKED ISSUE #518: [FEATURE REQ] Allow actions to be re-run
I believe there's a missed opportunity when something has triggered a filter which contains multiple actions (rtorrent+radarr for example). For instance, it's something new and it was not on my radarr. I would like to be able to add that to radarr (manually via radarr's ui) and then re-run the radarr action.

From the Dashboard page, the pop-up when I click a rejected action could have a set of steps, for instance "retry".

Currently, there's no clean alternative that I can think of. I still would have to go to radarr, add whatever it is, wait for it to be finished and then use manual import. I

**Additional context**
Another similar example would be to test if a new filter could match an already matched approved push. This would be helpful to avoid "breaking a filter"

zze0s:
Allow actions to be replayed.

Closes #518

### DIFF
--- a/internal/action/run.go
+++ b/internal/action/run.go
@@ -99,7 +99,7 @@ func (s *service) RunAction(ctx context.Context, action *domain.Action, release 
 	payload := &domain.NotificationPayload{
 		Event:          domain.NotificationEventPushApproved,
 		ReleaseName:    release.TorrentName,
-		Filter:         release.Filter.Name,
+		Filter:         release.FilterName,
 		Indexer:        release.Indexer,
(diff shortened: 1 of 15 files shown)


s0up4200 (commented):
Tested replay with both bad and good client credentials. Both work as you'd expect 👍 

One thing though, when I tried to update a filter that existed before the db migration, I got the following error:
image

1. Opened existing filter
2. Set max 2 downloads per hour
3. Hit save

s0up4200 (approved):
🚀

PR #935, opened 2023-05-15
build(web): use pnpm from corepack

fabricionaweb:
[Corepack](https://nodejs.org/api/corepack.html) is enable since node v14, it has the package manager built-in and we can just use it.

When using corepack it will respect the version at package.json but it needs to be locked version.

I also think its nice to lock node-version on the github actions to the same as the docker image is using.

I have not tested the github actions changes, just Google and it seems to work, but we need to run to test it.

For Dockerfile I have tested and it works fine:

image

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,8 +30,8 @@ jobs:
       - name: Set up Node
         uses: actions/setup-node@v3
         with:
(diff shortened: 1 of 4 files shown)


zze0s (approved):
Looks like a good improvement with some speed-up! :+1:

PR #944, opened 2023-05-17
feat(indexers): add API support for Orpheus to fetch size

zze0s:
Add API support for Orpheus to get torrent size.

## Future goal

Expose and filter by uploader.


internal/domain/indexer.go
```diff
@@ -339,6 +339,7 @@ type TorrentBasic struct {
 	TorrentId string `json:"TorrentId,omitempty"`
 	InfoHash  string `json:"InfoHash"`
 	Size      string `json:"Size"`
+	Uploader  string `json:"Uploader"`
```
KyleSanderson, line 342: is this relevant to any other API members?

zze0s: Will be for RED, OPS to start with. Maybe PTP later!

PR #948, opened 2023-05-19
feat(irc): view announces per channel

### LINKED ISSUE #826: [FEATURE REQ] View IRC channel
Would like to be able to view the IRC channel raw i.e. see all releases in autobrr directly.

zze0s:
View messages/announces per channel of IRC networks.

With the improved functionality in the Server Sent Events library we can now store events from channels and replay those easily. The current number is set to replay 1000 messages when opening the channel log. This might be too many messages but they are streamed and not a giant blob. We can always reduce this in the future.

## Screenshots

image

Expanded
image

Fixes #826

### DIFF
--- a/cmd/autobrr/main.go
+++ b/cmd/autobrr/main.go
@@ -56,8 +56,7 @@ func main() {
 
 	// setup server-sent-events
 	serverEvents := sse.New()
-	serverEvents.AutoReplay = false
-	serverEvents.CreateStream("logs")
+	serverEvents.CreateStreamWithOpts("logs", sse.StreamOpts{MaxEntries: 1000, AutoReplay: true})
 
 	// register SSE hook on logger
 	log.RegisterSSEHook(serverEvents)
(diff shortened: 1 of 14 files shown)


s0up4200 (approved):

stacksmash76:
You're not using IrcContext (plus onSet is a no-op in a way) and you should never omit parentheses when instantiating a new class. (`new Map()`)

zze0s:
> You're not using IrcContext (plus onSet is a no-op in a way) and you should never omit parentheses when instantiating a new class. (`new Map()`)

You are correct :) Started out trying to build an irc client but bailed midway since it's more complexity than we currently need. This should probably be a draft since there's still more changes coming.

KyleSanderson:
> > You're not using IrcContext (plus onSet is a no-op in a way) and you should never omit parentheses when instantiating a new class. (`new Map()`)
> 
> This should probably be a draft since there's still more changes coming.

That's well communicated, so no worries 😺.

PR #950, opened 2023-05-21
feat(releases): torrent file downloads improve error handling

s0up4200:
checks if torrent file is a valid torrent file when content-type is text/html


KyleSanderson (approved):

internal/domain/release.go
```diff
@@ -363,10 +365,10 @@ func (r *Release) DownloadTorrentFile() error {
 }
 
 func (r *Release) downloadTorrentFile(ctx context.Context) error {
-	if r.Protocol != ReleaseProtocolTorrent {
-		return errors.New("download_file: protocol is not %s: %s", ReleaseProtocolTorrent, r.Protocol)
-	} else if r.HasMagnetUri() {
-		return fmt.Errorf("error trying to download magnet link: %s", r.MagnetURI)
+	if r.HasMagnetUri() {
+		return errors.New("downloading magnet links are not supported: %s", r.MagnetURI)
```
KyleSanderson, line 369: is not*

internal/domain/release.go
```diff
@@ -423,58 +425,72 @@ func (r *Release) downloadTorrentFile(ctx context.Context) error {
 			// Continue processing the response
 		case http.StatusMovedPermanently, http.StatusFound, http.StatusSeeOther, http.StatusTemporaryRedirect, http.StatusPermanentRedirect:
 			// Handle redirect
-			return retry.Unrecoverable(errors.New("redirect encountered for torrent (%v) file (%v) from '%v' - status code: %d. Check indexer keys.", r.TorrentName, r.TorrentURL, r.Indexer, resp.StatusCode))
+			return retry.Unrecoverable(errors.New("redirect encountered for torrent (%v) file (%v) - status code: %d - check indexer keys for %s", r.TorrentName, r.TorrentURL, resp.StatusCode, r.Indexer))
```
KyleSanderson, line 428: Redirects are a great way to see a user visible error, but they're also still valid for a tracker announcing on one url during a domain flip. These events are pretty rare in comparison, but just wanted to mention it.

internal/domain/release.go
```diff
@@ -423,58 +425,72 @@ func (r *Release) downloadTorrentFile(ctx context.Context) error {
...
-			return errors.Wrap(err, "metainfo could not load file contents: %v", tmpFile.Name())
+
+			// explicitly check for unexpected content type that match html
+			var bse *bencode.SyntaxError
+			if errors.As(err, &bse) {
+				// regular error so we can retry if we receive html first run
+				return errors.Wrap(err, "metainfo unexpected content type, got HTML expected a bencoded torrent. check indexer keys for %s - %s", r.Indexer, r.TorrentName)
+			}
+
+			return retry.Unrecoverable(errors.Wrap(err, "metainfo unexpected content type. check indexer keys for %s - %s", r.Indexer, r.TorrentName))
```
KyleSanderson, line 475: is this really fatal? shouldn't we retry.

internal/domain/release.go
zze0s: That's a valid point. Then we should probably not return err on `StatusMovedPermanently, StatusTemporaryRedirect, StatusPermanentRedirect`?

internal/domain/release.go
zze0s: The bencode.SyntaxError is what happens when it tries to parse `<html>..</html>` and that is handled above without unrecoverable so it can retry if for some reason it gets different content first time because it's too fast or something.

The last err is for all other cases, and there is a lot of other errors metainfo can return but all of them are bad. `Metainfo` also does a horrendous amount of panics internally 👀

KyleSanderson: I wrote this eons ago, maybe it's time for it's debut.

```
func Validate(data string) error {
	depth := 0
	buf := strings.NewReader(data)
	for {
		b, err := buf.ReadByte()
		if err == io.EOF {
			break
		}

		switch b {
		case 'd', 'l':
			depth++
		case 'e':
			depth--
			if depth < 0 {
				return fmt.Errorf("Unexpected depth at %d: %c", depth, b)
			}
		case 'i':
			init := true
			for {
				i, err := buf.ReadByte()
				if err == io.EOF && depth != 0 {
					return fmt.Errorf("Unexpected end to integerstream at a depth of: %d", depth)
				}

				if init {
					init = false
					if i == '-' {
						if k, err := buf.ReadByte(); k == '0' || err != nil {
							return fmt.Errorf("Unexpected value in integerstream at a depth of: %d| %d", depth, k)
						}

						buf.UnreadByte()
						continue
					}
				}

				if i < '0' || i > '9' {
					if i == 'e' {
						break
					}

					return fmt.Errorf("Unexpected end to integerstream at a depth of: %d | %c", depth, i)
				}
			}
		case '0', '1', '2', '3', '4', '5', '6', '7', '8', '9':
			i := int64(b - '0')
			for {
				ru, err := buf.ReadByte()
				if err == io.EOF && depth != 0 {
					return fmt.Errorf("Unexpected end to bytestream at a depth of: %d", depth)
				}

				k := int64(ru - '0')
				if k < 0 || k > 9 {
					if ru == ':' {
						break
					}

					return fmt.Errorf("Unexpected length in bytestream at a depth of: %d | %c", depth, ru)
				}

				i *= 10
				i += k
			}

			buf.Seek(i, io.SeekCurrent)

		default:
			return fmt.Errorf("Unexpected character in bytestream at a depth of: %d | %c", depth, b)
		}
	}

	if depth != 0 {
		return fmt.Errorf("Unexpected depth at end of stream: %d", depth)
	}

	return nil
}
```

internal/domain/release.go
KyleSanderson: I don't think so, we should parse the results still but log that it was a redirect. I don't know what happens when we tell the go httpclient to follow redirects.

zze0s: It will follow (10) redirects by default, so that should be fine.

PR #952, opened 2023-05-22
fix(PWA): Don't intercept /api route

KyleSanderson:


KaiserBh:
This breaks local development, also that proxy is just for local dev, it doesn't do anything to the production/build one.

image

fabricionaweb:
Mind to remove the [setupProxy.js](https://github.com/autobrr/autobrr/blob/develop/web/src/setupProxy.js) as well? 🥺

web/vite.config.ts
```diff
@@ -89,13 +89,6 @@ export default ({ mode }: { mode: any }) => {
       hmr: {
         overlay: true
       },
-      proxy: {
-        "/api": {
-          target: "http://127.0.0.1:7474/",
-          changeOrigin: true,
-          secure: false
-        }
-      }
```
KaiserBh: This needs to be added back for frontend local development, otherwise we have to recompile frontend every time we make a change.

KaiserBh (changes requested):

KaiserBh (approved):
Looks good now.

martylukyy (approved):
frontend dev mode works again and RSS feed issues are fixed. Thanks Kyle!
(Note: hard reload needed to make `View latest run` work again)

fabricionaweb (approved):

PR #956, opened 2023-05-22
build: goreleaser parallel builds

KyleSanderson:
So... goreleaser does not allow opensource users to use GGOOS (which lets you filter the goreleaser file) so matrix builds are possible but it's constructing the file each time (not a problem, just work).

The linux nodes on GHA only have a single CPU which is fine, but takes a lot of time with cross-compilation. MacOS and Windows still run on (or close to) metal. Windows with setup-go from years ago still suffers from zstd compression times (taking 4 minutes), with a pretty beefy node. MacOS runs in a similar fashion, but uses gzip which decompresses faster (3 minutes) to be exact.

As these are both faster than linux... lets change nodes until such a time that this no longer makes sense, or there's an easier way.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -74,7 +74,7 @@ jobs:
   goreleaserbuild:
     name: Build Go binaries
     if: github.event_name == 'pull_request'
-    runs-on: windows-latest
+    runs-on: macos-latest
     needs: [web, test]
     steps:
       - name: Checkout
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,7 +16,7 @@ permissions:
 jobs:
   web:
     name: Build web
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


zze0s (approved):
:rocket:

.github/workflows/release.yml
```diff
@@ -100,15 +100,17 @@ jobs:
...
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Upload assets
         uses: actions/upload-artifact@v3
         with:
           name: autobrr
-          path: dist/*.tar.gz
+          path: |
+            dist/*.tar.gz
+            dist/*.zip
```
fabricionaweb, line 113: Ok a little bit later to the party but why does we need the .zip?

.goreleaser.yml
```diff
@@ -67,6 +67,9 @@ archives:
       - none*
     replacements:
       amd64: x86_64
+    format_overrides:
+      - goos: windows
+        format: zip
```
fabricionaweb, line 72: ah ok... I see...

PR #957, opened 2023-05-23
chore(web): relocations and cleanups

fabricionaweb:
We are having a warn message from Vite because having only one thick js file isnt ideal.

before

and it will continue complaing


web/vite.config.ts
```diff
@@ -1,38 +1,30 @@
...
       useCredentials: true,
-      includeAssets: [
-        "favicon.svg",
-        "favicon.ico",
-        "robots.txt",
-        "logo.png",
-        "apple-touch-icon-*.png",
-        "manifest.webmanifest",
-        "assets/**/*"
-      ],
```
fabricionaweb, line 28: I see the docs, the `includeAssets` by default already gets public folder and icons... I tested with and without this code and the generated manifest had no changes.

web/vite.config.ts
```diff
@@ -98,7 +90,25 @@ export default ({ mode }: { mode: any }) => {
     },
     build: {
       manifest: true,
-      sourcemap: true
+      sourcemap: true,
+      rollupOptions: {
+        output: {
+          manualChunks (id: string) {
```
fabricionaweb, line 96: This `manualChunks` is kinda weird, but without it, Vite by default only splits `index` and `vendor`, which isnt good enough to me.

web/vite.config.ts
fabricionaweb: Actually let me do more tests

web/pnpm-lock.yaml
```diff
@@ -159,7 +159,7 @@ packages:
     dependencies:
       '@jridgewell/gen-mapping': 0.3.3
       '@jridgewell/trace-mapping': 0.3.18
-    dev: false
+    dev: true
```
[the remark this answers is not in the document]
fabricionaweb: Because I moved the vite-plugin to devDependencies

web/vite.config.ts
fabricionaweb: Yes, but to be honest, I could not understand well 🤓

And the `manualChunks` does not work without the plugin

web/pnpm-lock.yaml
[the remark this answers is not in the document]
fabricionaweb: To be honest we dont follow strategy about what is dev or not in this app, as the whole web will always need to build and the runtime is done without any node dependencies. Maybe we should move everything to devDependencies (or dependencies). 

That makes sense to me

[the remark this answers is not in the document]
fabricionaweb: I will drop this commit and we see/decide it in another MR

zze0s (commented):
I think it's worth exploring using lazy imports in the router to only load per page. https://react.dev/reference/react/lazy

stacksmash76:
Late to the party, but my 5 cents:
- React Suspense should be implemented first before lazy loading.
- Too much chunking is a bad practice (relative to project size, of course; you don't want a 1 MB file split across more than, say, 10 files). You've got ping roundabout (out & back) plus HTTP overhead response plus V8 doing it's thang for every little navigation [the browser can't burst the requests in some cases, see below]. Also, cache busting can be tedious/a nuisance.
- When going the React.lazy route, only the main chunks are embedded into the HTML (i.e. manualChunks in this instance), which is a bullet-proof way of loading JS. Problems occur when you load chunks from JS. There are quite a few browser bugs, with differing errors (source: ancedata from one of my Sentry instances where I experimented with this; sadly, deleted, can't give you concrete/concise examples). This happens on Chrome and Safari predominantly. For Firefox this is a rarer occurrence (when normalized to sampled distribution).

Most of the time you're serving gzip (or in case of newer browsers brotli anyways). So, improve terser/optimizer to conform to the GUWOP web standard:
- 3 compression passes
- mangle symbols (don't keep class/function names)
- don't output comments (well, minus licenses OR bundle them separately, somewhere else than in the source code)
- ditch IE
- Safari 10 also goes out, probably along with 11

Additionally, generate gzip/brotli-encoded chunks:
a) on the run from embedded files (as in a cache warm-up round; smaller binary size [by maybe 2MB? + gz/br encoding algorithms]) and serve from memory
b) during frontend compilation and bundle in the binary along with the non-encoded JS files. You need to respect the `Accept-Encoding` header in that case.

stacksmash76:
The splitting isn't great. Loading small chunks (<10kB) is almost always a waste (additional pointless packet roundabout time) vs. initially loading it with a bigger chunk.

Check out `vite-bundle-visualizer` (or a similar tool) to inspect the bundle. It doesn't make much sense to split on pages which consume a lot of dependency code, which gets bundled to a different chunk.

When you just mindlessly split on pages, you don't get to pick what will get bundled and how. You're pretty much leaving everything to chance, and the chances are, you aren't going to get what you expect or hope for.

node_modules consumes 70% of bundle size. Other 30% come from our code. 99% of the time you can serve gzip. Our code weighs only ~112kB gzipped. Do we split 112kB into chunks? Doesn't make sense.

A far bigger improvement would be a better optimization pass + gzip/brotli guaranteed by our backend (instead of hoping a reverse proxy will handle it for us).

To be more concrete with my comment, React lazy is fine, but don't overdo it. And you should go back to manualChunks since they'll do what you need them to do.

PS. Also, why are you _constantly_ force-pushing?

zze0s:
@fabricionaweb could you update this and resolve the conflicts whenever you get some free time? There's no rush but if it could be done in the upcoming 2-3 weeks that would be good :smile:

PR #965, opened 2023-05-28
chore(web): add ignore pattern to eslint for the unused-vars rule

fabricionaweb:
I know we dont need to be crazy about linting rules... but...

The idea is to allow the use of `_` or `_myVariable` when you **really** need an unused variable (like for skip a paramether) this way eslint wont complaing to you

There are more files to be updated but those files is being touched in another MR so I prefer to leave out for now avoiding possible conflict.


zze0s (approved):
:+1:

PR #967, opened 2023-06-01
fix(indexers): enthralled IRC network

KyleSanderson:


internal/indexer/definitions/enthralled.yaml
```diff
@@ -26,8 +26,8 @@ settings:
     help: Right click DL on a torrent and get the torrent_pass.
 
 irc:
-  network: DigitalIRC
-  server: irc.enthralled.me
+  network: digitalirc.org
```
zze0s, line 29: We have 3 using network name `DigitalIRC` and MTV using `digitalirc.org`. Name are mostly visual, I think :thinking: 

Maybe we should change MTV to be consistent with others on the same network.

KyleSanderson: I don't remember which I copied, but they should all be the same.

zee0s (approved):

zze0s (approved):

Issue #966, opened 2023-05-31
[INDEXER REQ] Sharewood


Mavyre:
**We currently only support indexers with IRC announces.**

# Indexer request

-  Do they have irc announce? Yes
-  Link to existing `autodl-irssi .tracker` file: Not existing, but existing YML prowlarr tracker file: https://github.com/Prowlarr/Indexers/blob/master/definitions/v7/sharewood.yml

### If no existing `autodl-irssi .tracker` file

- Indexer name: Sharewood
- Short name: SW
- URL: https://www.sharewood.tv
- Download link examples: https://www.sharewood.tv/download/[torrent-slug].[torrent-id] (needs to be authenticated before downloading) OR https://sharewood.tv/api/[passkey]/[torrent-id]/download (download using passkey, without authentication)

## Announce examples

Please provide 1-5 different announces. And change release titles for good measure. If they have tags like freeleech and internals etc make sure to get examples of those.

- Vidéos | La Brea S02E08 MULTi 1080p WEB x264-FW | 1.27 GiB | 1080p/i <https://sharewood.tv/torrents/la-brea-s02e08-multi-1080p-web-x264-fw.66870>

Other informations:

The announce format is "Category | Torrent Name | Size | Resolution <Link>"
The link is formatted as https://sharewood.tv/torrents/torrent-slug.torrent-id, changing it to https://sharewood.tv/download/torrent-slug.torrent-id allows to download it directly, if the user is already logged in on the tracker (see the Prowlarr yml). Using only the torrent-id and user passkey should allow torrent direct download without authentication

## IRC Network

- Network name: Hiddenforest
- Network address: irc.hiddenforest.ml
- Port: 6667
- SSL/TLS: No
- NickServ registration required? Check wiki, guides etc: No
- Is there a required format for the nick? No
- Invite command: None

## Bonus

| Category on site/announce | ID | Category              |
|---------------------------|----|-----------------------|
| Vidéo                     |  1 | Videos (Movies/TV)    |
| Audio                     |  2 | Audio (Music)         |
| Application               |  3 | Apps                  |
| Ebooks                    |  4 | Ebooks                |
| Jeu-Vidéo                 |  5 | Video Games           |
| Formation                 |  6 | Tutorials / Trainings |
| XXX                       |  7 | P0rn                  |

Tracker API:

- Latest torrents: `https://www.sharewood.tv/api/[passkey]/last-torrents`
[...]

PR #968, opened 2023-06-01
feat(indexers): add Sharewood

Mavyre:
Added the French Sharewood indexer on autobrr. Closes my own Request #966.

### DIFF
--- a/internal/indexer/definitions/sharewood.yaml
+++ b/internal/indexer/definitions/sharewood.yaml
@@ -1,0 +1,69 @@
+---
+name: Sharewood
+identifier: sharewood
+description: Sharewood (Sw) is a Private French Torrent Tracker for MOVIES / TV / GENERAL
+language: fr-fr
+urls:
+  - https://www.sharewood.tv/
+privacy: private
+protocol: torrent
+supports:
+  - irc
+  - rss
+source: custom
+settings:
+  - name: passkey
+    type: secret
+    required: true
+    label: Passkey
+    help: "Aller sur Mon Profil, puis cliquer sur Afficher Passkey"
+
+irc:
+  network: HiddenForest
+  server: irc.hiddenforest.ml
+  port: 7000
+  tls: true
+  channels:
+    - "#rss"
+  announcers:
+    - Lana
+  settings:
+    - name: nick
+      type: text
+      required: true
+      label: Nick
+      help: Bot nick. Eg. user_bot
+
+    - name: auth.account
+      type: text
+      required: false
+      label: NickServ Account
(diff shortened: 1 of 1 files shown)


internal/indexer/definitions/sharewood.yaml
```diff
@@ -0,0 +1,69 @@
...
+supports:
+  - irc
+  - rss
+source: custom
+settings:
+  - name: passkey
+    type: secret
+    required: true
+    label: Passkey
+    help: "Aller sur Mon Profil, puis cliquer sur Afficher Passkey"
```
s0up4200, line 19: This should be in English.

```suggestion
    required: true
    label: Passkey
    help: "Go to My Profile, then click on Show Passkey"
```

zze0s:
> By the way, it seems that by default, it uses SASL authentication. From my tests, the authentication doesn't work using SASL but only NickServ. Is there any way to force it in the definition?

There's no way to do that currently, but it's something I've been thinking of adding for some time!

zze0s (approved):

Mavyre:
> There's no way to do that currently, but it's something I've been thinking of adding for some time!

Can we add a tooltip or a disclosure somewhere to tell the user to change from SASL to NickServ?

zze0s:
> Can we add a tooltip or a disclosure somewhere to tell the user to change from SASL to NickServ?

Could do it in the NickServ Account help text. `Change Mechanism from SASL to NickServ` or if you create a setup guide on the indexers wiki/forum and point it out.

Mavyre:
@zze0s It seems like Sonarr/Radarr both tend to refuse the release pushed by autobrr, because they don't recognise the language. Is it possible, like in prowlarr, to consider MULTi releases as some language? In Prowlarr, I had to setup "Replace MULTi by this language" to MULTi FRENCH to allow Sonarr/Radarr to accept the language upgrade.

Should something like that be configurable in autobrr for "foreign" indexers?

zze0s:
> @zze0s It seems like Sonarr/Radarr both tend to refuse the release pushed by autobrr, because they don't recognise the language. Is it possible, like in prowlarr, to consider MULTi releases as some language? In Prowlarr, I had to setup "Replace MULTi by this language" to MULTi FRENCH to allow Sonarr/Radarr to accept the language upgrade.
> 
> Should something like that be configurable in autobrr for "foreign" indexers?

We currently have nothing in place to be able to do that but there's other indexers that need similar features as being able to hardcode categories and other values, as well as parse titles differently. It's on the todo but no ETA currently.

zze0s:
Related issue #1

PR #970, opened 2023-06-03
feat(web): use svg logo instead of png

fabricionaweb:


KyleSanderson:
SVG gets weird on some browsers (or was the case many many many moons ago). HN recently tried this, and it looks like it's been retained: https://news.ycombinator.com/item?id=35894326

zze0s (approved):
:+1:

PR #978, opened 2023-06-07
feat(downloadclients): Porla support preset

vktr:
Some improvements to the Porla integration,

- Omit download and upload limits completely if they're set to 0.
- Prefix JWT with `Bearer ` if not already specified.
- Allow setting a preset.

Re: the preset changes - I'm reusing the label field from other clients since we don't have that concept in Porla and it seemed easier than introducing a whole new field. Lemme know if this was bad 😃

Also, I'm not sure of the idiomatic Golang way of handling pointers when I want nil, specifically the download and upload limits. If there's a better way, I'm all for it.

### DIFF
--- a/internal/action/porla.go
+++ b/internal/action/porla.go
@@ -64,12 +64,19 @@ func (s *service) porla(ctx context.Context, action *domain.Action, release doma
 		uploadLimit = &ulValue
 	}
 
+	var preset *string = nil
+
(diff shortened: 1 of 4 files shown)


zze0s (approved):
Great improvements! Null values can be a bit of a pain but this should do!

The token check for Bearer is 🏅

PR #983, opened 2023-06-13
feat(indexers): add freeleech support for TS

martylukyy:
This PR adds freeleech to the regex and the vars for new TS announce format.

Since there is no seperator between the username and freeleech,
i had to get a bit creative and need to change the group for the uploader var to `\w` which is the character class `[a-zA-Z0-9_]`.
Hope this doesn't come back to get us, but i didn't see usernames with any other characters upon a quick look.

https://regex101.com/r/Y3n2Vu/1


s0up4200 (approved):

martylukyy:
After this PR has been merged and a new version is out this can be merged too:
https://github.com/autobrr/autobrr.com/pull/111

zze0s (approved):

PR #989, opened 2023-06-16
feat(web): improve irc view 

fabricionaweb:
- inherit text-sm and cursor-point to avoid repeated code
- add mouse hover and event click to the entire row (it had a small gap that wasnt being triggered)
- fix react key to use elements.id instead of array position (also removed repeated keys)
- fix little square corner on the scrollbar


KyleSanderson:
Can you do a before/after picture?

zze0s (approved):
:+1:

PR #990, opened 2023-06-16
fix(irc): urlencode SSEKey for SSE streams

fabricionaweb:
Apply RFC4648 to base64 sseKey to fix urlEncoding of SSE stream key.


martylukyy (approved):
Fixes problem with BTF indexer not showing any announces due to characters that are not allowed in the URL.

EDIT: Ran this PR on live for some hours now. All my indexers have data in them now.
Looks good to be merged. :)

PR #991, opened 2023-06-16
feat(filter): add rejection on invalid size set

s0up4200:
If user sets a min or max size without specifying a unit, the rejection returns blank

### DIFF
--- a/internal/domain/filter.go
+++ b/internal/domain/filter.go
@@ -551,7 +551,7 @@ func (f Filter) checkSizeFilter(r *Release, minSize string, maxSize string) bool
 		// string to bytes
 		minSizeBytes, err := humanize.ParseBytes(minSize)
 		if err != nil {
-			// log could not parse into bytes
+			r.addRejection("size: invalid minSize set")
 		}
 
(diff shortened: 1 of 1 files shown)


internal/domain/filter.go
```diff
@@ -551,7 +551,7 @@ func (f Filter) checkSizeFilter(r *Release, minSize string, maxSize string) bool
 		// string to bytes
 		minSizeBytes, err := humanize.ParseBytes(minSize)
 		if err != nil {
-			// log could not parse into bytes
+			r.addRejection("size: invalid minSize set")
```
KyleSanderson, line 554: Shouldn't these return?

s0up4200: Oopsie, you are probably right.

zze0s: Yeah, return false would be a good idea.

So maybe change to this:

```diff
-			r.addRejection("size: invalid minSize set")
+			r.addRejectionF("size: invalid minSize set: %s err: %q", minSize, err)
+			return false
```

And same for max size.

zze0s (approved):

PR #883, opened 2023-04-26
fix(releases): max downloads per hour

zze0s:
Speculative fix for max downloads per hour for docker.

PR #993, opened 2023-06-18
fix(releases): delete older than X

s0up4200:
Resolves a bug in the `DeleteOlder` function where recent 24-hour data wasn't deleted when set to delete everything. We now correctly set the `olderThanTimestamp` to a future date when duration is zero, ensuring complete deletion of all records.


s0up4200:
Another option is to do this I guess:

```go
olderThanTimestamp := time.Now().UTC().Add(-time.Duration(duration) * time.Hour)

if duration == 0 {
	olderThanTimestamp = time.Unix(1<<63-62135596801, 999999999)
}
```

internal/database/release.go
```diff
@@ -594,7 +594,14 @@ func (repo *ReleaseRepo) DeleteOlder(ctx context.Context, duration int) error {
...
+
+	if duration == 0 {
+		olderThanTimestamp = time.Now().UTC().AddDate(100, 0, 0) // set future date to delete everything
+	} else {
+		olderThanTimestamp = time.Now().UTC().Add(-time.Duration(duration) * time.Hour)
+	}
+
 	log.Debug().Msgf("Deleting releases older than: %v", olderThanTimestamp)
 
 	result, err := tx.ExecContext(ctx, `DELETE FROM "release" WHERE timestamp < $1`, olderThanTimestamp)
```
KyleSanderson, line 607: This still isn't correct. It needs to go through the process that was done for limits.

s0up4200: Can you elaborate in soup terms?

KyleSanderson: https://github.com/autobrr/autobrr/pull/883/files all those casts are needed unfortunately.

zze0s (approved):
Made some changes and now it's good to go!

PR #995, opened 2023-06-19
fix(config): load from env vars

### LINKED ISSUE #969: Environment variables are not picked up unless option is defined in `config.toml`
**Version**
1.26.2

**Describe the bug**

Env vars are not picked up unless setting is defined in `config.toml`

**To Reproduce**

1. Set a env var that is not defined or commented out in the `config.toml`
2. Notice how the option does not get set

**Expected behavior**

Any env var should be able to be set wether defined in the `config.toml` or not.

zze0s:
Manually map and override non-empty env vars to config keys.

Hardcoded prefix as before `AUTOBRR__` (double underscore!).

Available ENV vars:

- `HOST` - string
- `PORT` - string
- `BASE_URL` - string
- `LOG_LEVEL` - string
- `LOG_PATH` - string
- `LOG_MAX_SIZE` - string/int without MB
- `LOG_MAX_BACKUPS` - string/int
- `SESSION_SECRET` - string
- `CUSTOM_DEFINITIONS` - string
- `CHECK_FOR_UPDATES` - bool
- `DATABASE_TYPE` - sqlite/postgres
- `POSTGRES_HOST` - string
- `POSTGRES_PORT` - string
- `POSTGRES_DATABASE` - string
- `POSTGRES_USER` - string
- `POSTGRES_PASS` - string
- `POSTGRES_SSLMODE` - string
- `POSTGRES_EXTRA_PARAMS` - string

Fixes #969


internal/config/config.go
```diff
@@ -203,13 +205,100 @@ func (c *AppConfig) defaults() {
...
-func (c *AppConfig) load(configPath string) {
-	// or use viper.SetDefault(val, def)
-	//viper.SetDefault("host", config.Host)
-	//viper.SetDefault("port", config.Port)
-	//viper.SetDefault("logLevel", config.LogLevel)
-	//viper.SetDefault("logPath", config.LogPath)
+func (c *AppConfig) loadFromEnv() {
+	prefix := "AUTOBRR__"
+
+	if v := os.Getenv(prefix + "HOST"); v != "" {
```
KyleSanderson, line 211: https://pkg.go.dev/os#Environ should spin through these, and hasprefix.

PR #997, opened 2023-06-24
fix(logs): release rejections cap line length at 1KB

KyleSanderson:


internal/domain/release.go
```diff
@@ -472,7 +472,7 @@ func (r *Release) downloadTorrentFile(ctx context.Context) error {
 				return errors.Wrap(err, "metainfo unexpected content type, got HTML expected a bencoded torrent. check indexer keys for %s - %s", r.Indexer, r.TorrentName)
 			}
 
-			return retry.Unrecoverable(errors.Wrap(err, "metainfo unexpected content type. check indexer keys for %s - %s", r.Indexer, r.TorrentName))
+			return errors.Wrap(err, "metainfo unexpected content type. check indexer keys for %s - %s", r.Indexer, r.TorrentName)
```
zze0s, line 475: I don't see why this should be changed from unrecoverable so it returns?

the `errors.As(err, &bse)` check will trigger when it gets html, and then we retry. If it's something else it should break and return.

zze0s (approved):
:+1:

PR #1011, opened 2023-07-10
fix(web): view button alignment in irc settings

martylukyy:
This PR fixes the View button on the IRC settings page sticking out to the right if the viewport is too narrow.

Old:
image

New:
image


s0up4200 (approved):
👍

zze0s (approved):

Issue #1016, opened 2023-07-17
F5'ing dark mode results in white flash


Silent-Remux:
**Version**
v1.27.1

**Describe the bug**

F5 the autobrr window and you get a momentary white flash before the dark mode activates.

Silent-Remux:
> Ok thanks, in Firefox it happened instantly. I will look at

Thank you!

Silent-Remux:
It would also be good if the colors could be more pronounced, a harsher green and a red for reject would be good.

PR #1017, opened 2023-07-17
fix(web): set theme before app loads

fabricionaweb:
It have been report on #1016 that firefox is flashing on dark theme.

This happens because the whole js is needed to set the user theme and Firefox likes to play as faster guy so the css renders first with default styles.

My idea was to quickly read the user theme and apply, setting the dark class ready there for the styles.

We know that is dangerous to use `JSON.parse` without try catch because it throws exception and stop execution, but I also notice theres not many places doing try/catch for the localStorage values as well.

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -25,6 +25,10 @@
     <script>
       window.APP = {};
       window.APP.baseUrl = "{{.BaseUrl}}";
+
+      const browserPrefers = !(window.matchMedia !== undefined && window.matchMedia("(prefers-color-scheme: light)").matches);
+      const { darkTheme = browserPrefers } = JSON.parse(localStorage.getItem("settings")) || {};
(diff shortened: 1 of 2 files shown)


s0up4200 (approved):

web/index.html
```diff
@@ -25,6 +25,10 @@
     <script>
       window.APP = {};
       window.APP.baseUrl = "{{.BaseUrl}}";
+
+      const browserPrefers = !(window.matchMedia !== undefined && window.matchMedia("(prefers-color-scheme: light)").matches);
+      const { darkTheme = browserPrefers } = JSON.parse(localStorage.getItem("settings")) || {};
```
zze0s, line 30: What happens on initial load with this if there's nothing in localStorage from before?

PR #1022, opened 2023-07-22
fix(releases): bad postgres fk migration

zze0s:
Different naming of foreign key constraints led to failing migrations.

This adds `IF EXISTS` check and the other naming variant which should fix the issue.

Reporters @jpalenz77 and @BaukeZwart


BaukeZwart:
PR fixed the issue.
Thanks for the quick fix.

PR #1028, opened 2023-08-01
ci: enable dependabot

KyleSanderson:


zze0s (approved):
Let's see how this goes :rocket:

PR #1034, opened 2023-08-04
build(docker): improve cache on web-builder stage

fabricionaweb:
The idea is that we improve the Docker cache layers on the web-builder.

If theres no changes on package.json or pnpm-lock, we dont need to run the install command.

This is the similar that we already have implemented on the app-builder stage.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,14 @@
 # build web
 FROM node:18.17.0-alpine3.18 AS web-builder
-COPY . ./
+RUN corepack enable
+
 WORKDIR /web
-RUN corepack enable && \
-    pnpm install --frozen-lockfile && \
-    pnpm run build
+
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -13,7 +17,7 @@ ARG VERSION=dev
 ARG REVISION=dev
 ARG BUILDTIME
 
-RUN apk add --no-cache git make build-base tzdata
+RUN apk add --no-cache git build-base tzdata
```
fabricionaweb, line 20: `make` is already inside `build-base`

zze0s (approved):
Sorry about the delay!

We already talked about this on Discord, but this Dockerfile is only used for local builds. The one we ship with is `ci.Dockerfile`.

This is however a good change so might as well pull it in 😄

PR #1035, opened 2023-08-04
fix(indexers): IRC URL for Sharewood

Mavyre:
Sharewood URL has been changed recently. This updates the IRC URL for newly created connections

### DIFF
--- a/internal/indexer/definitions/sharewood.yaml
+++ b/internal/indexer/definitions/sharewood.yaml
@@ -20,7 +20,7 @@ settings:
 
 irc:
   network: HiddenForest
-  server: irc.hiddenforest.ml
+  server: hiddenforest.joe.dj
   port: 7000
   tls: true
   channels:
--- a/internal/indexer/definitions/sharewood.yaml
+++ b/internal/indexer/definitions/sharewood.yaml
@@ -26,7 +26,7 @@ irc:
   channels:
     - "#rss"
   announcers:
-    - Lana
+    - lana
   settings:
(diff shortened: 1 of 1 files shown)


s0up4200 (approved):

Mavyre:
I have changed the announcer (from Lana to lana), because the case sensitivity changed on the new server. That modification works on my end and the announces are now detected again.

zze0s:
> I have changed the announcer (from Lana to lana), because the case sensitivity changed on the new server. That modification works on my end and the announces are now detected again.

The announcers are not case-sensitive. When autobrr checks if it's a valid announcer that comparison is made as lowercase to get around issues if the casing changes.

Do you have some kind of official news/announcement from the site so we're not just blindly trusting to change this?

Mavyre:
> The announcers are not case-sensitive. When autobrr checks if it's a valid announcer that comparison is made as lowercase to get around issues if the casing changes.

Oh strange. I don't understand why the announcements were not catched by autobrr. They were actually displayed in the autobrr channel window, but the Last announce column stayed "N/A" all the time. Now it works, but with your insight I'm afraid that it comes from something else.

> Do you have some kind of official news/announcement from the site so we're not just blindly trusting to change this?

They announced the URL change on top of the shoutbox, but not the announcer name change. I did a whois to verify the case sensitivity (username instead of real name).

If this isn't coming from the announcer name, why would the announcements to not be detected anymore by autobrr? I've searched the logs on my end and couldn't find anything else...

PR #1038, opened 2023-08-05
feat(config): check for updates reading from config

fabricionaweb:
I'd like the endpoint `api/updates/latest` to be able to parse the `config.CheckForUpdates` at runtime and return the `204: no content` with no need to restart the app for it.

## How to reproduce:

- Running a previous version like 1.27.1
- Go to `Settings > Application` and toggle the `Check for updates`
  - if the current state is disable and you change to enable: 
    - after enable: it will right away display a notification (check always happens on start server)
    - after disable it again: the notification dont disappear - it only goes away after restart 
  - if the current state is enable and you change to disable:
    - the notification dont disappear - it only goes away after restart


internal/http/update.go
```diff
@@ -36,11 +39,13 @@ func (h updateHandler) Routes(r chi.Router) {
 }
 
 func (h updateHandler) getLatest(w http.ResponseWriter, r *http.Request) {
-	latest := h.service.GetLatestRelease(r.Context())
-	if latest != nil {
-		render.Status(r, http.StatusOK)
-		render.JSON(w, r, latest)
-		return
+	if h.config.CheckForUpdates {
```
KyleSanderson, line 42: This should be inverted, and return NoContent or similar to prevent errors going forward (and burrito code).

internal/http/update.go
```diff
@@ -36,11 +39,13 @@ func (h updateHandler) Routes(r chi.Router) {
...
+		latest := h.service.GetLatestRelease(r.Context())
+		if latest != nil {
+			render.Status(r, http.StatusOK)
+			render.JSON(w, r, latest)
+			return
+		}
 	}
 
 	render.NoContent(w, r)
```
KyleSanderson, line 51: This likely should return something else, so there's an intelligible error presented to the client if the update fails to run. Maybe an icon under the rabbit with a broken cable, or something like that with a mouse-over message.

zze0s (commented):
I'm not sure this should be fully disabled if the check for updates is disabled. By doing so we can't even have a button to manually fetch for updates. We don't have that currently but it could be a good improvement.

I think the UI should decide if it should check for new updates or not since that is where this endpoint currently is used.

zze0s (approved):

PR #1043, opened 2023-08-07
fix(vite): update code for updated dependencies

KyleSanderson:


web/package.json
```diff
@@ -29,54 +29,54 @@
...
-    "@headlessui/react": "^1.6.4",
-    "@heroicons/react": "^2.0.11",
-    "@hookform/error-message": "^2.0.0",
-    "@tailwindcss/forms": "^0.5.2",
-    "@tanstack/react-query": "^4.29.3",
-    "@tanstack/react-query-devtools": "^4.29.3",
-    "@types/node": "^18.0.0",
-    "@types/react": "^18.0.12",
-    "@types/react-dom": "^18.0.5",
+    "@fontsource/inter": "^5.0.8",
```
stacksmash76, line 32: `pnpm remove @fontsource/inter` then `pnpm i @fontsource-variable/inter`

web/src/index.tsx
```diff
@@ -6,7 +6,7 @@
 import { StrictMode } from "react";
 import { createRoot } from "react-dom/client";
 
-import "@fontsource/inter/variable.css";
+import "@fontsource/inter";
```
stacksmash76, line 9: `import "@fontsource-variable/inter";`
then in index.css change so the first font-family line starts like this (keep both variants):
`font-family: "InterVariable", "Inter Variable",`

web/src/screens/filters/Action.tsx
```diff
@@ -76,7 +76,7 @@ export function FilterActions({ filter, values }: FilterActionsProps) {
   return (
     <div className="mt-10">
       <FieldArray name="actions">
-        {({ remove, push }) => (
+        {({ remove, push } : {remove:any, push:any}) => (
```
stacksmash76, line 79: `import type { FieldArrayRenderProps } from "formik";`
then
`({ remove, push }: FieldArrayRenderProps) => (`

web/src/forms/settings/IrcForms.tsx
```diff
@@ -26,7 +26,7 @@ interface ChannelsFieldArrayProps {
 const ChannelsFieldArray = ({ channels }: ChannelsFieldArrayProps) => (
   <div className="p-6">
     <FieldArray name="channels">
-      {({ remove, push }) => (
+      {({ remove, push } : {remove:any, push:any}) => (
```
stacksmash76, line 29: See comment at the bottom.

stacksmash76 (changes requested):
Next time use at least #0 buck or go bust. Last time my hog Smarty ran out of the "play" dungeon, I got my Remi and shot him right between the eyes. God damn.

Anyhow, thanks for playing sport. See you later.

stacksmash76:
@zze0s time to close this https://github.com/web-ridge/react-ridge-state/issues/26

KyleSanderson:
@stacksmash76 It's very likely not new, but there's a font flicker on each load. Do you know how this can be resolved?

stacksmash76:
> @stacksmash76 It's very likely not new, but there's a font flicker on each load. Do you know how this can be resolved?

Yeah, correct. Problem is, the browser figures out too late when the font needs to be fetched + suffers from resource ordering during loading. So, in order to find the font and load it (as it is at the moment):
- load HTML
- parse HTML
- in order of occurrence load auxiliary files (icons/stylesheets/fonts)
- parse auxiliary files
- load nested auxiliary files from parsed sources

I did play with this in Webpack some years ago, it's definitely possible to optimize this so it's much less noticeable on slower connections. On moderately fast connections for the most part this should be barely noticeable. Especially with huge browser caches. 

However, I've got no clue on how to do it with Vite, maybe @fabricionaweb has an idea (or is willing to research it, in contrast to me). Basically, we'd want to have a custom <style> code block in the HTML (before favicons and other stuff) which would specify the font-family sources along with the entire body {} block inlined. This should (hopefully) make the browser prefetch the font sooner than CSS gets parsed or rendered.

Maybe the easiest way is to just bundle the fonts ourselves (i.e. get rid of the fontsource dependency) and not rely on the bundler to do the work for us. I'll do some tinkering when I get the chance.

stacksmash76 (approved):

zze0s (approved):

PR #1052, opened 2023-08-17
fix(filters): replace newline with comma

### LINKED ISSUE #1019: Newlines in input fields causes matching logic to fail
**Version**
1.27.1

**Describe the bug**
Input fields allow newlines, the newlines are added to the matching logic, which results in nothing matching.

**To Reproduce**
```js
{
    "name": "sadsad",
    "version": "1.0",
    "data": {
        "shows": "*Show1*,\n*Show2*"
    }
}
```

```
rejections: (shows not matching. got: Show2 want: *Show1*,\n*Show2*)"}`
```

**Expected behavior**
Newlines should be stripped or replaced by commas I guess?

zze0s:
Replace newline of `shows` with comma.

Fixes #1019

### DIFF
--- a/internal/filter/service.go
+++ b/internal/filter/service.go
@@ -164,6 +164,10 @@ func (s *service) Update(ctx context.Context, filter *domain.Filter) error {
 		return errors.New("validation: name can't be empty")
 	}
 
+	// replace newline with comma
+	filter.Shows = strings.ReplaceAll(filter.Shows, "\n", ",")
+	filter.Shows = strings.ReplaceAll(filter.Shows, ",,", ",")
+
 	// update
 	if err := s.repo.Update(ctx, filter); err != nil {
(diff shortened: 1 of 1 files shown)


internal/filter/service.go
```diff
@@ -164,6 +164,10 @@ func (s *service) Update(ctx context.Context, filter *domain.Filter) error {
 		return errors.New("validation: name can't be empty")
 	}
 
+	// replace newline with comma
+	filter.Shows = strings.ReplaceAll(filter.Shows, "\n", ",")
+	filter.Shows = strings.ReplaceAll(filter.Shows, ",,", ",")
```
KyleSanderson, line 169: https://github.com/KyleSanderson/upgraderr/blob/6db1d724e7672d4da3c28b36287d64b972dddb8b/main.go#L1226

PR #1058, opened 2023-08-24
feat(web): Flip to Bun from pnpm

KyleSanderson:


zze0s:
Good experiment but I don't think we should make the move at this time 🐢

PR #1062, opened 2023-08-28
fix(indexers): update iPlay IRC port

sigma2017:
Port changed to 6668 due to botnet attack


zze0s:
Hi! Thanks for the PR. Is this a permanent change or temporary one? If it's temporary I don't think we should change. If it however is permanent then sure!

sigma2017:
Yes, it is permanent.
Thanks in advance!

zze0s (approved):

PR #1063, opened 2023-08-28
feat(indexers): support improved ANT announce

onedr0p:
Confirmed this is the new IRC format announced coming from ant.


zze0s (approved):
Thanks for your contribution! 🚀

PR #1065, opened 2023-08-29
fix(web): implement offline detection

KyleSanderson:
This belongs in the service worker obviously, but I'm not a web head - so my poor mans attempt at fixing the most critical bug we've had since the PWA shipped.


web/src/components/alerts/ErrorPage.tsx
```diff
@@ -17,11 +17,19 @@ export const ErrorPage = ({ error, resetErrorBoundary }: FallbackProps) => {
     }
   });
 
-  return (
+  var online = true;
+  const backendEndpoint = window.location.host + "/api/healthz/liveness"
+  fetch(backendEndpoint, { method: 'HEAD' })
+  .catch(() => {
+    online = false
+  });
```
zze0s, line 25: Would recommend to add this in the APIClient sicne that handles paths with baseUrl.

zze0s:
Is this fully compatible with the changes in https://github.com/autobrr/autobrr/pull/1089? And vice versa

zze0s (approved):

PR #1066, opened 2023-08-29
enhancement(web): IRC logs view

martylukyy:
This PR now implements auto scrolling for IRC logs.
Sorting is now newest at the bottom again and you start out scrolled all the way down when the setting is activated.
Additionally I added the timestamp to the logs because in one case I wanted to see the timestamps at a glance instead of moving the mouse over the different entries.
I am not too worried about it there being to much information in for a single entry since we have the fullscreen view too.
Also, I added some margin to the show/hide button because it looked crowded.

I noticed when `settings.scrollOnNewLog` gets changed it loads the entire map again..
To circumvent duplicate entries it will now clear the div once, resulting in resetting the position to the top of the div or respectively the bottom of the div depending on the value of `settings.scrollOnNewLog` .
If someone has a better way of handling this please ping me on discord or comment here.

### DIFF
--- a/web/src/components/Checkbox.tsx
+++ b/web/src/components/Checkbox.tsx
@@ -15,7 +15,7 @@ interface CheckboxProps {
 export const Checkbox = ({ label, description, value, setValue }: CheckboxProps) => (
   <Switch.Group as="li" className="py-4 flex items-center justify-between">
     <div className="flex flex-col">
-      <Switch.Label as="p" className="text-sm font-medium text-gray-900 dark:text-white" passive>
+      <Switch.Label as="p" className="text-sm font-medium whitespace-nowrap text-gray-900 dark:text-white" passive>
         {label}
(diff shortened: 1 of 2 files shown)


onedr0p:
I agree, it should be sorted newest to oldest!

einsern:
Most intuitive would be to copy the way every chat app on the planet does it.
Newest at the bottom and you start out scrolled all the way down.

onedr0p:
That works too 👍🏼

KyleSanderson:
> Most intuitive would be to copy the way every chat app on the planet does it. Newest at the bottom and you start out scrolled all the way down.

Indeed. No need to reinvent anything here. With logs there's no stable position if I recall correctly, so there's some work to be done to "pin" you at a message if it starts to receive additional input (new messages).

s0up4200 (approved):

PR #1073, opened 2023-09-02
feat(feeds): add scheduled cleanup

### LINKED ISSUE #1055: *znab doesn't use a rendered key in feed_cache
**Describe the bug**
Very minor because it's not being used by anything critical, but right now we use the torznab/newznab job name for adding entries into the cache which does not align to the rendered identifier that's used by torrents. I haven't looked too closely at the fix, but IndexerIdentifier may already just do this for us.

https://github.com/autobrr/autobrr/blob/ea35ef7fe8b8dec2158a952aacc9659dd7815bc6/internal/feed/newznab.go#L153
https://github.com/autobrr/autobrr/blob/ea35ef7fe8b8dec2158a952aacc9659dd7815bc6/internal/feed/torznab.go#L217

zze0s:
Add daily scheduled cleanup job for feed cache. Cleans cache older than 30 days.

This PR also changes:
* The feed cache from a generic "bucket" style KV store to using foreign keys connected to `feed.id`
* Refactor of feed jobs and their identifiers for better control

Fixes #1055

### DIFF
--- a/internal/database/feed_cache.go
+++ b/internal/database/feed_cache.go
@@ -212,7 +212,7 @@ func (r *FeedCacheRepo) DeleteByFeed(ctx context.Context, feedId int) error {
 		return errors.Wrap(err, "error exec result")
 	}
 
-	r.log.Debug().Msgf("deleted %d rows from feed cache: %s", rows, feedId)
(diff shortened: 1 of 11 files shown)


KyleSanderson (approved):
much closer to where we want to be

PR #1081, opened 2023-09-03
feat(downloadclient): set downloadClientId for arr clients

### LINKED ISSUE #961: [FEATURE REQ] Sonarr/Radarr: add ability to specify downloadClientId
**Version**
v1.26.2

**Describe the bug**
I'm not sure this is a bug/feature or how the decisions are made. In fact, I'm still trying to figure out if autobrr causes this.
In Sonarr/Radarr, I have 2 clients set up. One for public/semi-private trackers that I don't seed very long on and one for private trackers with a minimum seed time of 90 days. **In Autobrr, I only have Sonarr/Radarr set up as download clients, nothing else.**

In both servarr applications, each indexer is assigned to a specific client. This is an example - qBittorrent is the wrong client. For TL, qBittorrent-semi-private is set:
image
image

What I've noticed is that when autobrr picks up a release and sends it to Sonarr/Radarr, it ends up in the wrong client. It's almost as if the client settings are ignored. Thankfully, I've caught it in time before it got removed from the client and potentially caused HnR.

So I guess my question is, is there any way to tell autobrr which client the servarr application is supposed to use? 

Log from Radarr:
```
2023-05-25 09:35:05.3|Info|ReleasePushController|Release pushed: Torrent Name 1999 1080p BluRay DD+5 1 x264-playHD - https://www.torrentleech.org/rss/download/<id>>/(removed)/Torrent+Name+1999+1080p+BluRay+DD+5+1+x264-playHD.torrent
2023-05-25 09:35:05.3|Info|DownloadDecisionMaker|Processing 1 releases
2023-05-25 09:35:05.7|Info|DownloadService|Report for Torrent Name (1999) sent to qBittorrent from indexer torrentleech. Torrent Name 1999 1080p BluRay DD+5 1 x264-playHD
```

zze0s:
Add ability to set `downloadClientId` to pass to the arr client for it to use a specific client.

Fixes #961


internal/action/lidarr.go
```diff
@@ -43,19 +43,28 @@ func (s *service) lidarr(ctx context.Context, action *domain.Action, release dom
 		cfg.Password = client.Settings.Basic.Password
 	}
 
-	arr := lidarr.New(cfg)
+	externalId := 0
+	if client.Settings.ExternalDownloadClientId > 0 {
```
KyleSanderson, line 47: This is backwards. This is saying the parent takes priority over the wellbeing of the child 😄. The outcome is if this is set at the global level, you'll need another id set to override at the time of execution with one that isn't set. If it's inverted, it can still be overridden in the action.

PR #1083, opened 2023-09-05
feat(downloadclients): arrs increase timeout for release/push

KyleSanderson:
For reasons unbeknownst to me, pushes can take lots of time before autobrr gets a return back. The pushes eventually succeed, but not before autobrr logs an error on the action as the connection has since "timed out". As these actions are run in a goroutine, we're not at risk of blocking main application activity, so just crank it.


zze0s:
Is this only a Sonarr issue or should all arrs be updated with this increased timeout?

KyleSanderson:
> Is this only a Sonarr issue or should all arrs be updated with this increased timeout?

I would suspect it's all, but I only use the former.

zze0s (approved):

PR #1087, opened 2023-09-08
enhancement(web): improve font loading performance and avoid page reflow

stacksmash76:
enhancement: improve font loading performance and avoid page reflow
chore: get rid of postcss-import (we're not using it and it's not needed for tailwind)


web/public/Inter.var.woff2
KyleSanderson, line 1: Shouldn't this come from the package?

stacksmash76: I got rid of the package. We just need the one .woff2 file anyways. Seems to play a bit nicer with Vite/Rollup bundling this way.

zze0s (approved):

PR #1093, opened 2023-09-09
enhancement(web): modernize APIClient and improve robustness

stacksmash76:
enhancement: APIClient now follows the more recent RFC3986 (this wasn't the case previously with URLSearchParameters)
enhancement: improved APIClient DX by adding a queryString parameter (avoiding URLSearchParameters)
fix: changed Notification type to ServiceNotification (collision with built-in browser API https://developer.mozilla.org/en-US/docs/Web/API/Notification -- so TS checks wouldn't function as necessary)

needs a bit more testing from my side, but should be pretty solid


web/src/api/APIClient.ts
```diff
@@ -163,51 +274,61 @@ export const APIClient = {
...
+    delete: (olderThan: number) => appClient.Delete("api/release", {
+      queryString: { olderThan }
+    }),
+    replayAction: (releaseId: number, actionId: number) => appClient.Post(
+      `api/release/${releaseId}/actions/${actionId}/retry`
+    )
   },
   updates: {
     check: () => appClient.Get("api/updates/check"),
-    getLatestRelease: () => appClient.Get<GithubRelease | undefined>("api/updates/latest")
```
stacksmash76, line 211: Someone should confirm here whether this `| undefined` part really was unintended or whether it was just legacy junk.

zze0s: Probably unintended or needed during dev, but the endpoint should always get the latest version.

web/src/api/APIClient.ts
```diff
@@ -7,132 +7,228 @@ import { baseUrl, sseBaseUrl } from "@utils";
...
+        if (v.length) {
+          params.push(`${serializedKey}=${encodeRFC3986URIComponent(v)}`);
+        }
       }
+    }
+
+    if (params.length) {
+      endpoint += `?${params.join("&")}`;
+    }
+  }
```
zze0s, line 80: What's the reason we want this over the built in `URLSearchParams`?

web/src/api/APIClient.ts
fabricionaweb: The endpoint will return nothing (no-content) if running the latest. I belive it is considered undefined

zze0s (approved):

PR #1105, opened 2023-09-10
fix(web): cant update from settings panels

stacksmash76:
Fix propagation bug with IRC network table dropdown.


martylukyy (approved):
You know we appreciate you... do you? 😂

KyleSanderson:
the IRC save button doesn't work in v1.29-rc1 (which is the latest tag), and is consistent with this PR unfortunately.

PR #1118, opened 2023-09-12
feat(database): setup integration tests

KaiserBh:
✅ Add Database Integration Tests

This PR introduces comprehensive database integration tests into our codebase. Although it took longer than anticipated due to other commitments, it is now complete and ready for review. I apologize for the delay and appreciate your patience. The tests have been designed to be compatible with both SQLite and PostgreSQL databases.

Please review extensively (I don't think there is any breaking changes but I did modify some of the database functions and found some issue with it as well while writing the tests) and let me know if there is any changes that I need to make.


internal/database/database_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/stretchr/testify/assert"
+	"os"
+	"testing"
+)
+
+func getDbs() []string {
+	return []string{"sqlite", "postgres"}
+}
+
+func SetupDatabase(t *testing.T, dbType string) *DB {
```
zze0s, line 15: This does not need to be exported, and could maybe be renamed to `setupDatabseForTest` to be more descriptive?

internal/database/database_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	// Open the database connection
+	if err := db.Open(); err != nil {
+		t.Fatalf("Could not open db connection: %v", err)
+	}
+
+	return db
+}
+
+func SetupLogger() logger.Logger {
```
zze0s, line 49: Same as above, does not need to be exported and rename to `setupLoggerForTest` makes it a bit more descriptive.

zze0s (commented):
This is great, kaiser!!

Lots of cases covered, well done!

Will do a couple of passes of review but starting small :smile:

internal/database/database_test.go
KaiserBh: :+1:

internal/database/database_test.go
KaiserBh: :+1:

zze0s (approved):
Awesome work @KaiserBh!

I did some changes to the setup to use `TestMain` that bootstraps the dbs and migrates data and cleans the PG schema before each run.

It only inits once per test run so it got a little bit faster as well.

Some future improvements could be to look at https://github.com/ory/dockertest and https://github.com/testcontainers to make the setup even better, but the current solution works well enough.

PR #1121, opened 2023-09-15
fix(web): manually specify build targets instead of browserlist

stacksmash76:
get rid of browserlist for now, manually specify minimal targets via vite to esbuild


fabricionaweb (approved):

stacksmash76:
Well, this fixes nothing. I guess we don't need this PR.

PR #1123, opened 2023-09-15
fix(web): font preloading

martylukyy:
This PR fixes the following console error for font preloading on certain systems.
```
A preload for 'https://example.com/autobrr/Inter.var.woff2' is found, but is not used because the request credentials mode does not match. Consider taking a look at crossorigin attribute.
```

I tested this fix on my server to make sure it is working with subfolder installs since there were concerns about that.
Network tab in DevTools shows it successfully preloads the fonts without any error about this in the console.

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -11,7 +11,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <meta name="theme-color" content="#141415" />
     <meta name="description" content="autobrr" />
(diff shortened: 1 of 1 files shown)


stacksmash76 (approved):

martylukyy:
After running this PR for a while there came up other problems than the console error which are more severe.
Going to close this.

PR #1145, opened 2023-09-22
feat(web): show more app info

zze0s:
Add more app info.

* database type used
* application binary path
* config path

image


martylukyy (approved):
LGTM.
Could truncate the field for Application and Config path for absurdly long paths,
in case it get's reported otherwise send it.

PR #1148, opened 2023-09-24
feat(notifications): Telegram display file size

we11adam:
This adds file size information to notifications sent via Telegram.

See also: https://github.com/autobrr/autobrr/pull/1143


zze0s (approved):
Thanks for your contribution! :clap:

PR #1154, opened 2023-09-27
fix(web): tooltips

### LINKED ISSUE #1150: enhancement: prevent browser from sending referrer data
All external links will have to look like:
```
<a rel="noopener noreferrer" target="_blank" href="https://autobrr.com">...</a>
```
(just setting target="_blank" is not enough)

Else the endpoint will be able to track where the user is coming from (e.g. private server domain), which is bad privacy-wise.

stacksmash76:
tooltips have been broken since 1.30, this fixes them and also closes #1150


s0up4200 (approved):

s0up4200 (commented):
Found something on 2nd review.

- We need to adjust the fadeOut time to make it easier to click the URLs in the tooltips.
- The headers are messed up on mobile, and clicking the headers and info icons focuses the fields they belong to - which pops the keyboard.

PR #1155, opened 2023-09-27
enhancement(web): ui overhaul

stacksmash76:
- made filter ui more mobile-friendly
- improve settings page active (selected) state styles
- add transition effects to filter page tabs
- normalize RowItemSelect text color on light theme (backport of commit d3ca3e5)
- fixed switch/toggle group not being aligned on mobile
- added border to 'pill'-shaped parent containers (a bit easier on the eyes, imo)
- made FormButtonsGroup mobile-friendly
- added hover effect and underline highlight for collapsible section on filters page (so sections pop out a bit more)
- fixed collapsible section bottom border
- moved react-multiselect-component styles to use tailwind css vars
- added border, shadow for rmsc dropdown
- added icons for user navbar dropdown
- added line dividers and border for user navbar dropdown
- made all inputs more contrastful (especially on filters page)
- made simple dropdown select (used for tags) in line with other changes
- added icons for external/actions add new button
- revamped filter details page, made it mobile-friendly
- filter details page has less gaps
- right user navbar dropdown now has proper transitions
- fixed most input components to be of a similar style (there's a shitload of them, so i've missed some most likely)
- added a required star next to filter name
- made releases search engine hint wrap vertically on mobile
- fixed a regex equivalence for search engine hint
- added borders around activity/release tables
- made activity/release table headers a bit darker
- added hover effect over activity/release table headers
- made activity/release table header text a bit stronger/contrastful
- (similar changes to filter list as for activity/release table)
- replaced filter action alert with a better-styled component
- corrected padding in form for actions
- made filter external/actions tabs more mobile friendly
- fixed "remove filter action" delete modal not using transitions
- made the > delimiter arrow in filter details page more pronounced
- added color transitions to some components
- tornzab pill is now a bit more pronounced on dark theme
- no indexers pill is overall more pronounced
- fix search tips child container having round top corners
- fix settings cell item upper/lower corner parts being cut off by 'truncate' (fix for regression introduced in commit b011abe6270a107fd04b921d2d794dcaa5c82ef3)
- every checkbox (at least for filter details) is now a switch
- pagination buttons are now better styled
- made PUSH_ERROR status use an X icon (instead of !) and picked a better color scheme. also picked better colors for PUSH_APPROVED.
- made (table/box) borders less prominent (introduced by previous commits in this branch)
- made filter title/subtitle components more prominent
- removed unneeded CheckboxField, Switch, SwitchFormik components
- corrected input fields to use a more pleasant color scheme (maybe there are more to be done?)
- fixed Select component uncontrolled->controlled warnings from React
- made toast component use a bigger shadow and added a more appropriate border
- made SlideOver and other forms not show scrollbar always, only when needed
- react-select components now fit in with the new color scheme
- removed "download log files" from **settings** page and delegated it to the logs page
- standardized settings page look (previously it was all over the place)
- removed subtitle from filter details collapsible sections when they were redundant
- CollapsableSection -> CollapsibleSection
- made filter details page more mobile friendly
- perfect flac/cue/log settings use a better layout
- fixed various table headers and their row items
- various other bug and style fixes, there are too many of them to list them
- got rid of unused Action Settings page
- ensure no eslint warnings
- checkbox: made gray bg slightly brighter on light theme
- react-select: organized all custom components into a single file
- fix indexer tag pill ordering 
- fix select dropdown color to match input boxes
- change SwitchGroupWideRed with SwitchGroupWide
- changed top navbar bg gradient to the closest lerped color
- fixed style ordering issue where errored border would not be applied properly
- use smaller horizontal margin on mobile for page content (px-4 -> px-2)
- revamped login page to be responsive
- made filter actions/externals responsive (no more overflow)
- compacted filter details page
- compacted the feed/indexer/irc/notifications settings page tables to be a bit friendlier for phone screens (could be better)
- set better dark red color for 'no indexer' pill (@martylukyy)
- change APISettings header column spans (@martylukyy)
- change indexers list item to use py-1.5 instead of mt-1 (@martylukyy)
- fixed padding on some table columns (feed/indexer/irc/notifications) to be a bit friendlier for phone screens (not perfect)
- made releases (settings) page design responsive
- changed ol (ordered list) -> ul (unordered list) where needed
- add text-shadow tailwind class
- add text-shadow to CollapsibleSection title prop on dark theme
- add custom box-shadow for tables (shadow-table) - mostly for the light theme
- lower background swirl pattern opacity (10% -> 7%)
- fix RetryActionButton on light theme
- use lighter bottom border color for header/navbar on dark theme
- use a bit stronger border on light theme
- fix \<p\> inside \<p\> warning for Switch component
- improve filter actions view
- improve filter external view
- begin work on dividing filter details page into standardized components
- add missing border color for FilterItemDropdown, FeedItemDropdown and ListItemDropdown (irc) for the light theme (it was black by default)
- use lighter alternating table row colors for FilterListItem
- highlight priority in FilterListItem when not 0
- re-adjusted gaps on filter details page
- fixed docs, corrected tooltip hints, disabled log score when its inactive and some other minor fixes
- word capitalization is mandatory for title in TitleSubtitle component
- picked same border color for SelectControl as for other inputs
- styled IndicatorSeparator and DropdownIndicator for the Select control
- made Toast component format whitespace such as \n, \t
- added formik error object formatting to toast (for relevant messages)
- corrected new gap classes
- changed look of Actions/External filters list
- made Log Level select component in settings page use 1 col-span instead of 3 (on desktop only)
- made input label text a bit darker on the light theme


KyleSanderson (commented):
CodeQL has my feedback, but nice job breaking these mostly out into disparate commits. Very reviewable going commit by commit.

martylukyy (commented):
I made some layout changes to this PR and i added the ability to add classnames to some components.
This was needed to set the appropriate boundaries to other elements since this could not be generalized.
Please check if the way the classname helpers were set is an acceptable way or if you would set them differently.

Furthermore I found one bug:
On the `/settings/indexers` page the description will not line break anymore.
Please fix.

After that i would propose one more change:
Set a line break in the description for the perfect FLAC filter option,
since the description gets pretty close to the switch which looks awkward.
I would propose to separate them evenly between the 2 lines for best results.

stacksmash76:
> I made some layout changes to this PR and i added the ability to add classnames to some components.

I didn't make the layout changes without a good reason. I wish you brought this up earlier (during one of our say 5 previous conversations).

The reason I made the layout more compact is because elements which are near each other are perceived as related (principle of proximity). This means that equidistant spacing between different categorical items (sections vs. regular items) is also usually a bad thing (especially when there's a lot of going on - like in our case).

The second problem here is for people with glasses. Often times they suffer from tunnel vision (like myself). Nearly halving the distance from 24px to 16px between items makes a big difference for me and helps with scanning the page. Ditto for borders and contrast on headers, huge helpers while navigating. I doubt these changes will make anyone bleed their eyes out.

Honestly, I think I have the better argument here, and ultimately I won't accept these changes (i.e. how it was). If you've got strong feelings about the spacing we can resolve this the democratic way by doing an `@all` vote with before vs. after screenshots. Probably the fairest way. I'll do the regression push myself if it comes to that.

Apart from that, you broke the layout for mobile phones:
broken

(You also re-introduced eslint warnings which I tried to get rid of)

> Furthermore I found one bug: On the `/settings/indexers` page the description will not line break anymore. Please fix.

Just a minor typo. Fixed.

> After that i would propose one more change: Set a line break in the description for the perfect FLAC filter option, since the description gets pretty close to the switch which looks awkward. I would propose to separate them evenly between the 2 lines for best results.

Fixed.

martylukyy:
The reason I didn't bring some things up earlier, is straight up because i had not discovered them back then. ^^

While I agree on some spacing changes i disagree on others.
One of those i disagree with is the changes of the freeleech section.
The element of the switch is too close to the field right of it and the elements are not vertically centered.
In my eyes there is no need for them to be so close since the separator indicates them as being related.
Because of that I think that `gap-6` is the better choice in comparison to `gap-3`
and centering it vertically makes it looks more refined.
I feel similar about the log/cue/log score section aswell.

Another change that got reverted that i disagree on is the font size of the IRC settings page.
I don't know why you want to remove the `text-sm` property here since this property spans itself
throughout all the settings pages **except** the IRC page now.
Feel free to elaborate on this or adding it again in case this is an oversight.

I also feel like that the Enable/Disable switch inside the filters are too close to the Save button
and not vertically centered to the above and beneath element.
While this seems rather minuscule i can see myself fatfingering the wrong element here.
It is not related to any other element so we might aswell add some spacing here.

Now to borders and headers.
If i recall correctly i just removed the outer borders for the dashboard, filters and releases table
since i don't think it fits our rather minimalistic design language.
It also doesn't add any value in separating 2 elements from each other i feel like since there is nothing to seperate.
For the headers i actually have a question here because i do have very good eyes still..
With the color being 825 instead of 850:
Do you have problems distinguishing the difference between head and body with it being 825 for example?
I have the same doubts about this change as the one mentioned before - 
I don't think it fits our rather minimalistic design language.

The thing about i am missing an opinion from you is the filters list page.
I dislike they way it looks now. The color of every second list entry should not be the color of the header 
and the absence of distance between the entries makes it look cluttered to me.
Personally, i feel like the color difference between the alternating list entries should be more subtle.
For me it is just not fitting into the design of the other pages where we do have more distance to neighboring elements.
A like a mix of 775,800 and 825 for the color choice better and i would reintroduce the color change on hover again.

While this is a lot of feedback please do not see this as critique and if you feel like we do have disagreements here
that we cannot find a consensus on, feel free to do a `@all` vote. :)
I am not going to be mad or angry because we need a third opinion about a design decision :D

The ESLint part i find very interesting though.. my ESLint didn't throw me any squiggly lines or warnings..
Might need to check my setup here - would appreciate if you could give some information on where i fucked up!

I don't think i say this enough but i feel like now is a good time to after that:
Thank you for your continued efforts on autobrr.
Without you autobrr wouldn't be what it is today.
And also thank you for fixing all the bugs i am producing along the way 😂

stacksmash76:
> The reason I didn't bring some things up earlier, is straight up because i had not discovered them back then. ^^

Not quite true. I know you said you wanted to change the borders (you didn't say exactly what), I thought maybe you'd make the lerping increment smaller and use a more nuanced color, but I definitely didn't think you'd revert some of the changes. (After I told you twice to forget commiting to my branch, hence didn't expect your changes at all)

So imagine my surprise when you revert my changes without discussing it with me.

> One of those i disagree with is the changes of the freeleech section. The element of the switch is too close to the field right of it and the elements are not vertically centered.

The elements were never vertically centered. But I've made the elements stack vertically instead. The gaps are getting smaller, I explained why.

> Another change that got reverted that i disagree on is the font size of the IRC settings page. I don't know why you want to remove the `text-sm` property here since this property spans itself throughout all the settings pages **except** the IRC page now. Feel free to elaborate on this or adding it again in case this is an oversight.

It looks better to me. And all the tables in the settings page are different. Why I think a larger font does good here is that:
1) it has a hovering highlight (others do not)
2) it expands vertically (others do not)
3) has a child (sub-)table (others do not)
4) all text is of a very similar gray color, hence bigger size helps.

I don't think a bigger font is an issue in this context. Generally, all tables in the app need to be reworked and a new look. 

> I also feel like that the Enable/Disable switch inside the filters are too close to the Save button and not vertically centered to the above and beneath element. While this seems rather minuscule i can see myself fatfingering the wrong element here. It is not related to any other element so we might aswell add some spacing here.

I don't agree, but I've moved the switch components to the top where they should be in categorical groupings (e.g. enabled has hierarchical precedence).

> Now to borders and headers. If i recall correctly i just removed the outer borders for the dashboard, filters and releases table since i don't think it fits our rather minimalistic design language.

What's our "rather minimalistic design language"? Who defined it?

> It also doesn't add any value in separating 2 elements from each other i feel like since there is nothing to seperate.

See enterprise solutions which have data table components, e.g. Baseweb, Carbon, SAP Fiori, Radix, etc. and then you realize it's basically a very sensible guideline. Only major exception is Material Design (which has been criticized extensively for its decision).

Same thing for row striping, etc. The tables need to be reworked and standardized (at least at a data/semantic level), but I'm not willing to put in the effort. Maybe you're up for it?

Also see https://www.nngroup.com/articles/data-tables/ (there is many more research like this, Google will help you find it)

> For the headers i actually have a question here because i do have very good eyes still.. With the color being 825 instead of 850: Do you have problems distinguishing the difference between head and body with it being 825 for example?

Yes, the difference is definitely noticeable for me. And not everyone has a color-accurate IPS screen nor OLED. Especially if someone is using some "eye-saver" technology which modulates the screen into a yellow-ish tint. Then the darks get shifted and become not very discernible.

> The color of every second list entry should not be the color of the header

In principle, I agree. But 775 and 825 use different hues than 800 and 850 (blue-ish tint vs. sand-ish tint). Mixing between them looks unnatural, so I picked more similar pairs. The new color interpolation plugin should be manually corrected and a new color put in place.

> and the absence of distance between the entries makes it look cluttered to me.

This is because the filters list page should be redesigned and priority should have its own column and not be shoved under the name. Then there will be enough distance.

> if you feel like we do have disagreements here that we cannot find a consensus on, feel free to do a `@all` vote. :)

That's what I told _you_.

> The ESLint part i find very interesting though.. my ESLint didn't throw me any squiggly lines or warnings.. Might need to check my setup here - would appreciate if you could give some information on where i fucked up!

`pnpm run lint` will show you the warnings on your commit.

I'll push a commit correcting/converging the mentioned things later this week.

web/src/components/inputs/switch.tsx
```diff
@@ -3,96 +3,46 @@
...
+      "flex items-center justify-between"
+    )}
+  >
     {label && <div className="flex flex-col">
       <HeadlessSwitch.Label
         passive
         as={heading ? "h2" : "span"}
         className={classNames(
-          "flex float-left cursor-default mb-2 text-xs font-bold text-gray-700 dark:text-gray-200 uppercase tracking-wide",
+          "flex float-left ml-px justify-content-center cursor-default text-xs font-bold text-gray-800 dark:text-gray-100 uppercase tracking-wide",
```
s0up4200, line 45:
```suggestion
          "flex float-left ml-px cursor-default text-xs font-bold text-gray-800 dark:text-gray-100 uppercase tracking-wide",
```

s0up4200 (changes requested):

web/src/forms/settings/IrcForms.tsx
```diff
@@ -27,44 +27,49 @@ interface ChannelsFieldArrayProps {
...
             channels.map((channel: IrcChannel, index) => {
               const isDisabled = channel.name === "#ptp-announce-dev";
               return (
-                <div key={index} className="flex justify-between">
-                  <div className="flex">
-                    <Field name={`channels.${channel.id}.name`}>
-                      {({ field }: FieldProps) => (
+                <div key={index} className="flex justify-between border dark:border-gray-700 dark:bg-gray-815 p-2 rounded-md">
+                  <div className="flex gap-2">
+                    <Field name={`channels.${index}.name`}>
```
stacksmash76, line 40: Shouldn't this be channel.id?

web/src/forms/settings/IrcForms.tsx
```diff
@@ -27,44 +27,49 @@ interface ChannelsFieldArrayProps {
...
+                              : "border-gray-300 dark:border-gray-700 focus:ring-blue-500 dark:focus:ring-blue-500 focus:border-blue-500 dark:focus:border-blue-500",
+                            "block w-full shadow-sm sm:text-sm rounded-md border py-2.5",
+                            isDisabled ? "disabled dark:bg-gray-700 dark:text-gray-400 cursor-not-allowed" : "bg-gray-100 dark:bg-gray-850 dark:text-gray-100"
                           )}
                           disabled={isDisabled}
                         />
                       )}
                     </Field>
 
                     <Field name={`channels.${index}.password`}>
```
stacksmash76, line 59: This too

web/src/forms/settings/IrcForms.tsx
zze0s: FieldArrays needs the index. It was id before but it did not load the data and errored on change.

zze0s (approved):
Massive effort. Overall good changes!

I made some changes that I did not agree with and fixed some other parts where needed.

In the future I do not want to see PRs of this size. It makes reviewing, testing and merging unnecessarily cumbersome.

PR #1162, opened 2023-10-01
feat(feeds): improve file size parsing

jolinarofmalkshur:
In rss feeds, sometimes the only place the size is mentioned is in the description field. If the size has not already been determined from another source try to read it from there.


zze0s (approved):
Thanks for your contribution! I think a lot of users will be happy about this! :smile: 

I made some improvements and also added tests for the new function.

PR #1164, opened 2023-10-02
fix(indexes): toggle on and off with switch

fabricionaweb:
The idea is the make it similar to the IRC page, also the code have been taken from there.

The issue Im currently facing here is the `mutation` at Indexers.tsx:131 it seems to override everything is sent... I can not spread the indexers because it refuses with `cannot unmarshal array into Go ... something` and looking at the onSubmit method at IndexesForms.tsx:317 it does **a lot** of things...

So... Im begging some help here :D

### DIFF
--- a/internal/database/indexer.go
+++ b/internal/database/indexer.go
@@ -232,3 +232,25 @@ func (r *IndexerRepo) Delete(ctx context.Context, id int) error {
 
 	return nil
 }
+
+func (r *IndexerRepo) ToggleEnabled(ctx context.Context, indexerID int, enabled bool) error {
+	var err error
+
+	queryBuilder := r.db.squirrel.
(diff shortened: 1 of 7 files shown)


stacksmash76:
Works fine with:
```ts
updateMutation.mutate({
      ...indexer,
      enabled: newState
} as Indexer);
```

fabricionaweb:
> Works fine with:
> 
> ```ts
> updateMutation.mutate({
>       ...indexer,
>       enabled: newState
> } as Indexer);
> ```

hmm not here... this way it worked only for torznab, when I tried some IRC it failed with some `cannot unmarshal array into Go`

zze0s (approved):
Good start! I made some improvements and refactored related parts while I was at it.

PR #1172, opened 2023-10-08
feat(indexers): FileList update download link

mynameisbogdan:


internal/indexer/definitions/filelist.yaml
```diff
@@ -67,6 +67,6 @@ irc:
 
     match:
       infourl: "/details.php?id={{ .torrentId }}"
-      torrenturl: "/download.php?id={{ .torrentId }}&file={{ .torrentName }}.torrent&passkey={{ .passkey }}"
+      torrenturl: "/download.php?id={{ .torrentId }}&passkey={{ .passkey }}"
       encode:
         - torrentName
```
martylukyy, line 72: Can also remove this since URL safe encoding for `torrentName` is not needed anymore due to removal.

zze0s (approved):

PR #668, opened 2023-01-22
fix(releases): force capture titles

varoOP:
Tested it, should be good to go.

PR #1173, opened 2023-10-08
Update animebytes.yaml regex pattern to parse title

danwilldev:
Title var was always null. Now providing it with the original value of torrentName


martylukyy:
Regression ofhttps://github.com/autobrr/autobrr/pull/668

PR #1188, opened 2023-10-16
feat(logs): sanitize basic auth in urls

danwilldev:
This regex is not perfect, any @ or : characters in a password will cause it to fail. If someone could suggest a better way of handling it then please comment or commit to this PR.
Thanks


danwilldev:
Test failing. Think I compiled the regex incorrectly

KyleSanderson:
might be worthwhile while specifying REDACTED_USER/etc. I don't think we do that elsewhere but it helps add clarity than just REDACTED literred everywhere

internal/http/logs.go
```diff
@@ -139,6 +139,10 @@ var (
 		{
 			pattern: regexp.MustCompile(`(SceneHD..invite).([a-zA-Z0-9]+)(\s+#announce)`),
 			repl:    "$1 REDACTED$3",
+		},
+				{
+			pattern: regexp.MustCompile(`(https?://)(.*?):(.*?)@(.*?)`),
```
KyleSanderson, line 144: the `:(.*?)` of this is optional, and won't match if the password is absent.

zze0s:
Looks like this needs some more love before it's ready to merge.

zze0s:
@danwilldev I was playing with this to try get it passing tests, but could not figure out why it does not work as expected. Maybe marty or soup can look into it since they built the log sanitizer.

danwilldev:
@zze0s it may be because I'm using "username:password" in the test string so it's getting double sanitized? Will leave it for now and hope one of the regex pros picks it up :D

s0up4200 (approved):

PR #1195, opened 2023-10-18
feat(indexers): PTP support freeleech

zze0s:
PTP has started to announce freeleech recently so this PR adds support to parse it properly instead of having to use `Match release tags` like `*freeleech*`.

https://regex101.com/r/QapE4W/1


martylukyy (approved):
Tested with 1000 announces of which 1000 matched. LGTM! 🚀

PR #1206, opened 2023-10-22
feat(filters): external webhook retry on status codes

buroa:
Adds the ability to retry specified status codes for external webhook filters


zze0s (approved):
Thanks for the PR @buroa!

This is a great addition that opens up for more advanced use! :rocket:

PR #1209, opened 2023-10-23
feat(indexers): add TheOldSchool

zze0s:
Add indexer TheOldSchool.

Related #1


martylukyy (approved):
LGTM under the assumption that the `uploader` var will always be announce even when anon.

PR #1211, opened 2023-10-23
feat(clients): Transmission support seedtime and ratiolimit

### LINKED ISSUE #581: [FEATURE REQ] - Transmission Seed Ratio/Time per action
As other clients already have, it would be nice if we were able to set Seed Ratio and Seed Time in the action Tranmission.

Thanks

zze0s:
Add support for setting Speed limits (ul/dl), ratio limit and seed time limits.

Related https://github.com/hekmon/transmissionrpc/issues/26

Fixes #581


zze0s:
This requires more work. The transmission library got updated and made breaking changes to how it's setup with url and port. And SeedIdleLimit needs fixing on their end.

Converting to draft for now.

PR #1212, opened 2023-10-25
fix(filters): do not split commas inside valid Regexp

### LINKED ISSUE #1241: Regex filter including commas not working as expected
**Version**
1.33.0

**Describe the bug**
Advanced filters allow the usage of Regex. Regexes allow commas, e.g. `\.[Ss]\d{1,2}[Ee]\d{1,3}\.` is a valid regex. But the backend will split the user-given regex at the comma, hence resulting in a faulty regex.

**To Reproduce**
Create a advanced filter, e.g. "match release" with a regex like `\.[Ss]\d{1,2}[Ee]\d{1,3}\.`. It will not work as expected.

**Expected behavior**
The regex should work as intended. It is not necessary to allow users to add multiple regexes, separated by comma. This can naturally be achieved by separating them with `|`.

**Additional context**
A workaround would be to use e.g. `\.[Ss]\d\d?E\d\d?\d?\.`. But it is not straight-forward to the user why a regex that uses commas doesn't work as expected. I only understood it after I had a look at the backend-code.

varoOP:
This fixes regex patterns that contain commas to not split inside valid regex delimiters.

Fixes #1241

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -61,6 +61,7 @@ require (
 	github.com/bradfitz/iter v0.0.0-20191230175014-e8f45d346db8 // indirect
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
 	github.com/gdm85/go-rencode v0.1.8 // indirect
+	github.com/go-andiamo/splitter v1.2.5 // indirect
 	github.com/google/uuid v1.3.0 // indirect
 	github.com/gorilla/securecookie v1.1.2 // indirect
 	github.com/gosimple/unidecode v1.0.1 // indirect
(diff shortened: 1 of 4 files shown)


jjxxs:
see https://github.com/autobrr/autobrr/issues/1241#issuecomment-1793883640

varoOP:
@zze0s I could fix the issue mentioned by @jjxxs - simply ignore commas within curly braces as well. However, would that be a good solution?

I feel that autobrr should accept the whole regex pattern without any splitting - as long as it is a valid regex pattern (in the go flavor) it should work. The only problem would be that the update would break filters for users who are already using it this way. 

Thoughts?

zze0s:
> @zze0s I could fix the issue mentioned by @jjxxs - simply ignore commas within curly braces as well. However, would that be a good solution?
> 
> I feel that autobrr should accept the whole regex pattern without any splitting - as long as it is a valid regex pattern (in the go flavor) it should work. The only problem would be that the update would break filters for users who are already using it this way.
> 
> Thoughts?

What we have currently is for sure a bug and having commas in regex patterns are totally valid. Didn't Kyle link a library that seemed useful and would handle a lot more cases?

varoOP:
Will look into that soon!

varoOP:
@zze0s [dc7af60](https://github.com/autobrr/autobrr/pull/1212/commits/dc7af60267350e3c072609488d8f98d1ce41d59a) fixes #1241 

The splitter package is very extensible too, for now I have only added the most likely cases.

KyleSanderson (approved):
Great start, I think eventually all those splits should be on.

zze0s (approved):
Well done! Thanks for fixing this 🎅

PR #1216, opened 2023-10-26
feat(indexers): BHD freeleech announces

martylukyy:
This PR adapts the BHD regex to match the new announces which now announce their various freeleech options.
I am unsure of what to do with the freeleech var so i improvised and used `freeleech` for a start..
Do we want to use `freeleech` or `freeleechPercent` here?
There might me some changes needed at other files to overcome that problem?

(pun intended)

Regex101 Link: https://regex101.com/r/KQeS0G/1


zze0s:
We need to add some logic to the freeleech parsing for this to work. I'll see what I can do.

zze0s (approved):

Issue #1157, opened 2023-09-28
bug(config): in-place edit overwrites logPath default comment


stacksmash76:
offending line https://github.com/autobrr/autobrr/blob/develop/internal/config/config.go#L317

it doesn't check whether the incoming line is non-empty (i.e. default) and just overwrites it with `#logPath = ""`

this is a problem in a dev env when one runs with the default config.toml and makes a change via the web ui (e.g. check updates).

solution is to either add it to .gitignore or fix the bug

PR #1219, opened 2023-10-27
fix(config): prevent overwriting default logPath

s0up4200:
~~Speculative~~ Fix for #1157 

- [x] Please test @stacksmash76


zze0s (approved):
Looks like this fixes the issue! :rocket:

PR #1220, opened 2023-10-27
feat(web): disable IRC fields

martylukyy:
The fields of an IRC channel entry will now be disabled for editing when `channels[index].name === "#ptp-announce-dev"`


s0up4200 (approved):
🚀

zze0s (approved):

PR #1227, opened 2023-10-30
fix(filters): external webhooks improve defaults

KyleSanderson:
This is a mega PR for the adjustments needed to make the retry webhooks usable that shipped in 1.32.

Namely:
Placeholders are used as production values (so all requests default to run 10 times now).
Body is closed after the first request (which makes the other 9 runs fail immediately).
"Default" is Select a HTTP method, and hides the fact it does a POST request, which is incredibly user hostile.
All of these options should be behind an "advanced" section.


zze0s (approved):

buroa:
Thanks @KyleSanderson :-)

PR #1231, opened 2023-10-30
feat(createAPIKey): set optional key length

s0up4200:
Possible improvements: Add scopes. Like `filters:read` `filters:all` etc.

CleanShot 2023-10-30 at 21 55 01@2x

CleanShot 2023-10-30 at 21 54 46@2x

CleanShot 2023-10-30 at 21 57 59@2x

CleanShot 2023-10-30 at 21 58 04@2x


s0up4200:
It was a fun little Monday project, but today the idea seems totally unnecessary. Closing this.

PR #1233, opened 2023-10-31
fix(web): tooltip z-index on filters list

martylukyy:
Fix wrong z index of tooltips on filters list.

bug1

(v1.33.1? 🤣 )


martylukyy:
fixed in #1155

PR #1239, opened 2023-11-02
feat(irc): log connection errors

frrad:
Fixes a race between IRC connection retries and user-initiated manual disconnect. 

I was able to trigger this by toggling the state of a network.

The sequence of events is:

1. This connect with retry https://github.com/autobrr/autobrr/blob/8c89481d880c2bcd6d3907c16af807ec0a86a7a0/internal/irc/handler.go#L218 wrapper tries once and fails
2. Then, as the retry delay is counting down, I toggle the network off which calls (I'm guessing) this function https://github.com/autobrr/autobrr/blob/8c89481d880c2bcd6d3907c16af807ec0a86a7a0/internal/irc/handler.go#L321 including a call to `h.client.Quit()`
3. When the retry wakes back up it calls connect again https://github.com/autobrr/autobrr/blob/8c89481d880c2bcd6d3907c16af807ec0a86a7a0/internal/irc/handler.go#L222 and gets an error.

credit to @slingamn for the actual fix in https://github.com/frrad/autobrr/pull/1

This PR also adds a new logline for connection errors from the IRC library to enable easier debugging. In my case the lack of logs made it hard to isolate a DNS problem I was encountering.


s0up4200 (dismissed):

internal/irc/handler.go
```diff
@@ -220,6 +220,7 @@ func (h *Handler) Run() error {
 				h.log.Debug().Msgf("connect attempt %d", connectAttempts)
 
 				if err := h.client.Connect(); err != nil {
+					h.log.Debug().Msgf("encountered connection error: %+v", err)
```
zze0s, line 223: This should probably be an error log instead of debug.

```diff
-h.log.Debug().Msgf("encountered connection error: %+v", err)
+h.log.Error().Err(err).Msgf("client encountered connection error")
```

zze0s (changes requested):
Hi! Thanks for the PR.

I think this will trigger on disconnect as well, and that's why it wasn't set before. Could you verify that? Made one change request as you can see.

frrad:
Updated to use the logger correctly. I also tried toggling a network under the IRC tab a few times to figure out what to filter. I got [this](https://github.com/ergochat/irc-go/blob/master/ircevent/irc.go#L63)  error one time. 

Questions / Comments:

1. Is there a better way to test the "on disconnect" situation? I only got it the once so I'm not even 100% sure that the condition I wrote works.
2. This error is not exposed by the lib which forced me to do this string matching thing. I'm going to try to ask upstream about making this error public. I wouldn't personally block on this but just wanted to flag it in case you think it would be better to.

frrad:
@slingamn had a suggestion for another way to perform this check over on https://github.com/ergochat/irc-go/pull/99 if you wouldn't mind weighing in, @zze0s

zze0s:
> @slingamn had a suggestion for another way to perform this check over on [ergochat/irc-go#99](https://github.com/ergochat/irc-go/pull/99) if you wouldn't mind weighing in, @zze0s

Thanks for digging a bit deeper into this! 

I do like exported errors since they can easily be checked with `errors.Is`. 

However, like slingamn points out, we do track manual disconnect in our Handler so we might as well use that if the err type won't get exported.

frrad:
Ah, good thinking! I just gave that a try and did not get the new error, just some preexisting debug loglines:

```
{"level":"debug","module":"irc","network":"irc.x.y.z","time":"2023-11-06T12:18:39-08:00","message":"DISCONNECT"}
{"level":"debug","module":"irc","network":"irc.x.y.z","time":"2023-11-06T12:18:39-08:00","message":"Error, disconnected: read tcp 1.2.3.4:5678->9.10.11.12:6697: read: connection timed out"}
```

Looks like the current filter works. (Or maybe this code path is not hit at all? I'm having a bit of trouble believing that `client has called Quit()` is what happens when the network goes away?)

frrad:
@slingamn has identified a race condition that causes the `clientHasQuit` error to be surfaced. See this comment https://github.com/ergochat/irc-go/pull/99#issuecomment-1797348609 and this PR https://github.com/frrad/autobrr/pull/1 of his. He furthermore claims (@slingamn hopefully I'm not putting words in your mouth, please correct me if so) that this error being surfaced is an error qua error and not something that should be returned except in the case of racey client code. I propose the following course of action:

1. Merge this PR, either as-is OR if you buy into this line of reasoning, after removing the current if-statement filter. I think I would personally remove the filter under the theory that this is an actual bug and the error being printed to the log is correctly revealing the bug.
2. Merge https://github.com/ergochat/irc-go/pull/99 because this is an error that might be returned to clients and so it would be good to export it for clients to be able to interact with, regardless of the actual resolution of the real issue here. (I  do think this is the Right Thing To Do™️, but don't feel super strongly about this. Whatever you think of course @slingamn )
3. Open a new PR against this repo with the contents of https://github.com/frrad/autobrr/pull/1 to fix the race-condition that we discovered here.

zze0s:
Thanks for looking further into it and finding bugs and proposing solutions!

If you merge the PR in your fork into the forks `develop` branch it should show up here as well.

slingamn:
I vote for squash-merging https://github.com/frrad/autobrr/pull/1 which should bring it into this PR as a single additional commit. The resulting PR won't depend on either string error comparison or having an exported `ClientIsQuit` error.

I am still undecided about https://github.com/ergochat/irc-go/pull/99, I'll follow up on the thread there later.

frrad:
I'm up for whatever. Just let me know where you land @zze0s . The reason I lean towards separating this from the actual bugfix is that @slingamn described that as a "bandaid" in https://github.com/ergochat/irc-go/pull/99#issuecomment-1797348609 and it felt like it might be worth taking a harder look at the best way to fix the real problem.

I also don't feel like I have a sufficiently good grasp of how things are intended to fit together here to drive that to a good conclusion. Kind of hoping to get in my one-line logging improvement and leave the real work to the experts 😅

KyleSanderson:
> I'm up for whatever. Just let me know where you land @zze0s . The reason I lean towards separating this from the actual bugfix is that @slingamn described that as a "bandaid" in [ergochat/irc-go#99 (comment)](https://github.com/ergochat/irc-go/pull/99#issuecomment-1797348609) and it felt like it might be worth taking a harder look at the best way to fix the real problem.
> 
> I also don't feel like I have a sufficiently good grasp of how things are intended to fit together here to drive that to a good conclusion. Kind of hoping to get in my one-line logging improvement and leave the real work to the experts 😅

It's fairly loose but the works already been done. I would merge their PR so it's in your branch, which we'll then merge. Great finds btw, the issue/reports on this one are older than time

frrad:
I've merged it and updated the PR description. Would you mind updating the title? I'm not sure about your naming convention. Also, if there are any "closes issue/1234" to add to the description that would probably be good too. I did a bit of a dig through the issues and was unable to find any I was sure were relevant.

frrad:
I (probably foolishly) decided to look a bit more at this race condition. I pushed a new commit that moves the unlock-time for our mutex until _after_ the call into the IRC library for both the case of connect and disconnect.

Without doing this I believe there is still a race condition. Namely, 

1. `Connect` tries to connect, gets past the `manuallyDisconnected

THe trade off

frrad:
whoops, misclick on the "close PR" button

slingamn:
I agree that this code probably has a number of race conditions that are not fixed by this patch, including around the creation and recreation of `Handler` objects.

internal/irc/handler.go
```diff
@@ -219,7 +223,17 @@ func (h *Handler) Run() error {
 			func() error {
 				h.log.Debug().Msgf("connect attempt %d", connectAttempts)
 
+				// #1239: don't retry if the user manually disconnected with Stop()
+				h.m.RLock()
+				defer h.m.RUnlock()
```
slingamn, line 228: I recommend against holding a mutex across a network operation like `Connect()`, even if it fixes a TOCTOU race

slingamn:
I don't think either form of this patch will actually fix the "ghost handler" problem: the added check here (don't attempt reconnect if `manuallyDisconnected` was set) stops the code from making reconnection attempts that *could not have succeeded anyway*, since `Connect()` can never succeed after `Quit()` was called. In the form without bf107db99bc2758b729a5995c63067f8fdf16aec, the check just cleans up some logspam, the way that filtering out `ClientIsQuit` would have done. In the form with bf107db99bc2758b729a5995c63067f8fdf16aec I think it creates potential UX problems in that the handler mutex is held for a macroscopic amount of time (possibly seconds).

I would recommend reverting bf107db99bc2758b729a5995c63067f8fdf16aec and then opening an issue to refactor some aspects of the handler system, for example:

1. This check here is probably racey: https://github.com/autobrr/autobrr/blob/73e76c4214261c3eeb1c6f189d7c44d86ac95788/internal/irc/service.go#L127-L130
2. `(*Handler).Restart()` should call `(*Handler).client.Reconnect()` instead of `Stop()` and `Start()`: https://github.com/autobrr/autobrr/blob/73e76c4214261c3eeb1c6f189d7c44d86ac95788/internal/irc/handler.go#L337-L341

zze0s:
There might indeed be more to it, but a minor improvement is good anyway. I'll take a closer look this weekend and see what I can find.

frrad:
I agree with @slingamn that the version without the last commit is better.

The race condition still persists, and we can still get that error in the logs. The window is significantly smaller now though, maybe to the point where it's best to just ignore it. After the revert, this PR at least leaves things strictly better than they were even if it doesn't fully address the problem.

For the follow-up I really do think it bears a little more thinking about how to manage concurrent calls to `Connect` and `Quit`. The library holds the lock while updating its internal record of state on connection:
https://github.com/ergochat/irc-go/blob/master/ircevent/irc.go#L699-L713

and disconnection:
https://github.com/ergochat/irc-go/blob/master/ircevent/irc.go#L335-L338

however, since the lock is not held while the state itself is changed there are no guarantees that the state of the client is kept in sync with the client's record of its state.

slingamn:
I support merging this in its current form.

I took a more detailed look at how autobrr uses irc-go and I think in the medium term, improvements to irc-go are needed. Right now, the `(ircevent.Connection)` object is hard to use correctly in the autobrr context:

1. The current API is that the user has to call `Connect()` and check the return value. On success, the object is valid (until `Quit()`) and the user has to call `Loop()`. On failure, the user can retry or fail. Deviating from this model creates problems:
2. Immediately after instantiation (before the first call to `Connect()`), *most* functionality is a no-op, but `Reconnect()` will panic
3. There's no API to stop the connection and reconfigure it (after `Quit()`, reconnection is impossible)
4. `Connected()`, which returns a bool, doesn't provide enough information to autobrr about the state of the connection. There are really at least three states: stopped, trying to connect, and connected.

I think irc-go should be extended in some of the following ways (not sure which ones yet):

1. Provide a new API (either `NewConnection()` or `(*Connection).Initialize`), after which the connection is in a fully valid state and all methods have well-defined behavior
2. Provide an API that can distinguish between the three connection states
3. Provide an API that stops the connection and waits for stop (probably we want to call the existing `Quit()` method, wait up to a timeout for the server to receive and process the QUIT line, but then close the connection and shut down if the timeout is exceeded)
4. Start/maintain `Loop()` automatically after `Connect()`? For compatibility, a public `Loop()` method could be provided that maintains the current behavior (sleep until `Quit()`)

This would allow autobrr's `Handler` object to maintain a single `*Connection` pointer, instead of recreating the object as needed (right now there are data races around replacing the pointer, and ensuring that the referent is in a valid state would require nil guards).

slingamn:
I took yet another look at this and I think some of the goals I outlined above (in particular, being able to stop, reconfigure, and restart connections) are invalid. I also drafted a PR against the current irc-go API that I believe will fix all the relevant races:

https://github.com/slingamn/autobrr/commit/5368f8df9d2981b15e092d04800cfad1eae44879

The core ideas are:

1. Only allow one `Run()` to proceed at a time
2. Acknowledge that `(Handler).client` can be invalid in some cases; manage this explicitly with nil guards

I can submit this as a separate PR after this one is merged.

zze0s (approved):

PR #1243, opened 2023-11-02
feat(feeds): add force run

s0up4200:
Add the ability to force run feeds - while warning users about the potential consequences.

CleanShot 2023-11-03 at 22 52 31


s0up4200:
- Added user input requirement to run.
- Made sure to reschedule after run.

Added demonstration to the OP.

I am unsure about this one though. It will slow down the process for starters.
https://github.com/autobrr/autobrr/blob/c63440b2730ca38a3e6f0ad070572c45c6796f0c/internal/feed/service.go#L518-L521

zze0s (approved):
Great improvement!

Made some changes on the backend and frontend.

PR #1125, opened 2023-09-16
feat(web): add a fully-featured irc client to the ui

ghost:
image

todo:

- [x] add license headers
- [ ] implement `/api/irc/cmd` (execute cmd)
- [ ] handle new sse format
- [ ] ...?

Issue #1114, opened 2023-09-11
Setup IRC CTCP replies


zze0s:
Enable [CTCP replies](https://thelounge.chat/docs/guides/ctcp) with the IRC client.

Kuredant:
Is there any concern other than setting `EnableCTCP` in `&ircevent.Connection{}`?

According to [irc-go/ircevent](https://github.com/ergochat/irc-go/blob/8d1f09a1b8d46e018ef369327057211860565e0e/ircevent/irc_ctcp.go#L48), this should automatically handle `VERSION`, `USERINFO`, `CLIENTINFO`, `TIME` and `PING`.

KyleSanderson:
> Is there any concern other than setting `EnableCTCP` in `&ircevent.Connection{}`?
> 
> According to [irc-go/ircevent](https://github.com/ergochat/irc-go/blob/8d1f09a1b8d46e018ef369327057211860565e0e/ircevent/irc_ctcp.go#L48), this should automatically handle `VERSION`, `USERINFO`, `CLIENTINFO`, `TIME` and `PING`.

Yeah, and I'll document it here. We've been incredibly cautious about not making (visible) mistakes on networks and have limited mechanisms to actually send data. The only real place we attempt multiple authentication schemes is with NickServ, which too is limited. The last thing we want is for autobrr to be a weaponized vector on networks.

TIME and PING give relative server location which for our actual user I'm sure they'd prefer not to do.

If we have a bug, VERSION can be the equivalent of the nginx server string (which can contain more than the version, the version, or a confirmation that it's indeed autobrr). Which could aid in the weaponization of the product.

USERINFO and CLIENTINFO I also don't see it adding any value for our end users.

So, I'm not vetoing this or saying no, but without a good reason I see it as (likely) actively harmful to add. I heard that there was some request in passing from a help desk member, if any staff would like to have a discussion about our behaviour on networks we're (very) open to it.

Kuredant:
Thanks for explaining your reasoning.
I don't have any argument in favor or against implementing CTCP replies, I was just wondering what the status of this issue was.

PR #1246, opened 2023-11-05
feat(irc): support optional bot mode

Kuredant:
Set [IRCv3 Bot Mode](https://ircv3.net/specs/extensions/bot-mode) if the server supports it.
Some servers recommend setting bot mode when the client is a bot.

This checks for an `ISUPPORT` `BOT` token and sends `MODE +{BOT_TOKEN}` (usually `MODE +B`) before authentication.


zze0s (commented):
Hi! Thanks for the PR!

I'm not sure if this should be automatic or behind a toggle, since it might not always be wanted.

KyleSanderson:
Indeed, we are working on interactivity in https://github.com/autobrr/autobrr/pull/1125 . From an onboarding perspective it's the largest gap we have with the product today. I don't see an issue with this being configurable per network, but do see limited value for our end users. Same thoughts as https://github.com/autobrr/autobrr/issues/1114#issuecomment-1793807155 applies - very open to a discussion.

KyleSanderson:
> Considering that autobrr _is_ a bot, I would like to have this functionality available, even as a toggle. Some private IRC networks require bots to set bot mode, and it bothers me that I can't do that with autobrr.

https://ircv3.net/specs/extensions/bot-mode
`This functionality was previously implemented by several IRC servers in independent ways using fixed mode letters and conflicting WHOIS numerics. This specification seeks to standardise the functionality, adding a message tag for increased visibility of bot status and an ISUPPORT token to allow for more flexible implementations.`

Totally get it. I'm sure you can see from a code perspective why this isn't necessarily going to always be correct (and, in fact, will do exactly the wrong thing on some servers). We can certainly add some sort of mode or toggle as an expert feature, but it's prime for user error and misuse which is why it's presently seen as adding little to no value to actual end users.

zze0s:
> Considering that autobrr _is_ a bot, I would like to have this functionality available, even as a toggle. Some private IRC networks require bots to set bot mode, and it bothers me that I can't do that with autobrr.

Do you happen to have a list of indexers that have it as a requirement? And if it is a requirement and autobrr doesn't support it, is there any repercussions for not doing so?

The v3 spec doesn't say much about what a bot IS. To me autobrr is a passive client in it's current form (could change in the future). Wouldn't a bot have to have some respond functionality from commands?

What would the benefit be of marking the client as a bot? Could it be targeted?

Slikkster:
> @Slikkster Do you happen to have a list of indexers that have it as a requirement? And if it is a requirement and autobrr doesn't support it, is there any repercussions for not doing so?
> 
> The v3 spec doesn't say much about what a bot IS. To me autobrr is a passive client in it's current form (could change in the future). Wouldn't a bot have to have some respond functionality from commands?
> 
> What would the benefit be of marking the client as a bot? Could it be targeted?

I don't have a list. The only one I know off the top of my head is BHD, they list it in their MOTD: `- All bots should set user mode +B (/mode botnick +B)`. I would have to look through the rules for the others. So far there isn't any repercussion for not doing so that I'm aware of; especially since (I assume) a significant percentage of users are using autobrr.

Your comment re: autobrr being a passive client actually makes a lot of sense to me now that I think about it. I previously assumed "it's not a standard IRC client, therefore it must be a bot", but I guess it's not so black and white. Autobrr doesn't send any commands or interact beyond signing in and joining the necessary channel, so in that sense it isn't a typical IRC bot.

I don't know what the benefit of marking the client as a bot would be, beyond (my previous thinking) of complying with the site's rules. Now I'm not sure if those rules even apply?

Slikkster:
> > Considering that autobrr _is_ a bot, I would like to have this functionality available, even as a toggle. Some private IRC networks require bots to set bot mode, and it bothers me that I can't do that with autobrr.
> 
> https://ircv3.net/specs/extensions/bot-mode `This functionality was previously implemented by several IRC servers in independent ways using fixed mode letters and conflicting WHOIS numerics. This specification seeks to standardise the functionality, adding a message tag for increased visibility of bot status and an ISUPPORT token to allow for more flexible implementations.`
> 
> Totally get it. I'm sure you can see from a code perspective why this isn't necessarily going to always be correct (and, in fact, will do exactly the wrong thing on some servers). We can certainly add some sort of mode or toggle as an expert feature, but it's prime for user error and misuse which is why it's presently seen as adding little to no value to actual end users.
> 
> EDIT: Back to the PR at hand, it does look correct, but I think it just needs a switch added to hit the "best of both worlds" scenario.

I actually wasn't aware of the problems around bot mode with IRC until now. Makes complete sense why you'd be hesitant about implementing it. Seems like it may not even apply to autobrr anyways -- was just my assumption that it did.

Kuredant:
> Indeed, we are working on interactivity in https://github.com/autobrr/autobrr/pull/1125 . From an onboarding perspective it's the largest gap we have with the product today. I don't see an issue with this being configurable per network, but do see limited value for our end users. Same thoughts as https://github.com/autobrr/autobrr/issues/1114#issuecomment-1793807155 applies - very open to a discussion.

The fully-featured IRC client will be there mainly for debugging/setup purposes, right? IMO, even with the possibility to interact manually, most of the functionality of autobrr is automatic, so setting the bot flag would make sense.

Though I've never seen any server *require* a specific mode to be set on bots, it's usually just a recommendation, so it would indeed add limited value.

> Totally get it. I'm sure you can see from a code perspective why this isn't necessarily going to always be correct (and, in fact, will do exactly the wrong thing on some servers). We can certainly add some sort of mode or toggle as an expert feature, but it's prime for user error and misuse which is why it's presently seen as adding little to no value to actual end users.

This PR sends `MODE +b` if the server sets an ISUPPORT tag `BOT=b`, i.e., the server exposes which mode letter is used for setting bot mode, which we'll use. Although I understand your concerns, I don't see how this can do the wrong thing.

> EDIT: Back to the PR at hand, it does look correct, but I think it just needs a switch added to hit the "best of both worlds" scenario.

That makes sense; I will add a switch to enable the feature.

config.toml
```diff
@@ -57,6 +57,12 @@ logLevel = "TRACE"
 #
 checkForUpdates = true
 
+# Bot mode
+#
+# Default: false
+#
+enableBotMode = false
+
```
zze0s, line 65: This should not be a global config but rather per network and stored in the database.

zze0s (changes requested):
The bot mode should be per network and stored in the `irc_network` table in the database, not in the config file which would make it global for all networks.

Kuredant:
> The bot mode should be per network and stored in the `irc_network` table in the database, not in the config file which would make it global for all networks.
> 
> If you need help with database migrations let me know, but it's pretty simple. Add new field to initial schema and a new migration at the bottom in both `sqlite_migrate.go` and `postgres_migrate.go`.

Sure, I can make it configurable per network.
In that case, though, the option will do nothing for networks that don't send an ISUPPORT BOT token.

zze0s:
> Sure, I can make it configurable per network. In that case, though, the option will do nothing for networks that don't send an ISUPPORT BOT token.

That would be the case either way, right?

Kuredant:
> > Sure, I can make it configurable per network. In that case, though, the option will do nothing for networks that don't send an ISUPPORT BOT token.
> 
> That would be the case either way, right?

Yes, that's right.
I wasn't sure whether it should be a global or per-network setting, but it does indeed seem better to have it in the network edit panel.

config.toml
Kuredant: This option is now in the `irc_network` table, configurable from the network edit panel.

internal/irc/handler.go
```diff
@@ -492,6 +502,20 @@ func (h *Handler) handleNickServ(msg ircmsg.Message) {
 	}
 }
 
+// botModeSupported checks if IRCv3 Bot Mode is supported by the server
+// See https://ircv3.net/specs/extensions/bot-mode
+func (h *Handler) botModeSupported() bool {
+	h.botModeChar = h.client.ISupport()["BOT"]
```
zze0s, line 508: Could this fail if it's named something else? Could be safer to check the map first for the result and only set if it's there.

zze0s (approved):
Code looks sane and I ran some tests and it seem to work as expected.

Thanks for your contribution!

internal/irc/handler.go
Kuredant: [`Connection.ISupport()`](https://github.com/ergochat/irc-go/blob/5474a63ef6b9e03232499f9d8b6fb646a6a2d7f5/ircevent/irc.go#L531C29-L531C29) returns a `map[string]string`. If `"BOT"` is not found in the map, it will simply return an empty string (the default string value), so it should already be safe.  
A similar check is done in the [irc-go samples](https://github.com/ergochat/irc-go/blob/5474a63ef6b9e03232499f9d8b6fb646a6a2d7f5/ircevent/examples/simple.go#L42), for example.

Issue #1247, opened 2023-11-05
Autobrr doesn't shutdown gracefully when stopped using its systemd service


Antorell:
**Version**
1.33.0

**Describe the bug**
The autobrr systemd service returns a failure when stopped using its default kill signal (SIGTERM) instead of shuting down without returning an error (gracefully). Any other sane kill signal also returns an error. 

**To Reproduce**
Enable the autobrr systemd service, 
Start the autobrr service and stop it. 

**Expected behavior**
The systemd service shouldn't return a `status=1/FAILURE` on exit when it is gracefully stopped. 

**Screenshots**
```
Nov 05 21:18:48 Ubuntu systemd[1]: Started autobrr.service - autobrr service for .
Nov 05 21:18:48 Ubuntu systemd[1]: Stopped autobrr.service - autobrr service for .
Nov 05 21:18:48 Ubuntu systemd[1]: autobrr.service: Failed with result 'exit-code'.
Nov 05 21:18:48 Ubuntu systemd[1]: autobrr.service: Main process exited, code=exited, status=1/FAILURE
Nov 05 21:18:48 Ubuntu systemd[1]: Stopping autobrr.service - autobrr service for ...
Nov 05 21:18:41 Ubuntu systemd[1]: Started autobrr.service - autobrr service for .
Nov 05 21:18:41 Ubuntu systemd[1]: Stopped autobrr.service - autobrr service for .
Nov 05 21:18:41 Ubuntu systemd[1]: autobrr.service: Failed with result 'exit-code'.
Nov 05 21:18:41 Ubuntu systemd[1]: autobrr.service: Main process exited, code=exited, status=1/FAILURE
Nov 05 21:18:41 Ubuntu systemd[1]: Stopping autobrr.service - autobrr service for ...
```

**Desktop (please complete the following information):**
 - OS: Ubuntu 23.10
 - Browser Edge
 - Version 119

**Additional context**

I believe this is the reason why systemd keeps complaining about the failures when the service is stopped. The SIGTERM exit code in the cases is 1 when it should be 0. I tried to compile from the master tree with the exit code changed to 0 but I sucked at building with Go, so I couldn't try myself. 

https://github.com/autobrr/autobrr/blob/8c89481d880c2bcd6d3907c16af807ec0a86a7a0/cmd/autobrr/main.go#L151-L166

zze0s:
Fixed with #1248

PR #1248, opened 2023-11-06
refactor: streamline shutdown signal handling

s0up4200:
Refactored the server shutdown code to handle all signals in a unified way and removed the unnecessary `SIGKILL` case. `SIGKILL` cannot be caught, so its inclusion is unnecessary. On database closure error, we exit with status code `1` to signal a fault, otherwise, we exit with `0` for a clean shutdown.

Ref #1247


zze0s (approved):
I'm unsure why exit 1 was chosen from the start but this should be a decent improvement! 👍

PR #1250, opened 2023-11-07
fix(filters): validate external filters

### LINKED ISSUE #1224: HTTP Method "POST" set to Default
**Version**
Updated :develop branch

**Describe the bug**
Setting up webhook to POST as default

**Screenshots**
image

Actually not an issue just a possibility to set to default.

Thank you

s0up4200:
Updated Formik initial values to ensure `webhook_method` defaults to `POST` for external filters. This prevents undefined values by automatically setting POST upon form submission if the user hasn't selected a method.

Left `optionDefaultText="Select http method"` as is.

Fixes #1224


zze0s (approved):
Explicitly setting the `webhook_method` in initial values would apply it to both Exec and Webhook types.

I removed that and instead added schema validation so there will instead be an error if it's not set. Added some other fields to validate as well.

PR #1253, opened 2023-11-09
feat(tests): verify external autobrr.com URLs

s0up4200:
A test aimed at verifying that all references to autobrr.com are resolving as expected. The test makes HTTP GET requests, checking for any that return a 404 status. This should help maintain the integrity of our documentation links.


KyleSanderson:
Excellent start! What you likely want to do is to scan through the .go files and pull links from there. Maybe there's a grep library to help facilitate this?

KyleSanderson (approved):

zze0s (changes requested):
Cool and useful idea!

Minor thing but it does not really belong in the `http` package since it checks external links.

Could you move it into  a new dir called `docs` in the test dir? 😄

s0up4200:
> Cool and useful idea!
> 
> Minor thing but it does not really belong in the `http` package since it checks external links.
> 
> Could you move it into a new dir called `docs` in the test dir? 😄

Done 👍

zze0s (approved):

PR #1254, opened 2023-11-09
fix(filters): external webhook remove jitter

buroa:
Tested and works great.

Semi-dupe of #1238


buroa:
@KyleSanderson Can you look this over, please?

zze0s (commented):
Looks good, haven't tested it myself tho.

Only comment is to add a down migration to remove the column that was added, and remove it from the initialSchema.

zze0s (approved):

PR #1256, opened 2023-11-13
feat(indexers): update base url for Milkie

mynameisbogdan:
This is needed to prevent redirects www to non-www.


zze0s (approved):
I read your report of this being an issue with the arrs with rediects, nice find!

PR #1257, opened 2023-11-14
feat(definitions): add expectations for test lines

frrad:
I was playing around with modifying one of these regular expressions and noticed that these test lines don't seem to be checked against anything.

This PR is a functional noop which adds the option to specify expectations for how test lines will be parsed.

In order to make this change I refactored, moving RE parsing out of the `announce` package to the `indexer` package.

Incidentally, I believe that I uncovered an actual parsing bug. Example 5 seems wrong to me.


internal/indexer/definitions/animebytes.yaml
```diff
@@ -67,6 +67,133 @@ irc:
...
+          # 5
+          - baseUrl: "https://animebytes.tv/"
+            category: "Live Action Special - Live Action TV Special"
+            freeleech: ""
+            groupId: "00000"
+            releaseEpisode: ""
+            releaseGroup: "Sub Group"
+            releaseTags: "Web / MKV / h264 / 848x480 / AAC 2.0 / Softsubs (Sub Group)"
+            tags: "manga"
+            title: "Awesome "
```
frrad, line 150: It seems that `title` and `category` are not divided as intended.

internal/domain/indexer.go
```diff
@@ -208,10 +208,11 @@ type IndexerIRCParse struct {
 }
 
 type IndexerIRCParseLine struct {
-	Test    []string `json:"test"`
-	Pattern string   `json:"pattern"`
-	Vars    []string `json:"vars"`
-	Ignore  bool     `json:"ignore"`
+	Test         []string             `json:"test"`
+	Expectations *[]map[string]string `json:"expectations"`
```
zze0s, line 212: It might make sense to change Test from a `[]string` into a new struct that combines the case and expectations, and not sure the Expectations needs to be a pointer 🤔 

```go
type LineTest struct {
    Line   string            `json:"line"`
    Expect map[string]string `json:"expect"`
}
```

That and add a new Tests field, but keep Test to not break unofficial custom definitions.

```go
type IndexerIRCParseLine struct {
	Test    []string   `json:"test"`
        Tests   []LineTest `json:"tests"`
	Pattern string     `json:"pattern"`
	Vars    []string   `json:"vars"`
	Ignore  bool       `json:"ignore"`
}
```
Something like that maybe?

zze0s (commented):
Good stuff! This should have been done from the start but it was only halfway done since the expects were missing 😄

I started work on [this branch](https://github.com/autobrr/autobrr/tree/fix/lossless-parsing) earlier this year to improve the IRC parsing and add support for parsing per indexer to help with special cases like AB, GGn and RED. I moved the parsing from announceparser into indexers there as well. It needs more work however, so not ready yet.

frrad:
Collecting the examples at all is definitely more than half the battle :)

Your structure is much nicer, good suggestion. The pointer bit was to distinguish between absent and zero-length `Expectations` but the new struct sidesteps the need for that very nicely.

frrad:
Threw in some a bonus simplification refactor https://github.com/autobrr/autobrr/pull/1257/commits/39cf81c363a3634edbdeb830096fa6e896493d8b let me know if you want that to be a separate PR.

frrad:
I reordered the expectation to match the order in which vars are defined which I think makes it easier to read. I would suggest adopting that as the convention going forward.

LMK if there's anything else you need to merge this! :) I'm planning a follow-up to fix the parsing bug but I will wait for this to be merged first so it can be nice and clean.

zze0s:
> I reordered the expectation to match the order in which vars are defined which I think makes it easier to read. I would suggest adopting that as the convention going forward.

Was going to mention but you beat me to it! 

> LMK if there's anything else you need to merge this! :) I'm planning a follow-up to fix the parsing bug but I will wait for this to be merged first so it can be nice and clean.

It's a big task/ask but it would be nice to fix all the definitions in this PR. We can split the work if want and do half each or something like that? Otherwise I can probably get it all done during the weekend.

frrad:
If we're going to do this all at once can we do a breaking schema change and kill `Test` in favor of `Tests`? I'm not sure about the "unofficial custom definition" concern you raised earlier, but this seems like an opportunity to avoid a little bit of cruft in the schema that would otherwise linger for the rest of time. The breakage doesn't seem too bad to me. Worst case people can just remove test cases altogether...

Either way, I'll take a stab at converting everything. I have some stuff coming up later this week so if you don't hear back from me by this weekend assume I've gotten distracted and you're on your own. Hopefully I can knock this out real quick first though. I'm thinking about just writing some throwaway code to generate it.

zze0s:
Good point, they were not used before in code and if they are not present they won't parse so that should be fine 🤔 

Code generation sounds like a great use :D

frrad:
So it turns out that it's not totally trivial to round-trip this yaml unmarshal -> marshal this yaml because of ordering in the maps. I hacked together something to generate the expectations (ordered correctly wrt the var definitions) and have started just merging these into the YAMLs by hand. It's all checked in so if I don't finish you should be able to pick this up easily. It's just a bunch of copy / paste / run `go test ./internal/indexer` at this point.

KyleSanderson:
This is excellent 😸 are some leaving the definitions folder though? Ideally tests and definition are single file, with the single runner in the same folder (?).

frrad:
The ones up a directory are the output of my helper script. I'm in the process of manually merging them back into the definitions. They'll be gone once I get to `Z`

KyleSanderson:
Ah, my apologies - thank you for doing this.

frrad:
No worries!

Should be good to go now. That was really painful lol. I definitely would have tried harder to get the end-to-end script if I had realized how darn many definitions there were 😅 !!! 

Anyways, one last thing that I wanted to make sure didn't get buried under all the `.yaml` changes:

I made a change to `internal/indexer/service.go` to have it throw an error when it encounters an unknown field in the YAML to give myself more confidence that I didn't introduce typos while doing all these yaml changes. 

1. I think we should leave this setting on. It will help catch yaml typos in the future and there's no good reason to have fake keys floating around in the YAML.  
2. I commented out some things in the YAML which caused this error to be thrown. I think we should just leave that as is for this PR which is already super bloated. Maybe file an issue or do a followup PR to clean these up. (Some probably contain information that should be preserved as comments which is why I didn't just delete them.) Alternatively, just get to it as you see it in the course of making future changes to the YAMLs.
3. There's a decent argument for leaving this setting off in the case of parsing custom `yaml`. This would make this PR non-breaking for people's custom YAML since:
    1. The change from `test` -> `tests` would be non-breaking since content under the `test` key would be silently ignored and `tests` is not actually consumed by anything but the new unit test.
    2. Existing extraneous keys would not throw a parse error once this PR is merged.
    
    I do think the strict parsing behavior is better even for custom YAML since in the same way that it will help catch typos in PRs it will help people catch typos in their own cutom YAML as they write them, but not sure it's worth a breaking change.

frrad:
Just adding the same "version: 1" line to the top of everything is easy to script without having to parse anything 😄 That said...

I think it would be nice to avoid explicitly versioning the schema in the case of `IndexerDefinition` if possible. These are all checked in which provides a really nice opportunity for you to just enforce a policy of "the schema is `HEAD`, backwards incompatible schema changes must be accompanied by changes to all the checked in yaml." Since the only way to get a `IndexerDefinition` is through the embedded fs, it should be fine to throw the error. You're not going to be breaking anyone.

Of course,  `IndexerDefinitionCustom` is a lot more complicated since you don't control the YAML. I think the fundamental question here is how do you want to handle consuming YAMLs from previous schemas after breaking changes. Do you want to maintain a compatibility layer that does best-effort translation into the current schema? You'll have to do something like this if you want to be able to make breaking schema changes without breaking custom YAML. Seems tricky.

I really encourage you to take the cop-out position that this is _not_ in fact a breaking change (which it isn't if we just ignore extra keys like I suggest in 3 above) and punt on solving this until you run into something significant enough to warrant it. I think you might actually be able to get quite far with a policy of "no _actually_ breaking schema changes". Most things can be done this way. In fact, that's how this PR started way back in https://github.com/autobrr/autobrr/pull/1257/commits/aba75b551c232f5d0d406e1d586737c02d82c5fb

KyleSanderson:
> Just adding the same "version: 1" line to the top of everything is easy to script without having to parse anything 😄 That said...
> 
> I think it would be nice to avoid explicitly versioning the schema in the case of `IndexerDefinition` if possible. These are all checked in which provides a really nice opportunity for you to just enforce a policy of "the schema is `HEAD`, backwards incompatible schema changes must be accompanied by changes to all the checked in yaml." Since the only way to get a `IndexerDefinition` is through the embedded fs, it should be fine to throw the error. You're not going to be breaking anyone.

Correct - we do indeed control these.

> Of course, `IndexerDefinitionCustom` is a lot more complicated since you don't control the YAML. I think the fundamental question here is how do you want to handle consuming YAMLs from previous schemas after breaking changes. Do you want to maintain a compatibility layer that does best-effort translation into the current schema? You'll have to do something like this if you want to be able to make breaking schema changes without breaking custom YAML. Seems tricky.

Not really, this is what the version field does. If we require it for external versions and not internal, all of a sudden new indexer developers to the project don't actually have a published starting place. This is why treating them the same, as silly as it is, actually helps the next developer coming in get ramped quickly.

> I really encourage you to take the cop-out position that this is _not_ in fact a breaking change (which it isn't if we just ignore extra keys like I suggest in 3 above) and punt on solving this until you run into something significant enough to warrant it. I think you might actually be able to get quite far with a policy of "no _actually_ breaking schema changes". Most things can be done this way. In fact, that's how this PR started way back in [aba75b5](https://github.com/autobrr/autobrr/commit/aba75b551c232f5d0d406e1d586737c02d82c5fb)

I agree that this needs to be done, but there's the wild west out there for definitions (primarily private Usenet and some tiny trackers). The ones we know about will error, which is why the version and path to deprecation is still the path forward imo. There's going to be users on older versions, using the latest and greatest definition file, and the version mismatch is a very easy way to tell them their application is out of date.

I don't want to force you to do something you feel is wrong, so we can land this as-is, but before the next release we're going to have to make a firm decision on do we ship with definitions failing if we don't understand them in their entirety.

frrad:
If you're all right with merging this as-is (I put in the "option 3" change as https://github.com/autobrr/autobrr/pull/1257/commits/a241553b41a794e7c72a1baefbaf31c97301b27b so note that current as-is is different from as-is at the time of your last comment. I'm cool with either let me know if you want me to revert it) I think that would be my preference. I think that this issue of schema-versioning deserves its own PR.

KyleSanderson (dismissed):

frrad:
That said, because I can't resist the opportunity to bloviate, let me respond to the substance of your comment with some random takes.

I agree with you that it makes sense to require even the internal `yaml`s to have versions so that external developers have a starting place.

However, I think it is okay to hold internal yaml to a higher standard than external yaml since there is a much tighter feedback loop. (Test fails pre-merge versus unexpected breakage).

If it were me, instead of using a numeric version field, I would use a [semver](https://pkg.go.dev/golang.org/x/mod/semver) string. 
* have parsing fail for internal yaml if they don't declare `version == latestVersion`
* bump the major version when you make breaking changes
* have parsing fail for external yaml if they declare a previous major version
* consider starting with a `0.xx` major version, indicating that the schema is not yet stable. (maybe just go for `1.0` and everything previous is implicit `0.0` haven't really thought this one through

Additionally, I would add a test that uses reflection to compare the actual `IndexerDefinition` struct with some hardcoded copy of its current state. Then, when this test fails you need to
1. think about the impact of parsing an old yaml in the new schema
2. do at least a minor version bump 
3. update the test to hardcode the new schema

Without this test you leave yourself open to not remembering to bump your version / not thinking through the consequences of schema changes.

frrad:
What is the merge process you all follow? I would love to merge it if it's good to go so I can rebase https://github.com/autobrr/autobrr/pull/1259

zze0s:
> What is the merge process you all follow? I would love to merge it if it's good to go so I can rebase #1259

Usually I do the final review and merge. Finally got some time this weekend to go over it all.

We try to do releases every couple of weeks, usually every 3rd week if nothing critical has been found. Aiming for a release on sunday.

zze0s (approved):
Great work! This will make the process of adding indexers and changing them a lot easier :clap: 

For some reason the embedded struct for custom definition did not load properly with the yaml.Decode, no idea why. Changed it back and now it works. :man_shrugging:

PR #1263, opened 2023-11-18
Validate Filter Size

frrad:
I noticed that there is no validation that filter sizes are parseable.

Prior to this PR, in the "additionalSizeCheck" case, parse problems would be silently ignored, leading to `minSize` not being enforced. In the not-"additionalSizeCheck" case, a filtering-time error would be thrown. 

In this PR, I:
* unify size checking in a new `f.ReleaseSizeOkay` method which is called from both codepaths
* add a new `f.Validate` method, which unifies pre-existing validation from `Update` with new size parseability validation
* add a call to `f.Validate` in `Store` where there was previously no validation

plus bonus:
* make a logging fix in commitXXX that I discovered in the course of testing


internal/domain/filter.go
```diff
@@ -566,41 +580,19 @@ func (f Filter) checkSizeFilter(r *Release, minSize string, maxSize string) bool
...
-			r.addRejectionF("size: invalid maxSize set: %s err: %q", maxSize, err)
-			return false
-		}
-
-		if r.Size >= maxSizeBytes {
-			r.addRejection("size: larger than max size")
-			return false
-		}
+	if !ok {
+		r.addRejection("size: release size out of bounds")
```
zze0s, line 594: This generic rejection might makes it harder for users to interpret if it was below or above their filters. With these current changes we don't expose the parsed min,max bytes. Thoughts?

internal/filter/service.go
```diff
@@ -149,10 +146,13 @@ func (s *service) GetDownloadsByFilterId(ctx context.Context, filterID int) (*do
 }
 
 func (s *service) Store(ctx context.Context, filter *domain.Filter) error {
-	// validate data
+	err := filter.Validate()
+	if err != nil {
+		s.log.Error().Err(err).Msgf("invalid filter: %v", filter)
+		return err
+	}
```
zze0s, line 153: Might as well do it inline.

```go
	if err := filter.Validate(); err != nil {
		s.log.Error().Err(err).Msgf("invalid filter: %v", filter)
		return err
	}
```

internal/filter/service.go
```diff
@@ -149,10 +146,13 @@ func (s *service) GetDownloadsByFilterId(ctx context.Context, filterID int) (*do
...
 
-	// store
-	err := s.repo.Store(ctx, filter)
+	err = s.repo.Store(ctx, filter)
```
zze0s, line 155: And same with this

```go
	if err := s.repo.Store(ctx, filter); err != nil {
		s.log.Error().Err(err).Msgf("could not store filter: %v", filter)
		return err
	}
```

zze0s (commented):
Overall good improvements but left some comments.

internal/domain/filter.go
frrad: So, I think the interface is cleanest as it was originally. It does hide the rejection details though. 

What do you think about this new version where we transmit some details up in an `error`? I'm not in love with tightly coupling the error with the text of the rejection reason like I did. 

Another option I considered was to return the too small / okay / too big info in an `int` a la https://pkg.go.dev/math/big#Float.Cmp or something, then letting the caller build the error message using the `f.MaxSize` string / even expose the parsing method and use that. 

IDK WDYT

zze0s: Much better now! Thank you!

zze0s (approved):
Good improvement!

PR #1265, opened 2023-11-18
builds(ci): containers build additional architectures

s0up4200:
### Changes to `Dockerfile`:

- Reducing the number of `RUN` commands decreases the number of layers in the Docker image. Fewer layers can lead to a smaller image size and potentially faster build times, as Docker processes fewer steps.

### Changes to `ci.Dockerfile`:

- Cached Go modules effectively by first copying `go.mod` and `go.sum`, then running `go mod download` before copying the rest of the source code. This approach should leverage Docker's layer caching mechanism, avoiding unnecessary re-downloading of dependencies when these files remain unchanged.
- Combined the `RUN` commands for the same reason as in `Dockerfile`.


KyleSanderson (approved):

KyleSanderson (dismissed):

.github/workflows/release.yml
```diff
@@ -208,20 +194,71 @@ jobs:
...
+          touch "/tmp/digests/${digest#sha256:}"          
+      - name: Upload digest
+        uses: actions/upload-artifact@v3
+        with:
+          name: digests
+          path: /tmp/digests/*
+          if-no-files-found: error
+          retention-days: 1
+
+  docker-merge:
```
zze0s, line 228: Why can't this be part of the `docker` job? No upload/download needed if it could.

zze0s (approved):

PR #1272, opened 2023-11-20
refactor(filters): move rejections from release to filter

frrad:
This is a follow-up tweak to https://github.com/autobrr/autobrr/pull/1263 I think this is a little cleaner and we don't really need to distinguish these errors like I had it originally.


zze0s (approved):
I took a good look at this and realized it could be improved further. It turned into a decent refactor to move from rejections on the release to the filter, which is more correct for how autobrr works in it's current form.

Changed some of the logging to be more clear and concise.

I also found a pretty major bug with the `findByIndexerIdentifier` method to find filters. If you had more than 1 external filter it would return duplicate rows since it was lacking any duplicate checks, but that has been taken care of now.

I'll leave this open for short time if you have any feedback. If not then I think this is ready to be merged. All tests pass and did manual testing as well.

frrad:
I don't have good enough of a grasp of how things all fit together to comment on the totality of this change but some random questions / comments that I had reading your change:

1.  You should update the PR description now that this PR actually does something more substantial :) I think  a little commentary about the history for rejections living on releases and why you're moving them to filters now would be interesting and useful to future historians. One thing (and maybe this is just my lack of context, feel free not to explain the whole world for my sake) that's not clear to me is - where are these rejection reasons eventually surfaced in the UI and how does this refactor impact that? Also, naively it seems like rejections belong to the cross product of filter x release. But it doesn't seem like both pieces of data are persisted in either the "before" or "after" worlds.
1. In https://github.com/autobrr/autobrr/pull/1263#discussion_r1398415222 we moved away from a `(bool, error)` return to `(error, error)` so that we could use the error to smuggle out more details for the rejection reason. We're moving back now with this PR now and writing the rejection reason directly to the filter. 
    1. Is it okay for this (new logging to the filter) to be happening in all codepaths that call this function?
    2. related to ^ - philosophically, what part of the code should be in charge of writing rejection reasons? Should we write at most one rejection reason per release x filter pair? It seems like it could be potentially confusing if any given helper can write its own rejection reason.
    3. nit: If the return value is a bool I would probably name it something other than "check". To me it's less clear that "check" true means that we passed, than if the function were something like "size okay." "check = true" could mean "we performed the check" without indicating whether or not we passed the check. :shrug: 
2. Nit: you could consider adding a test for the bug you fixed.

zze0s:
Will try to give some context and explain!

> 1. You should update the PR description now that this PR actually does something more substantial :) I think  a little commentary about the history for rejections living on releases and why you're moving them to filters now would be interesting and useful to future historians. One thing (and maybe this is just my lack of context, feel free not to explain the whole world for my sake) that's not clear to me is - where are these rejection reasons eventually surfaced in the UI and how does this refactor impact that? Also, naively it seems like rejections belong to the cross product of filter x release. But it doesn't seem like both pieces of data are persisted in either the "before" or "after" worlds.

Currently **filter rejections** are not surfaced or shown anywhere in the UI except the logs.
However, **action rejections**, which does run after all filter checks pass does surface as **action status** on releases. 

To new users this is a bit confusing and they often wonder why no releases shows up, but it's only the filtered releases that makes it to an action which are saved and will show up in releases.

The reason why it was built like that is because otherwise ALL processed releases would show up and if you are on a couple of general indexers it would be thousands or tens of thousands of entries per week and that would clutter the database with pretty much useless junk (to most people).

There is a feature request to have that as an opt-in feature and that will eventually be implemented.

There's also feature requests about having a release log per filter so users can actually see per checked filter what was rejected. This change will make that easier to implement.

> 2. In [feat(filters): validate filter size #1263 (comment)](https://github.com/autobrr/autobrr/pull/1263#discussion_r1398415222) we moved away from a `(bool, error)` return to `(error, error)` so that we could use the error to smuggle out more details for the rejection reason. We're moving back now with this PR now and writing the rejection reason directly to the filter.

Good questions for sure! When I looked at this PR and the changes I realized that the rejections are per filter and actually not per the release. It did even reset the release rejections per check in `CheckFilter` so that made it more clear it did not belong per release. 

> 2.1. Is it okay for this (new logging to the filter) to be happening in all codepaths that call this function?

I'd say where it makes sense, but that might be a bit loose :smile: 

> 2.2. related to ^ - philosophically, what part of the code should be in charge of writing rejection reasons? Should we write at most one rejection reason per release x filter pair? It seems like it could be potentially confusing if any given helper can write its own rejection reason.

Currently all the "touched/active" filters can add rejections and I still think it's a good idea so the user will get multiple rejections in a single filter for every thing that the release did not match.

> 2.3. nit: If the return value is a bool I would probably name it something other than "check". To me it's less clear that "check" true means that we passed, than if the function were something like "size okay." "check = true" could mean "we performed the check" without indicating whether or not we passed the check. 🤷

Fair point, might change that to be more clear.

> 3. Nit: you could consider adding a test for the bug you fixed.

Might be a good idea indeed. It only happened when there was multiple External Filters so it was not obvious to me at first.

s0up4200:
@frrad in case you’re unaware, we’re all at discord if you’re ever up for a chat👍

frrad:
Thanks for the detailed explanation @zze0s ! I'll drop by discord and say hi :)

PR #1275, opened 2023-11-20
fix(filters): could not delete filter without actions

KaiserBh:
This fixes the no sql row set, no idea why it does that but removing the row check I added fixes it, Not sure if it affects others rowsAffected in the delete function.  I haven't come across it and it happened on postgresdb, probably gives same error in sqlite but I could be wrong, someone can just test the dev version and confirm it, if needed.

image

### DIFF
--- a/internal/database/action.go
+++ b/internal/database/action.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"database/sql"
 	"encoding/json"
+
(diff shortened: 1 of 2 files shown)


zze0s (approved):
Good find! It's only an issue when the filter does not have any actions, so should not be that big of an issue.

I looked at some other methods using the `RowsAffected` and those _should be fine_.

PR #1276, opened 2023-11-20
build(web): bump vite and cjs node api refactor

KaiserBh:
https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated

Some clean up and remove the warning for the deprecation, I tried to remove as many `any` uses and fix as many warning from the lint but I think there was 5 or 8 places that uses explicitly `any` that's need to be fixed.
There is some warning for `eslint-plugin-react-refresh`, which I added into the eslintrc. More info on react-refresh:
https://github.com/ArnaudBarre/eslint-plugin-react-refresh#readme


martylukyy (commented):
This PR reintroduces the source maps errors/warnings during building about which concerns were expressed..
ref https://github.com/autobrr/autobrr/pull/1274

After reading some of the references in the issues upstream it seems like it shouldn't cause any issues...
(Personally, i couldn't find any issues caused by this in the frontend either.)
Still, i would like to get your opinion about it! 🙂 

Would you just ignore them and send it? 
Apply a workaround?
Wait for the fix upstream?

KaiserBh:
Hmm I think as long it's not breaking anything it should be fine and by the looks of it everything is fine, but that being said let's just wait for a bit, no rush anyway. There is no point merging this if we are not using vite 5.0 since that's where the deprecate CJS Node API is at.

martylukyy:
After checking back with @KaiserBh and @stacksmash76 the source map warnings introduced by the vite 5.x.x upgrade,
shouldn't be a problem since the stacktraces artificially introduced in random components of autobrr are the same.
Or in other words - the content of the generated source maps did not change in the corresponding parts where an artificial exception was thrown after the upgrade.

A workaround has been applied to the vite.config.js to mute the warnings mentioned in https://github.com/autobrr/autobrr/pull/1274
as suggested in https://github.com/vitejs/vite/issues/15012#issuecomment-1815854072

The warnings are not actually introduced vite itself, but by the upstream dep `rollup`.
For more information see https://github.com/rollup/rollup/issues/4699

web/src/components/debug.tsx
```diff
@@ -23,5 +23,3 @@ const DEBUG: FC<DebugProps> = ({ values }) => {
     </div>
   );
 };
-
-export default DEBUG;
```
zze0s, line 27: Curious why switch from a default export for a file with a single export?

web/src/components/alerts/index.tsx
```diff
@@ -3,5 +3,5 @@
  * SPDX-License-Identifier: GPL-2.0-or-later
  */
 
-export * from "./ErrorPage";
-export * from "./Warning";
+export { ErrorPage } from './ErrorPage';
+export { WarningAlert } from "./Warning";
```
zze0s, line 7: Any benefit of doing it this way?

web/src/screens/settings/Feed.tsx
```diff
@@ -257,8 +261,13 @@ const FeedItemDropdown = ({
...
     },
-    onError: (error: any) => {
-      toast.custom((t) => <Toast type="error" body={`Failed to force run ${feed?.name}. Error: ${error.message}`} t={t} />, {
+    onError: (error: unknown) => {
+      let errorMessage = 'An unknown error occurred';
+      if (isErrorWithMessage(error)) {
+        errorMessage = error.message;
+      }
+
+      toast.custom((t) => <Toast type="error" body={`Failed to force run ${feed?.name}. Error: ${errorMessage}`} t={t} />, {
```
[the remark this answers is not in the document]
KaiserBh: > Is the `isErrorWithMessage` needed everywhere or only here?

Only needed there at the moment, I was trying to get rid of the `any` I think.

web/tailwind.config.ts
```diff
@@ -33,29 +34,20 @@ module.exports = {
...
-      // Pipe --tw-shadow-color (i.e. shadow-cyan-500/50) to our new text-shadow
-      // Credits: https://www.hyperui.dev/blog/text-shadow-with-tailwindcss
-      // Use it like: text-shadow shadow-cyan-500/50
-      matchUtilities(
-        { "text-shadow": (value) => ({ textShadow: value }) },
-        { values: theme("textShadow") }
-      );
-    }),
-  ],
-}
```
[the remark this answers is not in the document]
KaiserBh: Good question, I am not sure if we need it, but that being said I did search the files and found only 1 location that uses it and there is a note there by @stacksmash76 says:

```// NOTE(stacksmash76): added text-shadow only for the dark theme - light theme is fine contrast-wise when it comes to headings
// ideally, this would need a redesign``` 

It's up to you if you want to keep that I can add it back just let me know.

File: web/src/components/debug.tsx (hunk 1 above)
Comment (KaiserBh) [reply web/src/components/debug.tsx:27]: > Curious why switch from a default export for a file with a single export?

Due to react fast refresh check:
```
warning  Fast refresh only works when a file only exports components. Use a new file to share constants or functions between components
```

More info on: https://github.com/ArnaudBarre/eslint-plugin-react-refresh#readme 

Mainly these:
```
The plugin rely on naming conventions (i.e. use PascalCase for components, camelCase for util functions). This is why there are some limitations:

export * are not supported and will be reported as an error
Anonymous function are not supported (i.e export default function() {})
Class components are not supported
All-uppercase function export is considered an error when not using direct named export (ex const CMS = () => <></>; export { CMS })
```

File: web/src/components/alerts/index.tsx (hunk 2 above)
Comment (KaiserBh) [reply web/src/components/alerts/index.tsx:7]: Same as above reasoning.

re: https://github.com/ArnaudBarre/eslint-plugin-react-refresh?tab=readme-ov-file#limitations

File: web/tailwind.config.ts (hunk 4 above)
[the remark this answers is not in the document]
Comment (martylukyy) [reply web/tailwind.config.ts:61]: I took a look at this and it's used for every `CollapsibleSection` we have.
It adds a shadow to the heading of the section which is barely noticeable to me when not hovering over it.
When hovering over it adds a nice effect but doesn't add to readability in my opinion.
I would propose getting rid of it but if anyone has doubts about it, just react with a thumbs down please.

**Review by martylukyy — DISMISSED** [review]
LGTM.

Added the curly braces on every switch statement of the frontend and
removed the text-shadow property 
( ref https://github.com/autobrr/autobrr/pull/1276/files#r1427380538 )
As already said, if anyone has doubts about it, just react with a thumbs down please.

I left 3 further conversations about the changes for the fast refresh feature open,
containing explanations and reasonings.

EDIT:
I triple checked the curly braces I set,
still it would be nice if a second set of eyes could have another look over it.

File: web/src/screens/filters/sections/Actions.tsx [hunk 5]
```diff
@@ -134,44 +134,67 @@ const TypeForm = (props: ClientActionProps) => {
...
+  }
+  case "PORLA": {
     return <FilterActions.Porla {...props} />;
+  }
   // arrs
-  case "RADARR":
-  case "SONARR":
-  case "LIDARR":
-  case "WHISPARR":
-  case "READARR":
```
zze0s, line 157: Does ESLint complain about these? IMO the old style is better for this and other places doing the same type of multi case matching.

martylukyy: If you leave it as it was, it does not complain.
If you set the curly braces it complains with:
`TS7029: Fallthrough case in switch.` and
`ESLint: Expected a 'break' statement before 'case'.(no-fallthrough)`

zze0s: Then I'd suggest to revert these back to the fall-through style.

zze0s (approved):

PR #1277, opened 2023-11-21
feat(web): migrate Tanstack Query to v5

KaiserBh:
Migrate from React Query v4 to React Query v5.

## Key Changes

- Updated @tanstack/react-query to version 5.
- Replaced the keepPreviousData flag with the placeholderData function to maintain the previous query's data.
- Adjusted query hooks to align with the new React Query v5 API.
- Removed useErrorBoundary and suspense options from QueryClient, aligning with v5's changes.
- [Callbacks on useQuery (and QueryObserver) have been removed](https://tanstack.com/query/v5/docs/react/guides/migrating-to-v5#callbacks-on-usequery-and-queryobserver-have-been-removed)

More info on https://tanstack.com/query/v5/docs/react/guides/migrating-to-v5#breaking-changes


web/src/App.tsx
```diff
@@ -21,8 +21,7 @@ const queryClient = new QueryClient({
       // See https://tanstack.com/query/v4/docs/guides/query-retries#retry-delay
       // delay = Math.min(1000 * 2 ** attemptIndex, 30000)
       retry: true,
-      useErrorBoundary: true,
-      suspense: true
```
martylukyy, line 25: Any reason this got yeeted?

KaiserBh: > Any reason this got yeeted?

```To make the useErrorBoundary option more framework-agnostic and avoid confusion with the established React function prefix "use" for hooks and the "ErrorBoundary" component name, it has been renamed to throwOnError to more accurately reflect its functionality.```

https://tanstack.com/query/v5/docs/react/guides/migrating-to-v5#the-useerrorboundary-option-has-been-renamed-to-throwonerror

and 

```With v5, suspense for data fetching finally becomes "stable". We've added dedicated useSuspenseQuery, useSuspenseInfiniteQuery and useSuspenseQueries hooks. With these hooks, data will never be potentially undefined on type level:```

```js
const { data: post } = useSuspenseQuery({
  // ^? const post: Post
  queryKey: ['post', postId],
  queryFn: () => fetchPost(postId),
})
```

`The experimental suspense: boolean flag on the query hooks has been removed.`

https://tanstack.com/query/v5/docs/react/guides/migrating-to-v5#new-hooks-for-suspense

KaiserBh: That's all good, I think we just need to use the suspense query for those:


```ts 
const { data: post } = useSuspenseQuery({ // ^? const post: Post queryKey: ['post', postId], queryFn: () => fetchPost(postId), })
```

More info here: 
https://tanstack.com/query/latest/docs/react/guides/suspense

When I am home, I will try and fix that. That being said, do we know where we are using the react suspense spinner?

martylukyy: I don't exactly know every place we use it, but we are definitely using it:
On the Dashboard, Filters and Releases page (i think others too but those are too fast to load / see routes.tsx)
Inside the filters when loading the filter tabs and
On the Indexers, IRC, Feeds, Clients, Notifications, API keys pages of the settings.
(same here - i think others too but those are too fast to load / see Settings.tsx)

KaiserBh: @martylukyy just pushed a change, added it where I think it's needed can you test it? I don't have many releases etc so can't really tell if it's working or not.

KaiserBh: > I added some more useSuspenseQuery instead of useQuery to reinact the previous version. Please check if the removal of keepPreviousData on ReleaseTable.tsx could have any repercussions. 🤔 In case you have some spare time ofc 🙂

Oh thank you for that, I think regarding the keepPreviousData I believe @zze0s added them to some of them so it's best to ask him about that and also maybe have a look at this, when I changed the keepPreviousData I was just trying to keep it 1:1 to what it was before.

https://tanstack.com/query/v5/docs/react/guides/migrating-to-v5#removed-keeppreviousdata-in-favor-of-placeholderdata-identity-function

martylukyy: Yes the spinner works now on the various pages!

If we want to use the spinner on the ReleaseTable though we need to get rid of `placeholderData: keepPreviousData` since it doesn't go with `useSuspenseQuery` afaict.

Tagging @stacksmash76 for his opinion about the removal of `placeholderData: keepPreviousData`

zze0s: Not really sure why it's there or if it's needed anymore, seems to work fine with it disabled even before all these changes.

zze0s (dismissed):
Well done! 🚀

zze0s (approved):

PR #1278, opened 2023-11-21
feat(filters): auto-expand used sections

s0up4200:
We now collapse any section that doesn't contain data, instead of having only some sections open by default.
This should improve UX by focusing attention on sections that are actively filled out.


zze0s (approved):
LGTM!

PR #1281, opened 2023-11-25
feat(web): link Dashboard stats to Releases page

s0up4200:
Stats are now clickable and leads to a filtered release table based on the clicked stat item.

- Added `Errored Pushes` to Stats in Dashboard
- Placed `Links` on a separate row in Release Table

CleanShot 2023-12-01 at 17 37 58
dashboard_new
releasetable


zze0s:
Good feature. Might want to add a button/anchor to make it more apparent the user can click it? Something like this:

image


That's using the Link icon from the iconset currently in use. Copy pasted quickly so should probably not have a background, or a darker one :smile:

stacksmash76:
if the entire card is clickable then i'd make it upscale on hover and add a bigger transitioning shadow

zze0s (dismissed):

zze0s:
Everyone assigned as reviewer has had plenty of time to review. Will merge this shortly.

KyleSanderson (dismissed):

martylukyy (approved):
LGTM! 🚀 

Pushed a small change where I adapted the gap class for mobile view to make the margin to the screen border and between the items the same.

zze0s (approved):

PR #1282, opened 2023-11-25
feat(releases): incognito mode

s0up4200:
When the button is clicked, the current releases in the table are replaced with random Linux ISOs, Porla style. I made sure to make the button as discrete as possible, but up for changes.

Was very hesitant about this one, but decided to push it and let you guys have your say. 😄

incognito


web/src/utils/Helpers.ts
zze0s, line 1: Could move into the existing `utils/index.ts`.

zze0s (commented):
Fun feature!

Got any screenshots on the Releases page? Curious about the div changes of the pagination buttons.

web/src/utils/Helpers.ts
s0up4200: CleanShot 2023-12-14 at 22 02 10@2x

zze0s (approved):

PR #1298, opened 2023-12-04
feat(http): implement proper BaseUrl support to coexist with legacy mode

### LINKED ISSUE #1166: Not all paths are rewritten to the baseUrl
**Version**
1.31.0

**Describe the bug**
Not all paths are rewritten to the baseUrl (from config) which is set to `baseUrl = "/autobrr/"`. In my case it 404's `Inter-Variable.woff2` (noticed from Chrome->Console) which leads to blank page when accessing `https://subdomain.domain.tld/autobrr`

**To Reproduce**
docker-compose autobrr traefik labels
```
    labels:
      - "traefik.http.routers.autobrr.rule=PathPrefix(`/autobrr`)"
      - "traefik.http.routers.autobrr.entrypoints=web"
      - "traefik.http.services.autobrr.loadbalancer.server.port=7474"
      - "traefik.http.routers.autobrr.middlewares=autobrr-stripprefix"
      - "traefik.http.middlewares.autobrr-stripprefix.stripprefix.prefixes=/autobrr"
```

Set baseUrl in config to "/autobrr/"

**Expected behavior**
Display autobrr interface

Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/117.0.0.0 Safari/537.36

kaiserbh:
Hopefully closes #1166, I am sure this is a breaking change so do test it thoroughly.
By breaking change I mean maybe proxy level, people might have to update their proxy configuration. 

* This fix also addresses the previously required rewrite on the proxy level for baseUrl.

I only tested it with caddy with below config
```prolog
localhost /autobrr {
    reverse_proxy :7474
}
```
sse seem to work fine and not seeing any errors. 

Anyway do test it.


internal/http/auth.go
```diff
@@ -62,7 +62,7 @@ func (h authHandler) login(w http.ResponseWriter, r *http.Request) {
 
 	h.cookieStore.Options.HttpOnly = true
 	h.cookieStore.Options.SameSite = http.SameSiteLaxMode
-	h.cookieStore.Options.Path = h.config.BaseURL
```
brettpetch, line 65: I don't think we should be setting `cookieStore` to /. A significant amount of our users host on baseurls and this will allow other applications on the machine to read the cookie.

internal/http/auth.go
```diff
@@ -62,7 +62,7 @@ func (h authHandler) login(w http.ResponseWriter, r *http.Request) {
...
+	h.cookieStore.Options.Path = "/"
```
brettpetch, line 65: I don't think we should be setting cookieStore to /. A significant amount of our users host on baseurls and this will allow other applications on the machine to read the cookie.

kaiserbh:
> Can we get this going again, as it seems pretty much broken at the moment?
> 
> @kaiserbh still interested in getting this done? Thanks!

I think this is done,  but it's breaking change. So it won't be released until autobrr hit ver 2.0, probably.

elsbrock:
I tried your change and noticed that when configuring `baseUrl = "/autobrr/"` (with trailing slash) it doesn't work on first load (see [HAR](https://github.com/user-attachments/files/15753135/localhost_Archive.24-06-09.17-44-47.json)).

Without trailing slash it works when opening `/autobrr` (but not with trailing slash), but API calls are done via `/api` (see 
[HAR](https://github.com/user-attachments/files/15753136/localhost_Archive.24-06-09.17-44-22.json)). I'd have expected them to use `baseUrl` too.

Can somebody confirm, should that be fixed?

zze0s:
> I tried your change and noticed that when configuring `baseUrl = "/autobrr/"` (with trailing slash) it doesn't work on first load (see [HAR](https://github.com/user-attachments/files/15753135/localhost_Archive.24-06-09.17-44-47.json)).
> 
> Without trailing slash it works when opening `/autobrr` (but not with trailing slash), but API calls are done via `/api` (see [HAR](https://github.com/user-attachments/files/15753136/localhost_Archive.24-06-09.17-44-22.json)). I'd have expected them to use `baseUrl` too.
> 
> Can somebody confirm, should that be fixed?

I'd advice you to just run our regular builds and configure your reverse proxy correctly to do uri rewrite. We have docs for nginx and some others.

We are unfortunately beyond the point where we can just change this without breaking probably thousands of autobrr instances running behind a RP with baseurl.

The way baseurl works in autobrr today is flawed and not like most other applications, and it is a bit frustrating.

In the future maybe we'll decide to do a lot of breaking changes that warrants a proper v2 but until then, use a proxy rewrite because it works.

zze0s (commented):
Did a bunch of refactoring on this to properly support base url without having to use a reverse proxy to do url rewrite while also working with existing setups.

There is a new config variable `baseUrlModeLegacy` which by default is enabled/true so no existing setups break.

New setups can however disable it by setting it to `false` and base url set. This will make the api and web accessible on `/baseurl/` and `/baseurl/api` etc with no need for url rewrite on a proxy.

I have tested on existing baseurl setups as well as old and without base url and they all work. There is however a warning in web console where it can't fetch a font but it looks to still fetch that in another call.

Still need to do some cleanup of the code.

zze0s (approved):
This should be ready now! :partying_face: 

Tested on:
- swizzin with nginx and baseurl, no changes needed
- traefik with and without baseurl with legacy mode on and off
- locally with built binary
- locally with pnpm dev

s0up4200 (approved):
Tested locally with built binary 👍

luckylittle:
Just FYI: On `lighttpd-1.4.67-1.el9.x86_64` + Autobrr v`1.59.0` I've got it working with the following configs:

```
# config.toml
host = "127.0.0.1"
port = 7474
baseUrl = "/autobrr/"
baseUrlModeLegacy = false
```

```
# proxy.conf
server.modules += ( "mod_proxy" )
$HTTP["url"] =^ "/autobrr" {
  proxy.server = ( "" => (( "host" => "127.0.0.1", "port" => 7474 )))
  proxy.header = ( "upgrade" => 1 )
}
```

brettpetch:
> Just FYI: On `lighttpd-1.4.67-1.el9.x86_64` + Autobrr v`1.59.0` I've got it working with the following configs:
> 
> 
> 
> ```
> 
> # config.toml
> 
> host = "127.0.0.1"
> 
> port = 7474
> 
> baseUrl = "/autobrr/"
> 
> baseUrlModeLegacy = false
> 
> ```
> 
> 
> 
> ```
> 
> # proxy.conf
> 
> server.modules += ( "mod_proxy" )
> 
> $HTTP["url"] =^ "/autobrr" {
> 
>   proxy.server = ( "" => (( "host" => "127.0.0.1", "port" => 7474 )))
> 
>   proxy.header = ( "upgrade" => 1 )
> 
> }
> 
> ```

This was already documented. Locking this thread. 

https://autobrr.com/installation/reverse-proxy/lighttpd

PR #1304, opened 2023-12-14
feat(filters): show enabled and disabled actions in list view

s0up4200:
## New web:
CleanShot 2023-12-14 at 23 04 37@2x
CleanShot 2023-12-14 at 23 04 25@2x

## Further improvements that can be done:

- [x] Make sure that the release table arent populated if no actions are enabled, or possibly show an error.
Right now i think it only looks for `actions_count` before processing.

CleanShot 2023-12-14 at 23 12 00@2x


internal/release/service.go
```diff
@@ -172,19 +189,6 @@ func (s *service) processFilters(ctx context.Context, filters []*domain.Filter,
...
-		actions, err := s.actionSvc.FindByFilterID(ctx, f.ID)
-		if err != nil {
-			s.log.Error().Err(err).Msgf("release.Process: error finding actions for filter: %s", f.Name)
-			return err
-		}
-
-		// if no actions, continue to next filter
-		if len(actions) == 0 {
-			s.log.Warn().Msgf("release.Process: no actions found for filter '%s', trying next one..", f.Name)
-			return nil
```
s0up4200, line 185: This has to be a bug right? I noticed with return nil, the log never shows the processing of the next filter in line, but if we continue instead, then it does.

s0up4200, line 185: The `return nil`has to be a bug, right? I noticed with return nil, the log never shows the processing of the next filter in line, but if we continue instead, then it does.

internal/release/service.go
```diff
@@ -162,27 +162,21 @@ func (s *service) processFilters(ctx context.Context, filters []*domain.Filter,
...
 		}
 
-		// if no actions, continue to next filter
-		if len(actions) == 0 {
-			s.log.Warn().Msgf("release.Process: no actions found for filter '%s', trying next one..", f.Name)
-			return nil
+		// if no actions, or all actions are disabled, continue to next filter
+		if len(actions) == 0 || !s.hasEnabledAction(actions) {
+			msg := "no actions found"
+			if len(actions) > 0 {
```
zze0s, line 175: I think it would be a better option to add a new optional parameter to the repo method `FindByFilterID(ctx context.Context, filterID int, active bool) ([]*Action, error)` which is called by `actionSvc.FindByFilterID`.

Then keep this as before with the `len(actions) == 0` check and change the message to `release.Process: no active actions found for filter '%s', trying next one..`.

internal/release/service.go
```diff
@@ -192,6 +186,16 @@ func (s *service) processFilters(ctx context.Context, filters []*domain.Filter,
...
+		// save release here to only save those with rejections from actions instead of all releases
+		if release.ID == 0 {
+			release.FilterStatus = domain.ReleaseStatusFilterApproved
+
+			if err = s.Store(ctx, release); err != nil {
+				l.Error().Err(err).Msgf("release.Process: error writing release to database: %+v", release)
+				return err
+			}
+		}
+
```
zze0s, line 198: This fixes that bug with empty actions, nice!

internal/release/service.go
s0up4200: Done 👍

internal/release/service.go
```diff
@@ -162,27 +162,18 @@ func (s *service) processFilters(ctx context.Context, filters []*domain.Filter,
...
 			s.log.Error().Err(err).Msgf("release.Process: error finding actions for filter: %s", f.Name)
 			return err
 		}
 
 		// if no actions, continue to next filter
 		if len(actions) == 0 {
-			s.log.Warn().Msgf("release.Process: no actions found for filter '%s', trying next one..", f.Name)
-			return nil
+			s.log.Warn().Msgf("release.Process: no active actions found for filter '%s', trying next one..", f.Name)
+			continue
```
s0up4200, line 176: @zze0s Do you agree that we should `continue` here? We returned `nil` before, which seems wrong as it does not log the next filter check that way. (It does process it, but no logs from what I can tell)

zze0s: Correct, continue will make it check the next filter then which is the intended way.

KyleSanderson (approved):

zze0s (approved):

PR #1305, opened 2023-12-15
feat(database): implement backups 

KyleSanderson:
Lots of love needed, but based on the fact there's now 2k+ consumers this is becoming quite critical.


internal/database/backup.go
```diff
@@ -0,0 +1,131 @@
...
+			retain++
+		}
+
+		return cleanupDatabase(base, db, retain)
+	}
+
+	return errors.New("backup: not implemented for database type: %s", db.Driver)
+}
+
+func databaseConsistent(tx *sql.Tx) error {
```
zze0s, line 68: Should probably make this clear it's for SQLite only, so rename it to `databaseConsistentSQLite` might be a good idea.

internal/database/sqlite.go
```diff
@@ -78,6 +78,7 @@ func (db *DB) openSQLite() error {
 		return err
 	}
 
+	db.BackupDatabase(false)
```
zze0s, line 81: Do we need to run this on both startup and stop?

KyleSanderson: Yeah

martylukyy:
Superseded by https://github.com/autobrr/autobrr/pull/1756

PR #1306, opened 2023-12-15
fix(http): specify ip stack when listen fails

KyleSanderson:
Linux capabilities can limit the usage of dual stack configurations through namespaces. It's rather silly, but there's been sporadic reports over the years about this - so try for the best, downgrade to v4, then v6.

### DIFF
--- a/internal/http/server.go
+++ b/internal/http/server.go
@@ -75,17 +75,31 @@ func NewServer(log logger.Logger, config *config.AppConfig, sse *sse.Server, db 
 
 func (s Server) Open() error {
 	addr := fmt.Sprintf("%v:%v", s.config.Config.Host, s.config.Config.Port)
-	listener, err := net.Listen("tcp", addr)
+
+	var err error
+	for _, proto := range []string{"tcp", "tcp4", "tcp6"} {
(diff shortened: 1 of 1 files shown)


internal/http/server.go
```diff
@@ -75,18 +75,32 @@ func NewServer(log logger.Logger, config *config.AppConfig, sse *sse.Server, db
 
 func (s Server) Open() error {
 	addr := fmt.Sprintf("%v:%v", s.config.Config.Host, s.config.Config.Port)
-	listener, err := net.Listen("tcp", addr)
+
+	var err error
+	for _, proto := range []string{"tcp", "tcp4", "tcp6"} {
+		if err = s.tryToServe(addr, proto); err == nil {
+			return err
+		}
```
zze0s, line 83: Shouldn't this be `; err != nil` so it can return the err, or have it return nil?

KyleSanderson: most certainly.

KyleSanderson: actually... no, that was intentional. we can force it to nil, just an old habit.

zze0s (approved):

PR #1309, opened 2023-12-16
fix(actions): qBit handle `Ignore slow torrents` rule correctly when disabled

s0up4200:
We did not check for the enabled state of the rule it seems. This should solve that.
Ref: https://discord.com/channels/881212911849209957/881967548143403058/1185543801494843452

### DIFF
--- a/internal/action/qbittorrent.go
+++ b/internal/action/qbittorrent.go
@@ -167,14 +167,14 @@ func (s *service) qbittorrentCheckRulesCanDownload(ctx context.Context, action *
 				return []string{rejection}, nil
 			}
 
-			if rules.IgnoreSlowTorrentsCondition == domain.IgnoreSlowTorrentsModeMaxReached {
+			if rules.IgnoreSlowTorrents && rules.IgnoreSlowTorrentsCondition == domain.IgnoreSlowTorrentsModeMaxReached {
 				// get transfer info
 				info, err := qbt.GetTransferInfoCtx(ctx)
 				if err != nil {
 					return nil, errors.Wrap(err, "could not get transfer info")
 				}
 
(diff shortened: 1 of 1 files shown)


internal/action/qbittorrent.go
```diff
@@ -174,7 +174,7 @@ func (s *service) qbittorrentCheckRulesCanDownload(ctx context.Context, action *
 					return nil, errors.Wrap(err, "could not get transfer info")
 				}
 
-				rejections := s.qbittorrentCheckIgnoreSlow(rules.DownloadSpeedThreshold, rules.UploadSpeedThreshold, info)
+				rejections := s.qbittorrentCheckIgnoreSlow(rules.IgnoreSlowTorrents, rules.DownloadSpeedThreshold, rules.UploadSpeedThreshold, info)
```
zze0s, line 177: Probably better to have check bool check in the if at line 170

```golang
if rules.IgnoreSlowTorrents && rules.IgnoreSlowTorrentsCondition == domain.IgnoreSlowTorrentsModeMaxReached {
```

internal/action/qbittorrent.go
```diff
@@ -196,7 +196,7 @@ func (s *service) qbittorrentCheckRulesCanDownload(ctx context.Context, action *
 			return nil, errors.Wrap(err, "could not get transfer info")
 		}
 
-		rejections := s.qbittorrentCheckIgnoreSlow(rules.DownloadSpeedThreshold, rules.UploadSpeedThreshold, info)
+		rejections := s.qbittorrentCheckIgnoreSlow(rules.IgnoreSlowTorrents, rules.DownloadSpeedThreshold, rules.UploadSpeedThreshold, info)
```
zze0s, line 199: Same as above but for line 192.

zze0s (approved):

PR #1313, opened 2023-12-20
enhancement(feeds): change max age default value to 0

### LINKED ISSUE #1280: [FEATURE REQ] Generic RSS default "Max Age" is too low; Log that a release is dropped due to old age
Hi, I set up Autobrr a few months ago but I couldn't get it to actually fetch things from two of my RSS feeds. I saw that the RSS feeds are downloaded correctly, and my only filter was just the one with all empty fields.

The culprit turned out to be the "Max Age" setting being too strict by default.
You can find it in "Settings > Feeds > (press 3 dots near one of the RSS feeds) > Edit".

The default value of 3600 (1 hour) is too low, because consider this case:
1. A torrent is automatically published at 18:00
2. The tracker is scraped at 21:00 and a RSS feed is generated.

The result is an RSS feed that has an entry that was published at 18:00, but .rss feed with that very entry could've been fetched at the earliest by 21:00, so the "Max age" will drop it.

Here's even a real life case: https://archlinux.org/feeds/releases/
At the top you can see that the RSS feed was added at:
Last updated: Wednesday, November 1, 2023 8:06:32 AM

but the newest entry is from Wednesday, November 1, 2023 1:00:00 AM, so 7 hours earlier.

----

But perhaps that's an one off case, Still, it'd be nice if the logs let the user know, as it's unexpected that feeds are being filtered outside the "Filters". 

Currently the logs look like this, given the case above:
```
023-11-24 17:50:08.868396+01:00{"level":"debug","module":"feed","feed":"Generic RSS","time":"2023-11-24T17:50:08+01:00","message":"found (0) new items to process"}
2023-11-24 18:20:07.325983+01:00{"level":"debug","module":"feed","feed":"archlinux","time":"2023-11-24T18:20:07+01:00","message":"refreshing rss feed: archlinux, found (3) items"}
```

I was wondering if it'd be possible to make Autobrr work like this:
1. Download every 30m like it always does
2. Compare the new RSS against its current RSS feed cache (does something like this exist?)
3. If there are new elements, see if they are older than maxAge. If they are, **LOG** "Found new release but it's older than maxAge, so dropping". If they're not, just accept them as normal.

1. Consider increasing default RSS maxAge to account for differences between "Publish date of entry as seen in RSS text" and "Date at which the RSS's .xml file actually becomes available on the Internet"
2. Consider logging when a new RSS entry gets dropped due to max age

-

**Additional context**
-

KaiserBh:
Updated help text.

image

Partially closes #1280.

### DIFF
--- a/internal/database/postgres_migrate.go
+++ b/internal/database/postgres_migrate.go
@@ -342,7 +342,7 @@ CREATE TABLE feed
 	url           TEXT,
 	interval      INTEGER,
 	timeout       INTEGER DEFAULT 60,
-	max_age       INTEGER DEFAULT 3600,
+	max_age       INTEGER DEFAULT 0,
 	categories    TEXT []   DEFAULT '{}' NOT NULL,
 	capabilities  TEXT []   DEFAULT '{}' NOT NULL,
 	api_key       TEXT,
@@ -831,5 +831,8 @@ ALTER TABLE filter_external
 `,
 	`ALTER TABLE irc_network
 		ADD COLUMN bot_mode BOOLEAN DEFAULT FALSE;
+`,
+	`ALTER TABLE feed
+	ALTER COLUMN max_age SET DEFAULT 0;
 `,
 }
--- a/internal/database/sqlite_migrate.go
+++ b/internal/database/sqlite_migrate.go
@@ -1425,30 +1425,55 @@ ALTER TABLE filter_external_dg_tmp
 	`ALTER TABLE irc_network
 	ADD COLUMN bot_mode BOOLEAN DEFAULT FALSE;
 `,
-	`ALTER TABLE feed RENAME TO old_feed;
-	CREATE TABLE feed (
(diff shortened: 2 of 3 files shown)


s0up4200:
I know z has thought about moving the max age setting into filters. I think that sounds like a good idea too. It would make the setting more visible to the user.

I don't really understand why it needs to default to 3600 if we recommend 0 though.

martylukyy:
I agree with soup here.
Setting it to 0 by default and letting the user decide the max age should be seems like the better choice in my opinion.
We had quite some troubleshooting to do because of it...

KaiserBh:
> I know z has thought about moving the max age setting into filters. I think that sounds like a good idea too. It would make the setting more visible to the user.
> 
> I don't really understand why it needs to default to 3600 if we recommend 0 though.

Yea I thought of the same thing, well changed the default to 0 for now.

zze0s (approved):

PR #1314, opened 2023-12-20
fix(web): improve tooltip responsiveness and click-through

### LINKED ISSUE #1194: bug(web): tooltip broken on mobile, click goes through to input element below tooltip
Can be reproduced on a real mobile on the filter details page.

s0up4200:
Closes #1194 

Tested and working as expected on iPad and iPhone with Safari. Someone should test on Android.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -33,6 +33,7 @@
     "@headlessui/react": "^1.7.17",
     "@heroicons/react": "^2.0.18",
     "@hookform/error-message": "^2.0.1",
+    "@popperjs/core": "^2.11.8",
     "@tailwindcss/forms": "^0.5.7",
     "@tanstack/react-query": "^4.36.1",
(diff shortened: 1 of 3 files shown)


KaiserBh:
> Closes #1194
> 
> Tested and working as expected on iPad and iPhone with Safari. Someone should test on Android.

Does this also fixed the mobile version where if the tooltip shows it expands a lot so we need to scroll to see it? @s0up4200

s0up4200:
> Does this also fix the mobile version where if the tooltip shows it expands a lot so we need to scroll to see it? @s0up4200

Should be good now 👍 @KaiserBh 
Defaults to right, if too narrow then top.

KaiserBh (approved):
Tested it and seem to be working fine now. 
LGTM!

PR #1315, opened 2023-12-20
feat(actions): qBittorrent add priority handling

### LINKED ISSUE #1012: [FEATURE REQ] Add to top of queue in qbit
Need Brrrr faster

A option to add torrents to the top of the queue in qbit 

Means less brrrr

**Additional context**
Soup said it sounds good

s0up4200:
~~It *almost* works as you'd expect, but if user sets either `Top of queue` or `Bottom of queue` when `Torrent Queueing` is disabled, the release is flagged as an `Action error` in the release history. It gets added to the client regardless. This results in it breaking the `MAX_DOWNLOADS` rule if set. This should be an easy fix I think.~~

~~When `Torrent Queueing` is enabled, it works just fine from my testing.~~

Tested on 4.3.9.

closes #1012

### DIFF
--- a/internal/database/sqlite_migrate.go
+++ b/internal/database/sqlite_migrate.go
@@ -199,7 +199,7 @@ CREATE TABLE action
     limit_download_speed    INT,
     limit_ratio             REAL,
     limit_seed_time         INT,
-    action_priority         TEXT,
+    priority                TEXT,
     reannounce_skip         BOOLEAN DEFAULT false,
     reannounce_delete       BOOLEAN DEFAULT false,
     reannounce_interval     INTEGER DEFAULT 7,
@@ -1426,9 +1426,6 @@ ALTER TABLE filter_external_dg_tmp
 	`ALTER TABLE irc_network
 	ADD COLUMN bot_mode BOOLEAN DEFAULT FALSE;
 `,
-	`ALTER TABLE action
-    ADD COLUMN action_priority TEXT;
-`,
 	`CREATE TABLE feed_dg_tmp
 (
     id            INTEGER PRIMARY KEY,
@@ -1480,7 +1477,7 @@ DROP TABLE feed;
 ALTER TABLE feed_dg_tmp
     RENAME TO feed;
 `,
-  `ALTER TABLE action
-	ADD COLUMN action_priority TEXT;
+	`ALTER TABLE action
+	ADD COLUMN priority TEXT;
 `,
 }
--- a/internal/database/sqlite_migrate.go
+++ b/internal/database/sqlite_migrate.go
@@ -199,6 +199,7 @@ CREATE TABLE action
     limit_download_speed    INT,
     limit_ratio             REAL,
     limit_seed_time         INT,
+    action_priority         TEXT,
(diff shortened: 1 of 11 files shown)


KaiserBh:
Is there a way to prevent that to begin with? 
If we can check if `Torrent Queueing` is enabled or not? if it's allow the changes be made otherwise don't allow it and give them warning or error?

s0up4200:
> Is there a way to prevent that to begin with? If we can check if `Torrent Queueing` is enabled or not? if it's allow the changes be made otherwise don't allow it and give them warning or error?

Yeah, I missed that part yesterday. We can already check if that is enabled or not. Question then is; should autobrr be set up to enable it?

KaiserBh:
> > Is there a way to prevent that to begin with? If we can check if `Torrent Queueing` is enabled or not? if it's allow the changes be made otherwise don't allow it and give them warning or error?
> 
> Yeah, I missed that part yesterday. We can already check if that is enabled or not. Question then is; should autobrr be set up to enable it?

Good question, I personally think it would be nice if we enable it for them. But I think most people who uses torrent queueing feature have it enabled? ~~Maybe we should just give some sort of helper text in the rules~~ never mind I just checked you already did, nice well done.

image

Honestly just leave it as it is and see how we go. 

Well done that. 👍

KaiserBh (dismissed):
Good stuff!

KaiserBh (dismissed):

s0up4200:
https://github.com/autobrr/autobrr/pull/1315/commits/b4f7c3765fe0d2a5537d7be5e36767def04a10dc relies on autobrr/go-qbittorrent#21

If user sets priority settings, we will now check if torrent queueing is enabled. If disabled, we enable it.

internal/database/sqlite_migrate.go
```diff
@@ -1422,7 +1423,10 @@ ALTER TABLE filter_external_dg_tmp
 	`ALTER TABLE filter_external
 	DROP COLUMN webhook_retry_max_jitter_seconds;
 `,
-  `ALTER TABLE irc_network
+	`ALTER TABLE irc_network
 	ADD COLUMN bot_mode BOOLEAN DEFAULT FALSE;
 	`,
+	`ALTER TABLE action
+    ADD COLUMN action_priority TEXT;
```
s0up4200, line 1430: Probably not the best name for it? 🤔

zze0s: Maybe just priority is enough since the table is action already. Becomes like this otherwise `actions.actions_priority`.

zze0s (dismissed):
Looks good!

Will the queuing not work without it being set globally in qBittorrent? Maybe not ideal to do the external check each time, but it's a minor thing.

I merged the PR in go-qbittorrent and tagged a new release `v1.8.0`.

internal/database/sqlite_migrate.go
s0up4200: It might be me overlooking something, but I had it as `priority` initially, but then it errored on expecting int64 somewhere, probably due to `filters.priority`. I can try again later tonight though!

zze0s (approved):
Well done! :1st_place_medal:

PR #1320, opened 2023-12-28
build(ci): enable architecture optimizations for tests

KyleSanderson:
https://go-review.googlesource.com/c/go/+/444396


zze0s:
Changing this seem to slow down the tests about 10 seconds with the current codebase. Could be due to the Hosted Runners running older hardware that does not support AVX, but not certain.

Maybe will be different in the future.

PR #1322, opened 2023-12-29
feat(irc): allow lazy announcer nicks

s0up4200:
Given how some indexers are “lazy” in the lack of a better word with NickServ identifying their announcers, this seems like a good addition.

https://github.com/autobrr/autobrr/pull/1322/commits/d27ef460387d4ebfddc079c083421abf02562d05 simplifies this by making it support a suffixed digit all the time. This commit reverts all the previous commits.

<details>
  <summary>Outdated</summary>
CleanShot 2023-12-29 at 22 53 11@2x
</details>


Flupster:
Considering the variability in bot nickname postfixes, it might be impractical to assume that a digit will be appended for the IRC nick. From my observation a `_` postfix appears to be more prevalent. I suggest the following approach:

```GO
// Verify if the nickname concludes with an asterisk and holds the correct prefix
if strings.HasSuffix(nick, "*") && strings.HasPrefix(nick, announcer) {
    return true
}

// Confirm if the nickname starts with the announcer and comprises one additional character
if strings.HasPrefix(nick, announcer) && len(nick) == len(announcer)+1 {
    return true
}
```

Since IRC nicknames typically do not permit asterisks, this adjustment accommodates cases where an announcer ends with an asterisk in its definition. Additionally, I have omitted the requirement for the last character to be a digit, which was the original issue I anticipated.

zze0s (commented):
If we want to allow this I think we should go over all definitions and hardcode the URLs so that the worst thing that can happen is someone just announcing old ids on the same domain. Currently most of our definitions just do a loose addr match and tries to find things like `id=(.+)` on the end.

zze0s (approved):
I think this will help more than cause issues. There are some bad announce channels where everyone can post, but we build the urls with the baseurl set in indexer settings so it will only try the correct site in that case.

PR #1329, opened 2023-12-30
enhancement(web): force run feed with enter key

martylukyy:
This PR enables the Force Run modal to be confirmed with the ENTER key instead of clicking the Force Run button


zze0s (approved):

s0up4200 (approved):
Works here 👍 Tested on Firefox Desktop and Safari iOS.

PR #1311, opened 2023-12-17
fix(auth): implement invalid cookie handling

KyleSanderson:
https://github.com/autobrr/autobrr/pull/1310

PR #1338, opened 2024-01-01
feat(web): move from react-router to @tanstack/router

### LINKED ISSUE #941: Auth issues / login issues
**Version**
v1.26.1

**Describe the bug**
So I did some testing with the auth issues we have been having lately.
Right now, there are three bugs with our login page i think.

First, if the basic auth of the webserver becomes invalid, users will get a toast with `Wrong password or username!`
when trying to log in, even though they have entered the correct credentials. 
This behaviour will persist until a hard refresh of the login page which results in the user getting a browser prompt for the basic auth.
After providing the correct credentials for the basic auth you can log into autobrr again.

So you might ask yourself right now: Why should the basic auth become invalid?
Let me introduce you to the second issue. 
When entering wrong credentials, the basic auth will become invalid, but the user won't notice that until a hard refresh 
because you are not getting a browser prompt for the basic auth if you don't hard refresh the page. 
So you are basically stuck on the login page without any hint of you basic auth being invalid.

Third, if the session secret in the config.toml or the session token of the cookie is not valid for the current session, 
it will throw an recoverable error while it should theoretically just put the user back to the login page.

The first two issues should have a higher priority, since it has the potential of getting the user stuck. 
On the third issue the user can just click the `Reset page state` button and the you will be able to login as intended.

martylukyy:
This PR fixes our login issues when an invalid `user_session` cookie was present.

The logic behind it is a bit fucked and needs some context so i will try my best to clarify:
On line 89 we try to get the `user_session` cookie which is likely to fail due to 2 reasons:
- There is no  `user_session` cookie
- There is an invalid  `user_session` cookie.

Regardless of what the error would be:
We ignore the error and save a new session since we check for bad credentials a little bit further up,
in order to set a new and valid `user_session` cookie.

The error handling from line 92 to 95 could maybe be improved? Idk.

ref https://github.com/gorilla/sessions/issues/16#issuecomment-143642144

---

Additionally this PR changes the HTTP status code when providing bad credentials for an existing user.

HTTP status code 401 is not an appropriate status code here even if it might seems so. Why?
When an application renders a page with status code 401, 
any webserver with enabled basic authentication is forced to log you out.

The most logical status code after 401 was 403 for me - feel free to suggest other codes though.

ref https://github.com/zammad/zammad/issues/2983

---

Yes, onboarding still works.

---

I tested this pretty extensively i would say - but still:
I would appreciate if some brains with more convolutions than mine could take an additional look at this PR. 😄 

closes #941


martylukyy:
This PR and #1311 fix most parts of #941 

There is still something wrong when the session secret changes where it throws an unrecoverable error.
After resetting the page state once though, autobrr continues it's work.

```
Error: [401] Unauthorized: "api/config"
    at fc (APIClient.ts:95:27)
```

martylukyy:
As internally discussed i went ahead and set the `throwOnError` property of the query client to `false`.
This means that if **any** of the queries fail to fetch the error will not be thrown anymore.
While this feels wrong and more like a bandaid solution it provides the best UX.
However, i just want to point out that the underlying problem - 
that it requests things that it shouldn't
 - is still there.

zze0s (dismissed):

martylukyy:
Something is still majorly fucked... 😅 

When logged in change the session secret in the config.toml and restart autobrr.

autobrr will now run into a "loop" where it won't log you out, let's you navigate to other pages, but the queries fail.
This seems to be due to the fact we changed line 36 in middleware.go to status code 500,
but we don't `AuthContext.reset()` for 500 hence autobrr still thinking we are authenticated iiuc.

Removing the auth context manually from localstorage, redirects you to the login page but:
`securecookie: the value is not valid` is back and upon providing correct credentials 
the user won't know he has to manually clear the cookie in order to login again.

martylukyy (commented):
When logged in, change the value of the session cookie to something invalid or
the expiration date of the session cookie to something before the current time.

autobrr will now delete the session cookie correctly but will not redirect to login.

Instead you get an "empty" UI and the infamous
`data?.filter is not a function` for the filters page or
`data?.map is not a function` for the releases page.
I suspect the latter option with the expiration date is what users have been hitting?

Manually logging out and in again fixes the filters page but the releases page needs a good ol' F5.

I think this problem occurs due to the `autobrr_user_auth` key in localstorage not changing it's `isLoggedIn` value to `false` hence it not redirecting for login and so on and so forth?

web/src/utils/Context.ts
stacksmash76, line 1: ContextMerger should (ideally) deep-copy. Oversight on my end when I built it.

```
function ContextMerger<T extends {}>(
  key: string,
  defaults: T,
  ctxState: StateWithValue<T>
) {
  let values = structuredClone(defaults);

  const storage = localStorage.getItem(key);
  if (storage) {
    try {
      const json = JSON.parse(storage);
      if (json === null) {
        console.warn(`JSON localStorage value for '${key}' context state is null`);
      } else {
        values = { ...values, ...json };
      }
    } catch (e) {
      console.error(`Failed to merge ${key} context state: ${e}`);
    }
  }
  
  ctxState.set(values);
}
```

zze0s: Ah, nice!

zze0s (approved):
This PR turned into a huge refactor of the whole routing layer to use `@tanstack/router` and remove `react-router`.

It plays really well with `@tanstack/query` and we also use `@tanstack/table` which will get an update to v8+ in the future.

We have tested these changes thoroughly with different setups and everything looks the be working as intended.

PR #1339, opened 2024-01-02
feat(filters): implement min and max uploader count filtering

luckyboy:
This PR implement feature for min seeders and max seeders filtering.
Seeders count filtering will improve share rate for torrents have less seeders.

### DIFF
--- a/internal/database/sqlite.go
+++ b/internal/database/sqlite.go
@@ -105,6 +105,7 @@ func (db *DB) migrateSQLite() error {
 		return errors.Wrap(err, "failed to query schema version")
 	}
 
+
 	if version == len(sqliteMigrations) {
 		return nil
 	} else if version > len(sqliteMigrations) {
--- a/web/src/screens/filters/sections/Advanced.tsx
+++ b/web/src/screens/filters/sections/Advanced.tsx
@@ -252,6 +252,28 @@ const Uploaders = ({ values }: ValueConsumer) => (
         </div>
       }
     />
+    <Input.NumberField
+      name="min_uploader_count"
+      label="Min Uploaders Count"
+      placeholder="Takes any number (0 is infinite)"
+      tooltip={
+        <div>
+          <p>Number of max seeders as specified by the respective unit.</p>
+          <DocsLink href="https://autobrr.com/filters#rules" />
+        </div>
+      }
+    />
+    <Input.NumberField
+      name="max_uploader_count"
+      label="Max Uploaders Count"
+      placeholder="Takes any number (0 is infinite)"
+      tooltip={
+        <div>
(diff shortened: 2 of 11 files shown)


internal/database/sqlite.go
```diff
@@ -105,6 +105,8 @@ func (db *DB) migrateSQLite() error {
 		return errors.Wrap(err, "failed to query schema version")
 	}
 
+	db.log.Info().Msgf("PRAGMA user_version %d", version)
```
KyleSanderson, line 108: snuck-in perhaps?

KyleSanderson (commented):
Neat! Doesn't look like freeleech time is wired up yet, but off to a very solid start.

internal/database/sqlite.go
luckyboy: Ha, it's indeed snuck-in.

luckyboy:
> Neat! Doesn't look like freeleech time is wired up yet, but off to a very solid start.

Freeleech time is defined by some site, i can't find a way to extract those information from current torznab implement. Maybe i should remove this part from this PR.

web/src/screens/filters/sections/Advanced.tsx
```diff
@@ -252,6 +252,28 @@ const Uploaders = ({ values }: ValueConsumer) => (
         </div>
       }
     />
+    <Input.NumberField
+      name="min_uploader_count"
+      label="Min Uploaders Count"
+      placeholder="Takes any number (0 is infinite)"
+      tooltip={
+        <div>
+          <p>Number of max seeders as specified by the respective unit.</p>
```
s0up4200, line 261:
```suggestion
          <p>Number of min seeders as specified by the respective unit.</p>
```

s0up4200 (changes requested):

zze0s (commented):
Good stuff!!

Some suggestions:

* I suggest changing `uploader` to `seeder` for db and structs
* Add min/max leecher filters like seeder already have
* Add a new filter tab/page called Feeds where these Feeds specific settings are grouped

We have recently discussed about moving the Feed Max Age setting from Feed settings to the filter. It probably makes more sense in the filter. I can take care of that in this or a new PR.

zze0s:
No need to close the PR. If you push to the same branch they will show up here.

luckyboy:
> No need to close the PR. If you push to the same branch they will show up here.

Thank you, i'm still adapting to github. i will try to improve this PR anyway.

luckyboy:
Created a new PR #1342, leechers filtering is included. And the filter settings for seeders and leechers ware arranged to RSS/Torznab/Newznab-specific(FeedSpecific).

PR #1340, opened 2024-01-02
build(ci): disable ghcr push from forks

KyleSanderson:


zze0s (commented):
~~Looks like this doesn't really work since it did not push or run the publish step.~~

I did not see this was from a fork. Looks like it's working as intended! :1st_place_medal:

zze0s (approved):

PR #1342, opened 2024-01-03
feat(filters): implement min and max seeders/leechers filtering for Torznab feeds

### LINKED ISSUE #815: [FEATURE REQ] Filter feeds by  `seeder` / `leecher` / `published_time`
I use the `freeleech` feature to build up buffers from PTs. However, I noticed that the torrents grabbed by autobrr are usually not the optimal ones. For example, it grabs torrents with 200 seeders and only 1 leecher, which results in almost zero uploads. At the same time, a new torrent with only 1 seeder but 200 leechers is neglected (simply because it ranks after the 200-seeder torrent in the feed). 

Introduce 3 new variables: `seeder`/`leecher`/`published_time` to `Filter > Advance > Freeleech`. 
The variables can be read from Jackett:
```
<item>
  <pubDate>Sat, 17 Dec 2022 01:00:17 -0800</pubDate>
  <torznab:attr name="seeders" value="302"/>
  <torznab:attr name="peers" value="302"/>
</item>
```
Add 3 filters on the variables:
1. `MAX SEEDER`: 10: Any torrent with more than 10 seeders is neglected.
2. `MIN LEECHER`: 100: Any torrent with less than 100 leechers is neglected.
3. `LATEST PUBLISH TIME`: 15 min: Any torrent with more than 15 minutes of published time is neglected.

Use 3rd party tools to monitor autobrr's downloads, [autoremove-torrents](https://github.com/jerrymakesjelly/autoremove-torrents) for example. The workflow will be like:
1. autobrr downloads a torrent $T_1$ and `max_download` is reached;
2. autoremove finds out A has seeder > 10, then delete $T_1$;
3. autobrr downloads a torrent $T_2$ since `max_download` is not reached;
4. Goto 2 until torrent  $T_n$ meets the requirement of autoremove.

However, this solution 
1. Needs A 3rd party tool monitoring the client (while the proposed solution is one shot);
2. Download bandwidth wasted for deleted download;
3. May trigger H&R rules due to unfinished downloads.

**Additional context**
N/A. Any thoughts and comments are welcome.

luckyboy:
This PR implement feature for min and max seeders/leechers filtering
Seeders/Leechers filtering will improve share rate for torrents.

closes #815

### DIFF
--- a/internal/database/filter.go
+++ b/internal/database/filter.go
@@ -239,12 +239,12 @@ func (r *FilterRepo) FindByID(ctx context.Context, filterID int) (*domain.Filter
 			"f.except_tags_match_logic",
 			"f.origins",
 			"f.except_origins",
-			"f.created_at",
-			"f.updated_at",
 			"f.min_seeders",
 			"f.max_seeders",
 			"f.min_leechers",
 			"f.max_leechers",
+			"f.created_at",
+			"f.updated_at",
 			"fe.id as external_id",
 			"fe.name",
 			"fe.idx",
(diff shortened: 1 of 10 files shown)


zze0s (approved):
Hey! Thanks for the PR! This is a great addition!

I did some formatting fixes and reordered some fields.

PR #1343, opened 2024-01-03
fix(database): postgresql release timestamps timezone handling

### LINKED ISSUE #1138: Release ages are offset by 3 hours in the web UI - Timezone postgres
**Version**
1.29.0

**Describe the bug**
In the releases page, the age of all releases is 3 hours ahead. My local timezone is GMT+3. This results in the latest releases showing up as being in the future, e.g. age is 3 hours but the release date is actually 3 hours in the future. Basically the releases page is showing me the time in GMT+6.

**Additional context**
I am running autobrr in a Debian 11 LXC container. The timezone of the container is GMT+3. I have a postgres 14 database as the autobrr database in another LXC container.

kaiserbh:
Changed the column type for release timestamps from 'timestamp' to 'timestamptz' in the PostgreSQL database. This update ensures that the release ages are accurately reflected in the web UI by storing the correct timezone information. Previously, the data was stored in UTC but interpreted as local user time, causing a 3-hour offset in the web UI or in my case 11-hour offset. The conversion now correctly interprets and displays the release times based on UTC.

For clarity:
image
image

The first 3 fields are before the fix and the last field with id `4` it's after the fix.

References #1138
closes #1138


KyleSanderson (commented):
Impressive they changed the definition of timestamp underneath people, mind you that was now 17 years ago.

How do the migrated dates work if they're missing the TZ? just stuck a couple hours off?

KaiserBh:
> Impressive they changed the definition of timestamp underneath people, mind you that was now 17 years ago.
> 
> How do the migrated dates work if they're missing the TZ? just stuck a couple hours off?

I thought by using with `ALTER COLUMN timestamp TYPE timestamptz USING timestamp AT TIME ZONE 'UTC';` in particular ` USING timestamp AT TIME ZONE 'UTC'` Should have converted it to UTC for existing ones. Apparently it doesn't? Because it was 11 hour offset before but it became 10 hour offset instead once I converted it to UTC/migrated the db.

image

KyleSanderson:
> > Impressive they changed the definition of timestamp underneath people, mind you that was now 17 years ago.
> > How do the migrated dates work if they're missing the TZ? just stuck a couple hours off?
> 
> I thought by using with `ALTER COLUMN timestamp TYPE timestamptz USING timestamp AT TIME ZONE 'UTC';` in particular ` USING timestamp AT TIME ZONE 'UTC'` Should have converted it to UTC for existing ones. Apparently it doesn't? Because it was 11 hour offset before but it became 10 hour offset instead once I converted it to UTC/migrated the db.

I would suspect the source isn't truly UTC, and is instead their local TZ, which aligns to the user reports around this (where we treat it as UTC, then augment further). It's far from critical for this, could land as-is, but feels a bit awkward to bring the struct in for a one-shot move.

KyleSanderson (approved):

zze0s (approved):
I can't recall why it wasn't timestamptz from the start but this does solve the issues users have been reporting.

Thanks for the work! :partying_face:

PR #1357, opened 2024-01-19
fix(web): irc settings light theme colors

martylukyy:
Fixes the colors of the IRC settings page for light mode.
Tested both modes now. Mea culpa.

Before:
image

After:
image


s0up4200 (approved):
LGTM on Chrome and FF!

PR #1362, opened 2024-01-21
fix(notifications): Pushover and Telegram formatting

zze0s:
Fixes a regression that removed the html attributes for Pushover and Telegram notifications.


vincejv (approved):
image
works great!

PR #1364, opened 2024-01-21
fix(web): releases page search and filtering

zze0s:
Fix for the search and filtering on Releases page.

### DIFF
--- a/web/src/screens/releases/ReleaseTable.tsx
+++ b/web/src/screens/releases/ReleaseTable.tsx
@@ -32,12 +32,11 @@ export const releaseKeys = {
   detail: (id: number) => [...releaseKeys.details(), id] as const
 };
 
-
 type TableState = {
-    queryPageIndex: number;
-    queryPageSize: number;
-    totalCount: number;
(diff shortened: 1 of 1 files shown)


martylukyy:
I think we should `gotoPage(0)` not only on `PAGE_SIZE_CHANGED` but also on `FILTER_CHANGED`...
That way the user doesn't end up with an empty page if he is further down the pages...
Opinions?

zze0s:
> I think we should `gotoPage(0)` not only on `PAGE_SIZE_CHANGED` but also on `FILTER_CHANGED`... That way the user doesn't end up with an empty page if he is further down the pages... Opinions?

Good call!! Fixed!

zze0s:
Would be great of someone could test this with a decent amount of releases in their db.

Changed from `useSuspenseQuery` to regular `useQuery` so that the page Suspense spinner isn't triggered. This should keep the state etc.

s0up4200 (approved):
Tested on SQLite db with 13k releases. LGTM!

PR #1367, opened 2024-01-22
feat(feeds): improve RSS size parsing

KyleSanderson:


internal/feed/rss_test.go
```diff
@@ -220,26 +223,111 @@ func Test_isMaxAge(t *testing.T) {
...
+	}
+
+	sizeStr = strings.Replace(sizeStr, " ", "", -1) // Remove any spaces
+	size, err := strconv.ParseFloat(sizeStr, 64)
+	if err != nil {
+		return 0, fmt.Errorf("failed to parse size: %v", err)
+	}
+
+	return uint64(size * float64(multiplier)), nil
+}
```
zze0s, line 333: Good effort, but `humanize.ParseBytes(str)` does this already.

internal/feed/rss.go
```diff
@@ -325,14 +324,12 @@ func isFreeleech(str []string) bool {
 	return false
 }
 
-// readSizeFromDescription get size from description
-func readSizeFromDescription(str string) string {
-	matches := rxpSize.FindStringSubmatch(str)
-	if matches == nil {
-		return ""
+// pullSizeFromDescription get size from description
+func pullSizeFromDescription(str string, r *domain.Release) {
```
zze0s, line 328: Minor nitpick but I think the old name `readSizeFromDescription` is better, or `parse` instead of `read`. we don't use the `pull` terminology anywhere else.

zze0s (commented):
Good improvements! :+1: Left some comments.

zze0s (approved):

PR #1368, opened 2024-01-23
docs: add contribution guide

zze0s:
Add contribution guide.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,182 @@
+# Contributing to autobrr
+
+Thanks for taking interest in contribution! We welcome anyone who wants to contribute.
+
+If you have an idea for a bigger feature or a change then we are happy to discuss it before you start working on it.
+It is usually a good idea to make sure it aligns with the project and is a good fit. Open an Issue or post in #dev-general on [Discord](https://discord.gg/WQ2eUycxyT).
+
+This document is a guide to help you through the process of contributing to autobrr.
+
+## Become a contributor
+
+* Code: new features, bug fixes, improvements
+* Report bugs
+* Documentation: The docs repo can be found here: [github.com/autobrr/autobrr.com](https://github.com/autobrr/autobrr.com)
+
+## Developer guide
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,182 @@
...
+
+This guide helps you get started developing autobrr.
+
+## Dependencies
+
+Make sure you have the following dependencies installed before setting up your developer environment:
+
+- [Git](https://git-scm.com/)
+- [Go](https://golang.org/dl/) (see [go.mod](go.mod#L3) for minimum required version)
+- [Node.js](https://nodejs.org)
```
martylukyy, line 26: Should we mention here that we historically always used the current node LTS version,
since we also mentioned a version for go the line before?

CONTRIBUTING.md
```diff
@@ -0,0 +1,182 @@
...
+First install the web dependencies:
+
+```shell
+cd web && pnpm install
+```
+
+Run the project:
+
+```shell
+pnpm start
```
martylukyy, line 65:
```suggestion
pnpm run dev
```

CONTRIBUTING.md
martylukyy, line 1: We also have the option for `make build` which does FE and BE at the same time...
Do we want to include that aswell since we currently only have FE and BE seperately?

CONTRIBUTING.md
zze0s: Mentioning node LTS is a good idea! We could add the Go version but I think it might change more often in the future once we use 1.21 and up.

CONTRIBUTING.md
zze0s: I think simply `pnpm dev` is enough isn't it?

martylukyy: Yes, pnpm dev is enough.
It assumes the run.

s0up4200 (approved):

martylukyy (approved):
Did a few changes. LGTM! 🚀

PR #1378, opened 2024-01-26
feat(macros): add Group

vincejv:
Add "Group" to available filter macros, ie available as `{{ .Group }}`


zze0s (approved):
I wasn't aware this was missing :smile: Thanks for the PR, and updating the docs as well! :partying_face:

PR #1385, opened 2024-01-31
feat(actions): Exec handle timeouts

s0up4200:


internal/action/exec.go
```diff
@@ -32,14 +32,20 @@ func (s *service) execCmd(ctx context.Context, action *domain.Action, release do
...
 	// setup command and args
-	command := exec.CommandContext(ctx, cmd, args...)
+	command := exec.CommandContext(timeoutCtx, cmd, args...)
 
 	// execute command
+	start := time.Now()
 	output, err := command.CombinedOutput()
 	if err != nil {
+		if timeoutCtx.Err() == context.DeadlineExceeded {
+			return errors.Wrap(err, "executing command timed out after 30 seconds: %s args: %s", cmd, args)
```
zze0s, line 47: Message says 30 seconds but the context timeout has 120 seconds.

zze0s (commented):
Timeouts are good to have but I'm actually now 100% sure if this is a good idea for actions :thinking: 

we might have to ask users how they are using actions if anyone have long running tasks from exec scripts or not. Alternatively we add a new setting for the timeout so if anyone needs to have a longer one they can set it themselves.

s0up4200:
Good input! Will close this for now as it's not really a requested feature. Don't want to break existing workflows etc.

It can be revisited later with a more robust implementation, making it an optional setting that is disabled on existing actions for example.

PR #1393, opened 2024-02-07
feat(build): add more stages to multi-stage Dockerfile builds

KyleSanderson:


zze0s (commented):
Is this change for the sake of change?

The app-builder builds, and then we have the runner. Why introduce a middle layer between?

And I thought we talked about the nonsense branch names.

PR #1394, opened 2024-02-07
feat(macros): add Type

bbeck:
Exposes the Type field to macros.  This is beneficial for external scripts and webhooks that may want to alter their behavior based on the type of the Release.


zze0s (approved):
Thanks for your contribution! This is a good addition and something we probably should have exposed in filters as well in the future :+1:

PR #1395, opened 2024-02-11
fix(indexers): DigitalCore InfoURL

broken-stones:
Append a trailing slash to infourl to avoid site redirecting to home page.


zze0s (approved):
Nice find! :+1:

PR #1396, opened 2024-02-12
feat(irc): disable auto chan part when using bouncer

Flupster:
This pull request addresses an issue where IRC clients connected through an IRC bouncer send a PART command which leaves a channel for other clients using the bouncer.


zze0s:
Hi! This is a dupe of https://github.com/autobrr/autobrr/pull/1163 which I closed because ZNC have built in handling for this, see this issue https://github.com/autobrr/autobrr.com/issues/125

I'm open for discussion however :slightly_smiling_face:

Flupster:
> Hi! This is a dupe of #1163 which I closed because ZNC have built in handling for this, see this issue [autobrr/autobrr.com#125](https://github.com/autobrr/autobrr.com/issues/125)
> 
> I'm open for discussion however 🙂

I just feel it should not be the job of the bouncer but more so the job of autobrr, we know if the connection is using a bouncer, so it doesn't hurt to just never part channels if that's the case. That's just my opinion though, znc has plugins and options to prevent a problem that autobrr causes.

Also upon thinking about it some more, the PartChannel function in handler.go should do the logic and return early if using a bouncer.

zze0s:
> I just feel it should not be the job of the bouncer but more so the job of autobrr, we know if the connection is using a bouncer, so it doesn't hurt to just never part channels if that's the case. That's just my opinion though, znc has plugins and options to prevent a problem that autobrr causes.

I'm unsure how many users we have that use a bouncer, but it probably makes sense to have it built in.

> Also upon thinking about it some more, the PartChannel function in handler.go should do the logic and return early if using a bouncer.

That sounds like a better solution than having all the callers check it. Go ahead and make that change and we'll get this merged.

If there's complaints later we can add a new setting and toggle to have it optional when using a bouncer.

zze0s (approved):
I went ahead and made the change myself.

Thanks for the original PR and starting the discussion again!

zze0s:
I think so yes, but I have not tested it.

secretcapy:
I have tested. It works as intended.

If anybody else is confused like I was, there is no toggle (at least for now). Channel auto-part is disabled as soon as you use a bouncer in 1.38.0.

Really appreciate this feature, thank you! @Flupster / @zze0s

yorah:
Seems like channel auto-part was re-introduced since then. I stumbled on https://github.com/autobrr/autobrr.com/issues/125 , and enabling sticky chan solved my problem (autobrr /part'ing other chans).

yorah:
Nevermind... that was on my side, works as expected when the Bouncer flag is on.

PR #1400, opened 2024-02-12
enhancement(web): debloat release loading animation

martylukyy:
This PR further cleans up the code for the loading animation of the releases table.

### DIFF
--- a/web/src/screens/releases/ReleaseTable.tsx
+++ b/web/src/screens/releases/ReleaseTable.tsx
@@ -241,17 +241,9 @@ export const ReleaseTable = () => {
           ) }
         </div>
         <div className="bg-white dark:bg-gray-800 border border-gray-250 dark:border-gray-775 shadow-lg rounded-md mt-4">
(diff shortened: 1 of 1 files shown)


zze0s (commented):
Before these changes the EmptyReleaseList component and Loading looks the same with that little darker header bit, but with this that is removed. Either keep leave as is or remove that as well so it doesn't look too different.

martylukyy:
I just checked out this branch again and started dev mode and for me it has the animation with the darker header 🤔 
I even changed the color of the spinner for a sanity check and used both FF and chrome..

It creates the main area div, then the header div and then simulates a `h-10` high span for the header to show...

https://github.com/autobrr/autobrr/blob/66bda9c7632a1a349654f876cc9c62fc50bed895/web/src/screens/releases/ReleaseTable.tsx#L243-L245

Mind double checking again?

zze0s:
Ah I see, just saw less code so thought it was different. Will check it out locally :+1:

zze0s (approved):

PR #1401, opened 2024-02-12
enhancement(web): filters list query loading animation

martylukyy:
When the filters query takes too long to finish, a message will be shown that on filters have been created.
This PR implements a loading animation similar to the release table one, just for the filters list.


zze0s (commented):
I moved the loader inside the table/content so it still has the headers. When changing Indexer or Sort filter it will trigger a isLoading so with this it still has the menu header.

image

zze0s (approved):

PR #1406, opened 2024-02-13
fix(indexers): BrokenStones add auth key to download url

broken-stones:


zze0s (approved):
Thanks for the fixes! :+1:

PR #1416, opened 2024-02-20
feat(macros): add FilterID

bbeck:
Exposes the FilterID field to macros. This is beneficial as another way to allow external tools to know which filter generated an event.  This should be more stable than the Filter Name which is already present because the Filter ID is immutable.


zze0s (approved):
Thanks for the PR! Another good addition :+1:

PR #1419, opened 2024-02-20
feat(irc): manually re-process announces

zze0s:
Implement re-process from network channel logs.

image

> [!IMPORTANT]
> Only supports single line announces to start with.

### DIFF
--- a/cmd/autobrr/main.go
+++ b/cmd/autobrr/main.go
@@ -8,7 +8,6 @@ import (
 	"os/signal"
 	"syscall"
 	_ "time/tzdata"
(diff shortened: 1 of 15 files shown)


martylukyy:
What do we think about this approach so that the reprocess button doesn't disappear behind the fullscreen one?

image

The text will still disappear behind the button but i cba fixing that for a couple of reasons.

zze0s:
> What do we think about this approach so that the reprocess button doesn't disappear behind the fullscreen one?
> 
> image
> 
> The text will still disappear behind the button but i cba fixing that for a couple of reasons.

That looks better indeed! :+1:

PR #1427, opened 2024-02-25
fix(web): make tooltips clickable if touchscreen is present

### LINKED ISSUE #1389: [WEBUI] Unable to click tooltips on certain setups
**Version**
v1.36.0

**Describe the bug**
Tooltips are not working as expected with a mouse and sometimes a touchpad if the app registers the device as a touch device.
E.g., a laptop with a touch screen or an iPad with a Magic Keyboard attached.

On the iPad, it only triggers the tooltip when touch is used. If using the touchpad on the Magic Keyboard, nothing happens.

**To Reproduce**
1. Go to the dashboard with a touch screen device
2. Click the action tooltip using a mouse (or an attached touchpad on the iPad)

Console error: `Unable to preventDefault inside passive event listener invocation.`

**Expected behavior**
It should work.

**Device tested:**
 - OS: iPadOS
 - Browser: Safari
 - Version: 17.3

**Additional context**
The bug was probably introduced by #1314

martylukyy:
This PR fixes/refactors the tooltips in order to make them clickable when a touchscreen is present.
I could only test this on a Android tablet so i would appreciate someone testing this with an iPad with Safari e.g.

potentially fixes #1389

### DIFF
--- a/web/src/components/tooltips/Tooltip.tsx
+++ b/web/src/components/tooltips/Tooltip.tsx
@@ -14,7 +14,7 @@ import { classNames } from "@utils";
 
 interface TooltipProps {
   label: ReactNode;
-  onLabelClick?: (e: React.MouseEvent<HTMLDivElement>) => void;
+  onLabelClick?: (e: React.MouseEvent<HTMLDivElement> | React.TouchEvent<HTMLDivElement>) => void;
   title?: ReactNode;
   maxWidth?: string;
   requiresClick?: boolean;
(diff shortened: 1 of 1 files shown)


martylukyy:
Upon further testing i noticed i produced another bug which has been fixed now.
There is still one minor inconvenience on touch devices on which a tooltip with a MouseEvent has been opened.
If you do not close the tooltip which got opened by that MouseEvent, 
it will keep disappear and appearing upon a TouchEvent.
So you need to have all tooltips, opened by a MouseEvent closed, before opening on with a TouchEvent.

If i get an idea on how to fix that i will but for now I think it's good enough.

zze0s (approved):
Nice work marty! :partying_face:

PR #1452, opened 2024-03-03
fix(notifications): ntfy.sh Typo in Content-Type header value

felix4643:
Header value should be text/plain and not text/pain


zze0s (approved):
Oh wow, that's embarrassing 🙈 

Thanks for the fix!!

Issue #1454, opened 2024-03-08
Smart Episode setting did not skip an older TV episode


kenstir:
**Version**
v1.39.1

**Describe the bug**
My filter has the "smart episode" option enabled.  I expected it to skip an old episode announcement, but it did not.

Timeline (edited for clarity):
* `The Dealy Show 2024 03 05 Guest 1` matched and downloaded
* `The Dealy Show 2024 02 28 Old Guest` matched and downloaded
* `The Dealy Show 2024 03 06 Guest 2` rejected due to "max downloads (1) this (DAY) reached"

I believe the metadata does not have the information needed and we would have to parse it from the name.  That is, we could deduce the season as 2024 and the episode as "0305" and reject S 2024 E 0228 as "older".

Does this make sense to you?  If so I am willing to try to fix it.  I could uses a hint, but the relevant code appears to be:
```
func (repo *ReleaseRepo) CanDownloadShow(ctx context.Context, title string, season int, episode int) (bool, error) {
	// TODO support non season episode shows
```

**Additional Context**
The FILTER_APPROVED log message suggests that season / episode were not parsed or present in the metadata:
```
Season:0 Episode:0 Year:2024 
```

kenstir:
So @KyleSanderson , do you think a change like this will cause upgrade problems?  I've been watching announces on one tracker for a week, and the US daily shows all seem to be announced with year=2024 season=0 episode=0, but with YYYY.mm.dd in the name.  I haven't dug deeply enough yet, but my hypothesis is that an upgrade would work like this:

* start autobrr v1.39.1
* filter `The Dealy Show 2024 03 05 Guest 1` as `Year=2024 Season=0 Episode=0`
* upgrade autobrr with this change
* filter `The Dealy Show 2024 02 28 Old Guest` as `Year=2024 Season=2024 Episode=228`

That is, Season=2024 Episode=228 should still be a valid new episode for smart episodes.

kenstir:
@KyleSanderson I trust your judgement on this, if you think this is not a good or useful enhancement, feel free to close this issue.  I will write an external filter.

I am not following the example however.  I get that you are inventing titles to make a point, but is it realistic?  Are those 3 titles referencing 3 different episodes or the same episode?

PR #1462, opened 2024-03-19
feat(database): Add month, day columns to release table

kenstir:
First step in fixing #1454 - add month, day to database

### DIFF
--- a/internal/database/release.go
+++ b/internal/database/release.go
@@ -38,8 +38,8 @@ func (repo *ReleaseRepo) Store(ctx context.Context, r *domain.Release) error {
 
 	queryBuilder := repo.db.squirrel.
 		Insert("release").
-		Columns("filter_status", "rejections", "indexer", "filter", "protocol", "implementation", "timestamp", "group_id", "torrent_id", "info_url", "download_url", "torrent_name", "size", "title", "category", "season", "episode", "year", "resolution", "source", "codec", "container", "hdr", "release_group", "proper", "repack", "website", "type", "origin", "tags", "uploader", "pre_time", "filter_id").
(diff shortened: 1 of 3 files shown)


zze0s:
Hope you don't mind me hijacking your PR @kenstir :smile: 

Since you started on adding month and day fields I went all the way to make them filterable and fix smart episode checks as well.

Still needs to test things but the bulk of it is there now.

kenstir:
I am very happy to have you hijack my PR @zze0s  😄  I am pretty green in Golang and was honestly a bit stuck!

zze0s (dismissed):
This should be good to go now!

Tested all changes with both daily shows and regular season/ep based and works great.

I never used the Smart Episode feature before, but is it correct that it allowed to grab multiple of the latest current release? I changed that behavior so it will only grab at most one of the currently latest release except for REPACK or PROPER which it now will take into consideration.

Could you help out with testing this? Backup your database before you continue since this does have database migrations.

kenstir:
Awesome, thank you!!  I will test it for sure (and thanks for the backup tip).

My position is that the Smart Episode should never grab multiple of any episode, except as you say to correct mistakes (REPACK or PROPER).

kenstir:
As far as I can see everything is working great with this PR.  Thank you @zze0s and @s0up4200!!

I tested using a local build and MockIndexer, and I replayed all the problematic announces from my autobrr.log.  Every check passed.  I tested:
* Releases with new dates - matched
* Release with different resolution for the same date - rejected as not new
* Old dates - rejected as not new
* Same date with PROPER - matched with Other:[PROPER]

zze0s:
So apparently this change to only allow for a single latest ep now breaks multiple resolutions, hdr variants, codecs and more.

> so, with [d1359de](https://github.com/autobrr/autobrr/commit/d1359de30e10041c63d8c1ffceb590561243abff) the devil is in the details. By not allowing the current episode you're preventing any possible qualities or future releases to ship. This includes the scenario for when the first release is pulled, or is transcoded from a VHS tape and then airs digitally after. This also breaks daily shows that have multiple episodes on the same day, or more commonly, parts that come out the same day (part 1, part 4, etc).
> 
> It's very subtle, but it's very important that at least the current episode/day is allowed to continue to get releases until such a point that there's a newly released title. Marty at some point suggested even a grace period around this where the new release is ignored, which is compatible with reasonable expectations.

What is your proposed solution? Change back to just `Gt` from `GtOrEq` and handle that with Skip Duplicates?

zze0s (approved):
Reverted the `GtOrEq` change so now it will allow multiple episodes to pass again.

With the introduction of **Skip Duplicates** users will be able to easily control this and only allow a single release per episode if they choose to.

PR #1468, opened 2024-03-20
fix(autobrrctl): prevent empty password

### LINKED ISSUE #1466: Locked out of authentication after changing Password to blank
**Version**
 v1.39.1
 docker compose

**Describe the bug**
I changed my password to blank with the GUI thinking it would disable authentication but when i try to log in it tells me a password is required. 
I have used autobrrctl --config /home/username/.config/autobrr change-password <USERNAME> 
which gives:
{"level":"info","module":"database","type":"sqlite","time":"2024-03-20T09:48:47Z","message":"Beginning database schema upgrade from version 0 to version: 55"}
{"level":"info","module":"database","type":"sqlite","time":"2024-03-20T09:48:47Z","message":"Database schema upgraded to version: 55"}
and prompts me for a password, however the new password does not work and i get "Wrong password or username!"

 - OS: Fedora 37
 - Browser Firefox

martylukyy:
fixes #1466

### DIFF
--- a/cmd/autobrrctl/main.go
+++ b/cmd/autobrrctl/main.go
@@ -139,6 +139,10 @@ func main() {
 			log.Fatalf("failed to read password: %v", err)
 		}
 
+		if len(string(password)) == 0 {
+			log.Fatal("Password cannot be empty.")
+		}
+
 		hashed, err := authSvc.CreateHash(string(password))
 		if err != nil {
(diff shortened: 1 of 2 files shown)


zze0s (approved):
Simplified it by moving the len check inside `readPassword()` and added a len check to the `auth.CreateHash` method.

PR #1470, opened 2024-03-21
feat(filters): add sources and codecs

martylukyy:
This PR adds:
- AHDTV tag to sources
- MPEG-2 tag to codecs

Also sort alphabetically.


zze0s (approved):
The previous order of the codecs and sources was `Most commonly used` to `least commonly used`, but alphabetical order is easier to scan :+1:

PR #1475, opened 2024-03-24
fix(auth): correct redirect on bad session cookie

zze0s:
Speculative fix for broken redirect to /login when using baseurl.


martylukyy (approved):
Squashed that bug like a boss! 😄

PR #1482, opened 2024-03-26
feat(indexers): Nyaa add trusted, remake and batch tags

Flupster:
The NekoNeko announcer for Nyaa now appends trusted/remake/batch to the end of the announcement line, this PR catches these as tags for use in filters.

While the batch tag is not used on Nyaa, it's mentioned in the FAQ so the addition is there.

**Tags added**:
 - trusted
 - remake
 - batch

### DIFF
--- a/internal/indexer/definitions/nyaa.yaml
+++ b/internal/indexer/definitions/nyaa.yaml
@@ -44,28 +44,37 @@ irc:
     type: single
     lines:
       - tests:
-          - line: '[Live Action - Raw] - 見て、お母さん、私は日本語が書けます！ - (220.30MiB) - https://nyaa.si/view/000000/torrent'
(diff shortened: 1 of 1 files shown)


Flupster:
Trusted and normal releases are tested working, I'm sure other tags will work also 👍

martylukyy (approved):
LGTM!
Info: Bot will get changes pushed as soon this get's merged.

zze0s (commented):
Marty probably already knows what this is but would be great if you could add a description for others not in the loop :smiley: @Flupster

Flupster:
> Marty probably already knows what this is but would be great if you could add a description for others not in the loop 😃 @Flupster

I've edited the PR to add a description 😁 let me know if any amendments are needed

PR #1483, opened 2024-03-27
fix(database-converter): use string concat instead of var substitution

vincejv:
We don't need to sanitize the input as it doesn't come from the user and this is simply not allowed in any sql database: https://stackoverflow.com/questions/20678725/how-to-set-table-name-in-dynamic-sql-query

Will test on my local to see if there are other migration issues

### DIFF
--- a/internal/database/tools/convert.go
+++ b/internal/database/tools/convert.go
@@ -93,7 +93,7 @@ func GetTables() []string {
 func (c *SqliteToPostgresConverter) migrateTable(sqliteDB, postgresDB *sql.DB, table string) []string {
 	var fkViolationMessages []string
 
-	rows, err := sqliteDB.Query("SELECT * FROM ?", table)
+	rows, err := sqliteDB.Query("SELECT * FROM " + table)
 	if err != nil {
 		log.Fatalf("Failed to query SQLite table '%s': %v", table, err)
 	}


vincejv:
Yes it doesn't running with the current code will just result in an error due to thing mentioned in the post, SQL Table names cannot have var substitution, query must be concatenated manually

```bash
autobrrctl db:convert --sqlite-db autobrr_final.db --postgres-url \
      postgres://autobrr:autobrr@postgres.docker.internal:5432/autobrrtest
2024/03/28 05:43:25 Failed to query SQLite table 'action': SQL logic error: near "?": syntax error (1)
```

vincejv:
There are other migration issues actually but not sure how to solve them, I've fixed it manually by running sql scripts

1. Run Auto-increment index reset for all tables with AI PK, or else new inserts would fail due to id already existing.
2. Failed migration of actions, indexer tables and etc due to FK violations, workaround is to run the command twice. Possible fix is to migrate tables in order.

zze0s:
It's indeed not super straight-forward to switch database engines 😄 

The migrate commands had some issues at time of merge, and I did not give it the priority it should have gotten.

If you make further improvements then keep it all in this PR so it's easier to manage.

vincejv:
I'm not too sure on how to fix the other 3 things, i will be closing this, but i guess this for future reference.
Had no issues so far after the manual migration

PR #1484, opened 2024-03-28
feat(indexers): add KeepFRDS

KyokoMiki:
### Regex Tests

https://regex101.com/r/X5fdZu/1

image

### DIFF
--- a/internal/indexer/definitions/keepfrds.yaml
+++ b/internal/indexer/definitions/keepfrds.yaml
@@ -1,0 +1,78 @@
+---
+#id: keepfrds
+name: KeepFRDS
+identifier: keepfrds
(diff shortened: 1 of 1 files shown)


martylukyy (approved):
Adapted regex pattern to account for ` - ` in release names 
Fixed indentation.

Would appreciate if you could take a look over the changes in the regex in case they cause some problems with other release tags!

KyokoMiki:
> Adapted regex pattern to account for `-` in release names Fixed indentation.
> 
> Would appreciate if you could take a look over the changes in the regex in case they cause some problems with other release tags!

Thank you! I took a look, and the release tags field will not contain '-', so there should not be any conflicts now. Thank you very much for your correction!

PR #1490, opened 2024-04-05
feat(web): auto refresh dashboard stats

### LINKED ISSUE #1488: [FEATURE REQ] Automatic Refresh of homepage or recent activity box, either by time or when a filter match is made.
I tried search but didn't see anything, sorry if its a duplicate.

when I see an item approved via notification on my phone with pushover but when I check the home page that's already in an open tab it doesn't show an updated recent box until I hit F5 to force a reload.

Ideally home page is refreshed to show updated info for recent capture, triggered by the item approval and would also be nice to have desktop notification from Chrome/Firefox browser of new show/movie being grabbed.

set time for auto refresh of homepage within settings?

**Additional context**
Love the program, really the only minor issue i found so far from testing

martylukyy:
This PR adds a refresh interval of 15 seconds for:
- The stats on the dashboard page
- The recent activity table on the dashboard page
- The releases table on the releases page

Closes #1488 

// EDIT:
Note: This will not refresh any data such as the time of the left on the table if there is no new releases to be shown.
This is by design.


web/src/api/queries.ts
```diff
@@ -108,21 +108,24 @@ export const ReleasesListQueryOptions = (offset: number, limit: number, filters:
...
+    refetchOnWindowFocus: false,
+    refetchInterval: 15000  // refetch recent activity table on dashboard page every 15s
   });
 
 export const ReleasesStatsQueryOptions = () =>
   queryOptions({
     queryKey: ReleaseKeys.stats(),
     queryFn: () => APIClient.release.stats(),
-    refetchOnWindowFocus: false
+    refetchOnWindowFocus: false,
```
zze0s, line 127: Maybe these should have `refetchOnWindowFocus: true` so they refetch whenever a user changes tab or window?

martylukyy: I tried that aswell and it only refetch when i select the tab again,  
not on taking the focus from chromium away and setting the focus on chromium again.
But let's add it.

zze0s (approved):

PR #1501, opened 2024-04-12
feat(filters): validate existence of external exec cmd

zze0s:
Add check for External Filter and Actions Exec program if it exists on save.

And improve cleaning of filter data.

### DIFF
--- a/internal/domain/filter.go
+++ b/internal/domain/filter.go
@@ -270,7 +270,7 @@ func (f *Filter) Validate() error {
 
 	for _, action := range f.Actions {
 		if action.Type == ActionTypeExec {
-			if action.ExecCmd != "" {
+			if action.ExecCmd != "" && action.Enabled {
 				// check if program exists
 				_, err := exec.LookPath(action.ExecCmd)
 				if err != nil {
--- a/internal/domain/filter.go
+++ b/internal/domain/filter.go
@@ -6,7 +6,6 @@ package domain
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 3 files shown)


internal/domain/filter.go
```diff
@@ -254,6 +256,63 @@ func (f *Filter) Validate() error {
 		return fmt.Errorf("error validating filter size limits: %w", err)
 	}
 
+	for _, external := range f.External {
+		if external.Type == ExternalFilterTypeExec {
+			if external.ExecCmd != "" {
```
s0up4200, line 261:
```suggestion
			if external.ExecCmd != "" && external.Enabled {
```
Would this make more sense? Right now it errors on not found, even if disabled.

internal/domain/filter.go
```diff
@@ -254,6 +256,63 @@ func (f *Filter) Validate() error {
...
+				if err != nil {
+					return errors.Wrap(err, "could not find external exec command: %s", external.ExecCmd)
+				}
+			}
+		}
+	}
+
+	for _, action := range f.Actions {
+		if action.Type == ActionTypeExec {
+			if action.ExecCmd != "" {
```
s0up4200, line 273:
```suggestion
			if action.ExecCmd != "" && action.Enabled
```
Same here I guess!

internal/domain/filter.go
zze0s: Yeah good idea, that way a user can still save it if disabled and fix it :+1:

s0up4200 (approved):

PR #1502, opened 2024-04-12
fix(user): change username

zze0s:
Fix for changing only username from settings.


s0up4200 (approved):
Tested with sqlite, works 🚀

PR #1505, opened 2024-04-12
fix(filters): export not working with Safari

### LINKED ISSUE #1487: [BUG] Export JSON does not work in Safari
**Version**
1.40.1

**Describe the bug**
The `Export JSON` function is not working in Safari.
It says:
```
Error
Failed to copy JSON to clipboard.
```

**To Reproduce**
1. Go to `...` next to a filter.
2. Click on `...`.
3. Scroll down and click on `Export JSON`.

**Expected behavior**
The JSON info gets copied to the clipboard.

macOS Sonoma Safari

**Additional context**
Also, is there no way to export actions and other things to JSON?

zze0s:
Fixes an issue where filters could not be exported with Safari.

Fixes #1487


s0up4200 (approved):
Tested with Safari on both latest iOS and MacOS. Both works.

PR #1508, opened 2024-04-14
feat(notifications): add support for custom Telegram api url

KyokoMiki:
Telegram API is blocked in some places. Therefore, I added the feature to allow customizing Telegram API domain name. Very few changes were made. If not specified, it directly uses "https://api.telegram.org/" itself.


zze0s (approved):
Hey! Thanks for the PR!

Was not aware this was needed or possible, but great to now have the possibility to change it.

PR #1511, opened 2024-04-16
feat(macros): add `IndexerName`

zze0s:
Add new macro variable `.IndexerName`. The `.Indexer` is the indexer identifier.

Initiated after a discussion with @zakkarry to improve data based matching in [cross-seed](https://github.com/cross-seed/cross-seed) v6.

This change required a bit of refactoring on both the backend and frontend, but is something that should have been done from the start.

Now we pass around a minimal indexer struct with Id, Name and Identifier so we can easier choose what we want to use instead of just the Indexer Identifier which has limited use.

The releases still store just the Indexer Identifier as key but in the future we should fix that to use the numeric ID instead but that will require a custom data migration to handle cases of deleted indexers. We might have to move to soft-deletes of indexers to make that easier in the future.

Side note: the arr action still use the Identifier for now, but the plan is to change these to a new External Arr ID/name of the indexer and this refactoring would have been required to support that either way.

### DIFF
--- a/internal/domain/macros.go
+++ b/internal/domain/macros.go
@@ -28,6 +28,8 @@ type Macro struct {
 	DownloadUrl         string
 	InfoUrl             string
 	Indexer             string
+	IndexerName         string
+	IndexerIdentifier   string
 	Title               string
 	Type                string
 	Category            string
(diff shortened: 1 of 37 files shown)


internal/domain/macros.go
```diff
@@ -67,7 +69,9 @@ func NewMacro(release Release) Macro {
 		GroupID:             release.GroupID,
 		InfoUrl:             release.InfoURL,
 		DownloadUrl:         release.DownloadURL,
-		Indexer:             release.Indexer,
+		Indexer:             release.Indexer.Identifier,
```
s0up4200, line 72: Should this one be just `release.Indexer.Id`?

zze0s: Ideally it could be the full struct but that would break filters that already use and expect the identifier. We could do a db migration tho 🤔

PR #1522, opened 2024-04-23
feat(releases): delete based on age/indexer/status

s0up4200:
Adds option to delete releases not only by age, but also by indexers and status.
Indexers and statuses are optional while age is still a mandatory setting to avoid accidental nuke.

The age-`select` can match the font color of the other two, but then it wont go back to white after selecting an option. Test it out to see what I mean.

Front end:
CleanShot 2024-04-28 at 14 39 09@2x


internal/database/release.go
```diff
@@ -581,8 +581,16 @@ func (repo *ReleaseRepo) Delete(ctx context.Context, req *domain.DeleteReleaseRe
...
+	defer func() {
+		if p := recover(); p != nil {
+			tx.Rollback()
+			panic(p)
+		} else if err != nil {
+			tx.Rollback()
+		} else {
+			err = tx.Commit()
+		}
+	}()
```
zze0s, line 593: We should try to avoid `panic()`.

Replace with this snipped in `internal/database/download_client.go:283`

```go
	defer func() {
		var txErr error
		if p := recover(); p != nil {
			txErr = tx.Rollback()
			if txErr != nil {
				r.log.Error().Err(txErr).Msg("error rolling back transaction")
			}
			r.log.Error().Msgf("something went terribly wrong panic: %v", p)
		} else if err != nil {
			txErr = tx.Rollback()
			if txErr != nil {
				r.log.Error().Err(txErr).Msg("error rolling back transaction")
			}
		} else {
			// All good, commit
			txErr = tx.Commit()
			if txErr != nil {
				r.log.Error().Err(txErr).Msg("error committing transaction")
			}
		}
	}()
```

internal/database/release.go
```diff
@@ -599,37 +607,68 @@ func (repo *ReleaseRepo) Delete(ctx context.Context, req *domain.DeleteReleaseRe
...
 	}
 
+	//if req.OlderThan > 0 {
+	//	thresholdTime := time.Now().Add(-time.Duration(req.OlderThan) * time.Hour)
+	//	qb = qb.Where("timestamp < ?", thresholdTime)
+	//}
+
+	if len(req.Indexers) > 0 {
+		placeholders := strings.Repeat("?,", len(req.Indexers)-1) + "?"
+		qb = qb.Where(fmt.Sprintf("indexer IN (%s)", placeholders), toInterfaceSlice(req.Indexers)...)
```
zze0s, line 617: There is a built in `sq.Placeholders(len(req.Indexers))` function you can use here.

```diff
-		placeholders := strings.Repeat("?,", len(req.Indexers)-1) + "?"
-		qb = qb.Where(fmt.Sprintf("indexer IN (%s)", placeholders), toInterfaceSlice(req.Indexers)...)
+		qb = qb.Where(fmt.Sprintf("indexer IN (%s)", sq.Placeholders(len(req.Indexers))), toInterfaceSlice(req.Indexers)...)
```

And it maybe should be using `sq.Expr` instead of a fmt.Sprintf 🤔 or just a `sq.Eq` per https://github.com/Masterminds/squirrel/issues/334 should work.

internal/database/release.go
```diff
@@ -599,37 +607,68 @@ func (repo *ReleaseRepo) Delete(ctx context.Context, req *domain.DeleteReleaseRe
...
+	}
+
+	if len(req.ReleaseStatuses) > 0 {
+		statusPlaceholders := strings.Repeat("?,", len(req.ReleaseStatuses)-1) + "?"
+		subQuery := sq.Select("release_id").From("release_action_status").Where(fmt.Sprintf("status IN (%s)", statusPlaceholders), toInterfaceSlice(req.ReleaseStatuses)...)
+		subQueryText, subQueryArgs, err := subQuery.PlaceholderFormat(sq.Question).ToSql()
+		if err != nil {
+			return errors.Wrap(err, "error building subquery")
+		}
+		qb = qb.Where("id IN ("+subQueryText+")", subQueryArgs...)
```
zze0s, line 627: Similar here as above for indexers.

zze0s (commented):
Great addition!

Some comments for improvements. I can fix them if you want.

zze0s (approved):

PR #1526, opened 2024-04-26
chore(indexers): danishbytes remove unused url

kbhsn4:
Domain no longer used


martylukyy (commented):
Thoughts on doing an SQL oneliner to substitute the removed domain with the default one?

PR #1530, opened 2024-05-02
fix(web): use absolute paths for router links

martylukyy:
This PR branches off the branch of #1527 

Due to fixes in `"@tanstack/router-devtools": "^1.31.1"`  
the routes in the arrays for the settings and the filter pages had to be adapted.

During fixing the routes i saw that we had a `ReleasesIndexRoute` that couldn't be accessed  
and that isn't in use for some other part from what i can tell.  
So took the freedom and removed it and moved the component into the parent route `ReleasesRoute`

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -40,7 +40,7 @@
     "@tailwindcss/forms": "^0.5.7",
     "@tanstack/react-query": "^5.29.2",
     "@tanstack/react-query-devtools": "^5.29.2",
-    "@tanstack/react-router": "1.31.0",
+    "@tanstack/react-router": "^1.31.6",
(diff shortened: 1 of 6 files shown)


zze0s:
Did you test this when running behind base url rp? Still OK? When it was implemented it was required, but the router lib has had at lot of changes since.

martylukyy:
Yes, i have been running this since yesterday morning on my server and the releases route works  
with all the filter, pagination and search functionalities on that page.
The rewritten routes for settings and filters also work with an nginx RP and baseURL.
I did not test other RPs.

zze0s (approved):

PR #1532, opened 2024-05-03
feat(indexers): update release baseurl on update

zze0s:
Update the indexer baseURL (for `info_url` and `download_url`) of releases when changed in indexer settings.


s0up4200 (approved):
Tested and works as you'd expect 🥳 

1. Changed baseurl in Indexer settings
2. Release table reflects the change with new `info_url` and `download_url` for old releases

PR #1534, opened 2024-05-03
feat(indexers): add External Identifier to map with ARR indexers

### LINKED ISSUE #1344: [FEATURE REQ] Override Indexer ID For Arr
I have an indexer configuration in Sonarr for a private tracker (TorrentLeech). This configuration named "TorrentLeech" lets me override the _Seed Ratio_ and _Seed Time_; this is to have different values for this private tracker than the default options I use for public trackers. When autobrr sends a release to Sonarr, the release is sent with the ID autobrr has for the indexer (`torrentleech`). This ID does not always match the name of the indexer in Sonarr (i.e. `"torrentleech" !== "TorrentLeech"`), causing it to fallback to the default indexer options.

As it is possible to override the download client ID for a filter action, I would like to have the same feature applied to the indexer ID.

I thought about creating multiple download clients configurations in Sonarr, but "Seed Ratio" nor "Seed Time" cannot be set there; I would need to have multiple instances of my download client with for it to work.

I tried looking in the issues and discussions of this repository, but could not find anything regarding this problem.

**Additional context**
This problem also applies to Radarr.

I do think that setting the ID there (in the filter action) is the best place. In Radarr, I have two indexer configuration for TorrentLeech: one for Freeleeches named "TorrentLeech (Freeleech)", and one for the rest named "TorrentLeech". Having the indexer ID override in the filter action would allow each action to send the release as the right Arr indexer (e.g. freeleeches to "TorrentLeech (Freeleech)" and others to "TorrentLeech").

This feature could be problematic if a filter has multiple indexers selected. Instead of simply having an override for the ID, it could be some sort of table or selective override.

### LINKED ISSUE #1533: When using Radarr/Sonarr as a client, no seed time is set for the torrent in qbittorrent.
**Version**
Docker v1.41.0

**Describe the bug**

When using autobrr and radarr/sonarr, the setting for “Seed Time” is not set in the qBittorrent client. In radarr/sonarr, a seed time is set in the download client settings for qBittorrent and is also set correctly for downloads that are sent to qBittorrent.

Only for downloads that are sent from autobrr to radarr/sonarr and then actually continue from the torrent client via the download client settings for qBittorrent, no seed time is set.

**To Reproduce**
1. setting the seed time in radarr/sonarr
2. setting up a sonarr/radarr client in autobrr
3. creating a filter for the clients set up in point 2.
4. wait until autobrr sends a torrent to radarr/sonarr and check the seed time in the qbittorrent client.
5. the torrent has no seed time
6. start the same torrent via sonarr/radarr and check in the qbittorrent client with active seed time setting in the torrent

**Expected behavior**

**Screenshots**
The screenshot shows the seed time option when sending the torrent directly to qbittorrent, which is unfortunately missing in Sonarr/Radarr.
image

**Additional context**
I had inquired about the seed time settings in autobrr. But then I saw that you can only set these settings if autobrr sends the torrent to qBittorrent itself. I think it makes sense to also allow these settings for the clients radarr/sonarr in autobrr so that you can set them.

zze0s:
Adds a new field `External Identifier` on indexer settings to map to the indexer name set in the ARR so it can understand which indexer the release came from and correctly set seedtime and ratiolimit as configured in the arr for the download client and indexer.

Fixes #1344 
Fixes #1533


zze0s:
Have tested with Radarr and Sonarr with great success! Correct indexer is shown in ui and seedtime/ratiolimit is correctly set in the download client as well.

PR #1539, opened 2024-05-06
feat(filters): skip duplicates

### LINKED ISSUE #375: Do not duplicate release download
I created filter which downloads releases from 5 different trackers through IRC announce channels.

The problem if within 1 minute is announced same release on every tracker, autobrr sends 5 torrents to client and although I have 5 items in clients

It is possible in filter to specify unique download no matter from which tracker it downloads ?

zze0s:
Implement skip duplicates handling in filters.

A user can set up different "Duplicate Profiles" where they can configure what properties must match for a release to be considered a duplicate. These profiles are selected per filter.

Todo

- [ ] Compare release (the one in the db might be different than what we get)
- [ ] Compare audio

Closes #375


vip5912:
I tried to test feat(filters): skip duplicates #1539.
I use latest commit https://github.com/autobrr/autobrr/pull/1539/commits/c211212f8e3f64f01381ea447f0ee11f1ba76655
If I use the existing database (1.43), then when migrating the database, the months and days columns are not created in the filter table and I get the error:
`error executing query: SQL logic error: table filter has no column named months`
If I use the new configuration, the database is created with months and days columns in the filter table.

kenstir:
As discussed in Discord, I am working on a new PR with these changes minus the conflicts.  I will add some comments in a review of PR though to settle some questions.

kenstir (commented):
Had some comments on sqlite3 migration and some possibly unrelated changes.

PR #1540, opened 2024-05-06
feat(web): add theme toggle to navbar

s0up4200:
Added a dynamic theme switcher to the navbar for light/dark mode toggling.

Additionally, it now listens for system-level theme changes, and defaults to system-preferred theme when localstorage is empty.

CleanShot 2024-05-06 at 15 47 08@2x
CleanShot 2024-05-06 at 15 46 59@2x

### DIFF
--- a/web/src/components/header/RightNav.tsx
+++ b/web/src/components/header/RightNav.tsx
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: GPL-2.0-or-later
  */
 
-import { Fragment, useEffect, useState } from "react";
+import { Fragment } from "react";
 import { UserIcon } from "@heroicons/react/24/solid";
 import { Menu, Transition } from "@headlessui/react";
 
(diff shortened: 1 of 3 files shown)


martylukyy (commented):
So while this solution works it has a certain caveat...
The solution assumes that RightNav is loaded.
If users are on /onboard /login or whatever else route that doesn't have RightNav loaded/imported,  
it will not trigger the transistion.

Ideally this should be done on a file that is always loaded?

s0up4200:
> So while this solution works it has a certain caveat... The solution assumes that RightNav is loaded. If users are on /onboard /login or whatever else route that doesn't have RightNav loaded/imported, it will not trigger the transistion.
> 
> Ideally this should be done on a file that is always loaded?

Good call! Only Tanstack-boys remember /login and /onboarding 😄  Did some improvements.

martylukyy (dismissed):
Works all on all routes now from what i can tell! 🚀

web/src/utils/Context.ts
```diff
@@ -77,9 +77,19 @@ const FilterListKey = "autobrr_filter_list";
...
+  const storedSettings = localStorage.getItem(SettingsKey);
+  let initialSettings: SettingsType;
+
+  if (storedSettings) {
+    initialSettings = JSON.parse(storedSettings);
+  } else {
+    // Fallback to OS theme preference if no settings are stored
+    const prefersDarkMode = window.matchMedia('(prefers-color-scheme: dark)').matches;
+    initialSettings = { ...SettingsContextDefaults, darkTheme: prefersDarkMode };
+  };
```
zze0s, line 89: Couldn't the initial state `SettingsContextDefaults` have `darkTheme: window.matchMedia('(prefers-color-scheme: dark)').matches,`?

The context merger just below this takes care of the same as above minus the colorscheme.

web/src/components/header/RightNav.tsx
```diff
@@ -10,14 +10,39 @@ import { Menu, Transition } from "@headlessui/react";
...
-        <Menu as="div" className="ml-3 relative">
+        <div className="mt-1 items-center">
+          <button
+            onClick={toggleTheme}
+            className="p-1 rounded-full focus:outline-none focus:none transition duration-100 ease-out transform hover:bg-gray-200 dark:hover:bg-gray-800 hover:scale-100">
+            {themeIcon ? (
+              <MoonIcon className="h-4 w-4 text-gray-500 transition duration-100 ease-out transform" aria-hidden="true" />
+            ) : (
+              <SunIcon className="h-4 w-4 text-gray-600" aria-hidden="true" />
+            )}
```
zze0s, line 42: Don't really see a need here for the new variable `themeIcon`. It's a bit clearer to read `settings.darkTheme` in the button conditional for the icon.

zze0s (approved):

PR #1541, opened 2024-05-07
feat(macros): add missing variables

MagiX13:
As discussed in #1461, this breaks out the macros that can be exposed. 
I went through /internal/domain/releases.go and compared what was there already vs what was not. I'm not sure if all are actually used/filled at some point in time, but this might make it a bit futureproof :)


internal/domain/macros.go
```diff
@@ -16,86 +16,125 @@ import (
...
-	TorrentDataRawBytes       []byte
-	MagnetURI                 string
-	Group                     string
-	GroupID                   string
-	DownloadUrl               string
-	InfoUrl                   string
-	Indexer                   string
-	IndexerName               string
-	IndexerIdentifier         string
-	IndexerIdentifierExternal string
```
zze0s, line 33: Looks like this `IndexerIdentifierExternal` is missing in the updated struct.

zze0s (commented):
A bit hard to see exactly what changed since everything was reordered, but I did see `IndexerIdentifierExternal` missing.

There are some values that are not really exposed in filters etc but we can keep those. 

It might be easier to just embed the Release struct if we use all values :thinking: 

```golang
type Macro struct {
	Release
	CurrentDay          int
	CurrentHour         int
	CurrentMinute       int
	CurrentMonth        int
	CurrentSecond       int
	CurrentYear         int
	IndexerIdentifier   string
}
```

Didn't check all unique fields but something like it.

Thoughts?

Once https://github.com/autobrr/autobrr/pull/1462 is merged we can add `Month` and `Day` to macros.

zze0s (approved):

PR #1580, opened 2024-06-20
refactor(web): update deprecated HeadlessUI v2 components

martylukyy:
This PR refactors the remaining HeadlessUI v2 components that moved away from the dot notation 
but did not contain breaking changes and are now just deprecated.

Note: The `Field` component of HeadlessUI v2 collides with the `Field` component of Formik in some places.
Please check if the name of the renamed imports are suitable or should be changed.


web/src/components/inputs/input_wide.tsx
```diff
@@ -5,7 +5,7 @@
 
 import { Field } from "formik";
 import Select from "react-select";
-import { Switch } from "@headlessui/react";
+import { Field as Group, Label, Description } from "@headlessui/react";
```
zze0s, line 8: It might make more sense to change the Formik imports in this case to either `* as Formik` and use dot import or `Field as FormikField`.

Issue #1535, opened 2024-05-03
OPS not parsing log score


xKDHx:
**Version**
1.41.0

**Describe the bug**
Autobrr isn't parsing the log score for OPS releases. 

Announce:
[08:11] <&xxxxxx> TORRENT: Amorphous Androgynous – Tales Of Ephidrina – [2023] [Album] CD/FLAC/Lossless/Cue/Log/100 – ambient,electronic – https://xxxxxxx.xxxxxxx/torrents.php?id=xxxxxxx – https://xxxxxxx.xxxxxxx/torrents.php?id=xxxxxxx&torrentid=xxxxxxx&action=download

```
DBG
(OPS CD FLAC Test) for release: Amorphous Androgynous - Tales Of Ephidrina [2023] (CD FLAC Lossless) rejections: (wanted: perfect flac. got: [Cue FLAC Lossless Log], log score. got: 0 want: 100) method=CheckFilter module=filter
DBG
filter OPS CD FLAC Test rejected release: wanted: perfect flac. got: [Cue FLAC Lossless Log], log score. got: 0 want: 100 filter="OPS CD FLAC Test" indexer=ops module=release release="Amorphous Androgynous - Tales Of Ephidrina [2023] (CD FLAC Lossless)"
```
```
{
    "name": "OPS CD FLAC Test",
    "version": "1.0",
    "data": {
        "enabled": true,
        "max_downloads": 1,
        "max_downloads_unit": "EVER",
        "formats": [
            "FLAC"
        ],
        "quality": [
            "Lossless"
        ],
        "media": [
            "CD"
        ],
        "perfect_flac": true,
        "cue": true,
        "log": true,
        "log_score": 100,
    }
}
```

I tested with and without the `perfect_flac` filter, same result, it parses the log score as 0.

zze0s:
Hey! OPS recently added log score and other values to the announce, so we need to update the parser to look for those values.

zze0s:
Fixed in #1592

PR #1592, opened 2024-07-12
fix(indexers): OPS log score parsing

ThyThal:
The message from OPS Announce does not contain a % sign and that seemed to break it. This is probably not the best fix but it is something.

> CD/FLAC/Lossless/Cue/Log/100

#1535 Attemping to fix this issue.


ThyThal:
#850

ThyThal:
#1535 Attemping to fix this issue.

zze0s (approved):
Thanks for the PR! 

It's good enough :smiley: I added some tests to verify it works as intended.

PR #1600, opened 2024-07-25
fix(web): filter dropdown cut off

### LINKED ISSUE #1597: Filter Options Cutoff On Lower Filters
**Version**
v1.43.0

**Describe the bug**
When attempting to access the '3 dot' options of a given filter, if they are on the lower part of the screen, the options are cut off. Once the options are visual, there should be a way to either scroll down, or render the page to add the options menu length to the bottom so you can see the last filters option menu

**To Reproduce**
1. Go to 'filters'
2. Have a filter that is at the bottom of the screen
3. Click on '...'
4. Notice you cannot see the options

**Expected behavior**
I expect that the page is rendered long enough to scroll all the way down, so when you click the bottom most filter, you can see the options

**Screenshots**
image

 - OS: win11
 - Browser: reproducible on edge and firefox
 - Version: edge 126.0.2592.102, firefox 128

**Smartphone (please complete the following information):**
n/a

**Additional context**
n/a

martylukyy:
This PR fixes the filter dropdowns being cutoff when there is not enough space towards the bottom edge of the viewport.

There is still *some* weird behaviour when there isn't enough space towards the bottom and the top.
In this case a scrollbar will appear for the dropdown but
you still won't be able to scroll to all the entries on viewports with very low height.
However, the case that there isn't enough space towards the bottom and the top feels very unprobable.

fixes #1597

### DIFF
--- a/web/src/screens/filters/List.tsx
+++ b/web/src/screens/filters/List.tsx
@@ -396,7 +396,8 @@ const FilterItemDropdown = ({ filter, onToggle }: FilterItemDropdownProps) => {
         leaveTo="transform opacity-0 scale-95"
       >
         <Menu.Items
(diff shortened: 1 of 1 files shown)


zze0s (approved):
:+1:

PR #1601, opened 2024-07-26
fix(web): root pending component placement

martylukyy:
This PR centers the pending component on the RootRoute vertically again.

I don't know which change I made, caused that regression (and i couldn't be bothered to check) 
but this PR fixes it again and the placement of the pending component in other places such as:
- Dashboard
- Filters
- Release
- Submenus of Settings

are still centered aswell.


zze0s (approved):
:+1:

PR #1604, opened 2024-07-29
refactor(http): implement bufio

KyleSanderson:
Pretty minor with most usages, fixed a resource leakage with the red client.


zze0s (commented):
Shouldn't Peek check for 1 instead of 0 to be able to check for the err?

From the [docs](https://pkg.go.dev/bufio#Reader.Peek)

> If Peek returns fewer than n bytes, it also returns an error explaining why the read is short. The error is [ErrBufferFull](https://pkg.go.dev/bufio#ErrBufferFull) if n is larger than b's buffer size.

As for the RED tests now failing it most likely need quite a bit of refactoring to work with the new style.

zze0s (approved):

PR #1610, opened 2024-08-02
fix(indexers): Lazy matching in PTP announce pattern (Resolves #1609)

### LINKED ISSUE #1609: Greedy matching of PTP announce will break if additional vars are added following the existing announce format
https://github.com/autobrr/autobrr/blob/6b3a4c5c63427c52720c1af61c0076cbf7b85b93/internal/indexer/definitions/ptp.yaml#L112

Currently this works with the current announce structure.

Given an announce like:

> That Movie [1972] by Some Director | x264 / Blu-ray / MKV / 1080p | 204371 | 964303 | That.Movie.1972.1080p.BluRay.FLAC.x264-GROUP | comedy, drama, romance

Will result in:

```yaml
year: "1972"
releaseTags: x264 / Blu-ray / MKV / 1080p
freeleech: ""
torrentId: "964303"
torrentName: "That.Movie.1972.1080p.BluRay.FLAC.x264-GROUP"
tags: "comedy, drama, romance"
```

However, if the announce format were augmented to add an additional var, such as IMDB ID, resulting in an announce like:

> That Movie [1972] by Some Director | x264 / Blu-ray / MKV / 1080p | 204371 | 964303 | That.Movie.1972.1080p.BluRay.FLAC.x264-GROUP | comedy, drama, romance | tt12345678

You would instead get an incorrect result:

```yaml
year: "1972"
releaseTags: 204371
freeleech: ""
torrentId: "That.Movie.1972.1080p.BluRay.FLAC.x264-GROUP"
torrentName: "comedy, drama, romance"
tags: "tt12345678"
```

Assuming PTP maintained the existing announce structure of `var | var | var | var | var | var` they could continue adding additional vars in similar format (e.g. `var | var | var | var | var | var | var` and so on...) without breaking the existing pattern, if the pattern were to be adjusted to use lazy matching with ~~lookahead~~ optional check for additional `|` delimiter(s).

This would allow PTP to add additional parameters to their announce format, following the existing structure format, without requiring changes to the existing pattern. Currently, such a change to the announce format would require publishing new code and end users to update their client version in order to continue functioning correctly, regardless whether autobrr makes any use whatsoever of the new parameters.

doomtop:
Proposed solution to #1609.

Lazy matching is more resilient to changes in the announce format that might add additional vars to the tail of the announce.

### DIFF
--- a/internal/indexer/definitions/ptp.yaml
+++ b/internal/indexer/definitions/ptp.yaml
@@ -109,6 +109,30 @@ irc:
             torrentId: "964303"
             torrentName: That.Other.Movie.1972.1080p.BluRay.FLAC.x264-GROUP
             tags: comedy, drama, romance
+          - line: That Movie [1972] by Some Director | x264 / Blu-ray / MKV / 1080p | 204371 | 964303 | That.Movie.1972.1080p.BluRay.FLAC.x264-GROUP | comedy, drama, romance | tt12345678
+          expect:
+            year: "1972"
+            releaseTags: x264 / Blu-ray / MKV / 1080p
+            freeleech: ""
+            torrentId: "964303"
+            torrentName: That.Movie.1972.1080p.BluRay.FLAC.x264-GROUP
+            tags: comedy, drama, romance
+        - line: That Other Movie [1972] | x264 / Blu-ray / MKV / 1080p / Freeleech! | 204371 | 964303 | That.Other.Movie.1972.1080p.BluRay.FLAC.x264-GROUP | comedy, drama, romance | tt12345678
+          expect:
+            year: "1972"
+            releaseTags: x264 / Blu-ray / MKV / 1080p / Freeleech!
+            freeleech: Freeleech
+            torrentId: "964303"
+            torrentName: That.Other.Movie.1972.1080p.BluRay.FLAC.x264-GROUP
+            tags: comedy, drama, romance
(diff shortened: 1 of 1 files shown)


martylukyy (changes requested):
Hi!
First of all thanks for the PR!

The standard regex library from GoLang that we use does not support positive lookaheads.
Please check the regex expression with https://regex101.com/ and pick `Golang` as the regex flavour to confirm it's compatible with the standard Golang regex library .

internal/indexer/definitions/ptp.yaml
```diff
@@ -109,7 +109,31 @@ irc:
...
+            torrentName: That.Other.Movie.1972.1080p.BluRay.FLAC.x264-GROUP
+            tags: comedy, drama, romance
+          - line: That Movie [1972] by Some Director | x264 / Blu-ray / MKV / 1080p | 204371 | 964303 | That.Movie.1972.1080p.BluRay.FLAC.x264-GROUP | comedy, drama, romance | tt12345678 | foo | bar
+          expect:
+            year: "1972"
+            releaseTags: x264 / Blu-ray / MKV / 1080p
+            freeleech: ""
+            torrentId: "964303"
+            torrentName: That.Movie.1972.1080p.BluRay.FLAC.x264-GROUP
+            tags: comedy, drama, romance
```
martylukyy: Indentation on these tests is wrong.
Run it through a yaml linter or check it manually please. 🙂

martylukyy (changes requested):

martylukyy (approved):
LGTM now - thanks for putting in the work!
Up for ze0s now to decide if he wants to merge that change! 🙂

PR #1627, opened 2024-08-25
feat(diagnostics): add pprof profiling

KyleSanderson:
One of those things that should have been there at the start of development, and during the period of rapid iteration for the project. Better late than never.


zze0s (commented):
Good improvement!

Tracing to me has [other meanings](https://go.dev/blog/execution-traces-2024) so can we rename it to `profiling`?

Maybe rename the package to `diagnostics` so we can have both tracing and profiling in there.

Should probably add config for separate profiling host and port.

Thoughts?

zze0s (approved):

PR #1631, opened 2024-08-28
fix(web): enable browser support polyfills

fabricionaweb:
https://vitejs.dev/guide/build.html#browser-compatibility
https://github.com/vitejs/vite/tree/main/packages/plugin-legacy
https://browsersl.ist/#q=%3E+0.2%25%2C+iOS+%3E+12.2%2C+not+dead

This will allow Vite to generate polyfills that should work for browsers without support to some modern API.


martylukyy (approved):
This PR enables autobrr to be browsed from the Safari app down to iOS 12.2.
Not everything could be polyfilled by the newly added `@vitejs/plugin-legacy` plugin,
which is why there is some manual polyfilling done in `App.tsx`.

The old `browserslist` that seemingly did not have any function afaict,
has been replaced by the `@vitejs/plugin-legacy` `modernTargets`.

The node version for the codeql, eslint and release workflow has been unified to 20.10.0 and converted to a file so that we don't have to bump the same node version in 3 different files.

Many other changes in `vite.config.js` are due to linting.

Dark mode in Safari 12.2 and newer are still an issue since there no CSS polyfilling done.

This PR has been tested on an iPhone 6 with iOS 12.5.7.

**EDIT:
There is one caveat with this PR:
Every component that uses a `react-multi-select-component` will cause an unrecoverable error due to:**
```
TypeError: t.addEventListener is not a function. (In 't.addEventListener("change",s)', 't.addEventListener' is undefined)
```

martylukyy:
On a closer look there are more components that use EventListeners that are not supported by iOS 12.5.7
Refactoring all these components would mean to get rid of every headlessUI Listbox, 
every SlideOver which use headlessUI and trigger errors, every react-multi-select-component and potentially even more.
To make autobrr backwards compatible these components would need to be refactored to react-select
and for the SlideOver i haven't looked into a replacement due to the scope that refactor would be.
Feel free to close this issue due to these circumstances.

zze0s:
Valiant effort!! Making the UX and codebase worse to support older browsers is not ideal, so yeah lets leave it alone.

PR #1638, opened 2024-08-29
fix(web): remove unused `checkForUpdates` JSON object

martylukyy:
This PR remove the `checkForUpdates` JSON object from the `autobrr_settings` key
which seems not to be used.

https://github.com/autobrr/autobrr/blob/34d6e0cf609e225fa22261177d4b43c258f03566/web/src/screens/settings/Application.tsx#L141

The checkbox for it just writes to the config file upon triggering but doesn't change the JSON object.
So i feel list this can be removed?

### DIFF
--- a/web/src/utils/Context.ts
+++ b/web/src/utils/Context.ts
@@ -8,7 +8,6 @@ import { newRidgeState } from "react-ridge-state";
 
 interface SettingsType {
   debug: boolean;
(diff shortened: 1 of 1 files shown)


zze0s (approved):
Looks to be unused although the config key is named like that, the json key is `check_for_updates`.

PR #1703, opened 2024-09-02
fix(irc): reintroduce security for announces

KyleSanderson:
big sigh.

this entire handler thing should be rewritten to glue announcers to channels. this makes it a bit more secure for single channel networks.

https://github.com/autobrr/autobrr/pull/1322


zze0s:
Superseded by #1709 with a lot more changes and improvements.

PR #1716, opened 2024-09-12
feat(filters): wildcard slice matching optimizations

KyleSanderson:


zze0s (approved):
Looks like the tests still pass so LGTM!

PR #1720, opened 2024-09-12
feat(releases): show indexer name in indexer filter

martylukyy:
While reviewing #1706 i missed that the indexer filter on the releases page was still showing the identifier
instead of the name now.

This PR adds the required changes to make the options and the current filter value show the name.

### DIFF
--- a/web/src/api/queries.ts
+++ b/web/src/api/queries.ts
@@ -134,10 +134,14 @@ export const ReleasesIndexersQueryOptions = () =>
   queryOptions({
     queryKey: ReleaseKeys.indexers(),
     queryFn: async () => {
-      const response = await APIClient.indexers.getAll();
-      return response.map((indexer: IndexerMinimal) => ({
-        name: indexer.name,
-        identifier: indexer.identifier
+      const indexersResponse: IndexerDefinition[] = await APIClient.indexers.getAll();
+      const indexerOptionsResponse: string[] = await APIClient.release.indexerOptions();
+      
+      const indexersMap = new Map(indexersResponse.map((indexer: IndexerDefinition) => [indexer.identifier, indexer.name]));
+      
+      return indexerOptionsResponse.map((identifier: string) => ({
+        name: indexersMap.get(identifier) || "Unknown",
+        identifier: identifier
       }));
     },
     refetchOnWindowFocus: false,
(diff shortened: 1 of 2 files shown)


web/src/screens/releases/ReleaseFilters.tsx
```diff
@@ -69,17 +69,22 @@ export const IndexerSelectColumnFilter = ({
...
+    ? data.find(indexer => indexer.identifier === filterValue)?.name 
+    : "Indexer";
+
   // Render a multi-select box
   return (
     <ListboxFilter
       id={id}
       key={id}
-      label={filterValue ?? "Indexer"}
+      label={currentIndexerName ?? "Indexer"}
```
zze0s, line 82: You can skip the conditional here since it's done above in `currentIndexerName`.

martylukyy:
In hindsight this solution feels very inefficient and dirty compared to (maybe) adding the indexer name in the query on the backend in order for the frontend to only query `APIClient.release.indexerOptions();` and not mapping the data from 2 queries together.

Up for grabs.

web/src/screens/releases/ReleaseFilters.tsx
```diff
@@ -69,18 +69,23 @@ export const IndexerSelectColumnFilter = ({
 }: FilterProps<object>) => {
   const { data, isSuccess } = useQuery(ReleasesIndexersQueryOptions());
 
+  const options = isSuccess && data.map((indexer, idx) => (
+    <FilterOption key={idx} label={indexer.name} value={indexer.identifier} />
+  ));
```
zze0s, line 74: I'd move this back to render in the component like previously. It's no reused anywhere so it makes more sense to render like before.

web/src/api/queries.ts
```diff
@@ -133,8 +133,18 @@ export const ReleasesStatsQueryOptions = () =>
...
-    queryFn: () => APIClient.release.indexerOptions(),
-    placeholderData: keepPreviousData,
+    queryFn: async () => {
+      const indexersResponse: IndexerDefinition[] = await APIClient.indexers.getAll();
+      const indexerOptionsResponse: string[] = await APIClient.release.indexerOptions();
+      
+      const indexersMap = new Map(indexersResponse.map((indexer: IndexerDefinition) => [indexer.identifier, indexer.name]));
+      
+      return indexerOptionsResponse.map((identifier: string) => ({
+        name: indexersMap.get(identifier) || "Unknown",
```
zze0s, line 143: This should probably be `indexersMap.get(identifier) || identifier` so name becomes indexer identifier instead of Unknown, so we can differentiate if there's multiple indexers in the list without name.

web/src/screens/releases/ReleaseFilters.tsx
martylukyy: Done.

web/src/api/queries.ts
martylukyy: Done.

zze0s (approved):

PR #1721, opened 2024-09-12
build: add web/dist tarball

fabricionaweb:
PR still under discussion

Checking if goreleaser would add web/dist to the source tarball.

Grabbing an `autobrr_src.tar.gz` file (instead of the git repository tar.gz) would have the web/dist included that can be used to just build the go app later.

Just to help to build on some platforms that present problems within nodejs.
Since web/dist is just web files it is universal anyway we could provide the artifacts.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -137,8 +137,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Create web-dist artifact
-        working-directory: web/dist
-        run: tar czf ../../dist/web-dist.tar.gz ./*
+        run: tar czf dist/web-dist.tar.gz --directory=web/dist ./
 
       - name: Upload assets
         uses: actions/upload-artifact@v4


.goreleaser.yml
```diff
@@ -89,6 +90,12 @@ release:
     - Read the [documentation](https://autobrr.com)
     - Join our [Discord server](https://discord.gg/WQ2eUycxyT)
 
+source:
+  enabled: true
+  name_template: '{{ .ProjectName }}_{{ .Version }}_src'
```
zze0s, line 95: I'd change this to `{{ .ProjectName }}_{{ .Version }}_web_dist` so it's clear it's only web and only the dist dir.

.github/workflows/release.yml
```diff
@@ -136,6 +136,10 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Create web-dist artifact
+        working-directory: web/dist
+        run: tar czf ../../dist/web-dist.tar.gz ./*
```
zze0s, line 141: Can we change this around so the working-dir is default and we call `tar czf dist/web-dist.tar.gz ./web/dist/`? Looks clearer to me

zze0s (approved):

PR #1723, opened 2024-09-13
feat(notifications): Telegram set sender

martylukyy:
This PR adds a conditional sender payload for telegram for the message to say which app the notification is coming from.

Fixes #1722


zze0s (approved):
LGTM!

mizifih:
> Dear @zze0s
> 
> I can understand that some people use a single bot for everything, but others, like me, use multiple bots for different purposes. Both sides have the right to set up their automation systems in the way they prefer, but this change affects people like me. I would appreciate it if this new feature, which fortunately is only in development for now, could be manually adjusted.

I agree, making it optional should be better for everyone.

PR #1724, opened 2024-09-14
fix(indexers): new baseURL for Fuzer

martylukyy:
Old URL just displays a white page while new URL shows login form.

**WARNING:**
This PR contains database migrations. You **cannot** go back to 1.46.x once you run a version with this fix


zze0s (commented):
Lets hold off for a couple of days until there's more info about this.

zze0s (approved):

PR #1726, opened 2024-09-16
feat(notifications): optional Telegram sender

martylukyy:
**WARNING:**
This PR contains database migrations. You cannot go back to 1.46.x or earlier once you run a version with these changes.

Since there were complaints about the hardcoded autobrr sender in the Telegram notifications
it's now an optional field in the Telegram form fields.

If that field is left empty, there will not be an additional line with a sender.
If you enter something in this field, there will be an additional line with the string you entered.

ref https://github.com/autobrr/autobrr/pull/1723

### DIFF
--- a/internal/database/notification.go
+++ b/internal/database/notification.go
@@ -30,7 +30,7 @@ func NewNotificationRepo(log logger.Logger, db *DB) domain.NotificationRepo {
 
 func (r *NotificationRepo) Find(ctx context.Context, params domain.NotificationQueryParams) ([]domain.Notification, int, error) {
 	queryBuilder := r.db.squirrel.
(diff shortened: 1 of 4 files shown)


zze0s (changes requested):
This is a well done PR and shows all the steps needed to add a field from the database to frontend.

However, we already have a lot of fields for notifications that are used for different things. I suggest we simplify this and use the existing `topic` field for example. The frontend can keep calling it sender but will use the `topic` field.

You should end up with changes in just 2 files then:
- `internal/notification/telegram.go`
- `web/src/forms/settings/NotificationForms.tsx`

zze0s (approved):
Great! :partying_face:

PR #1730, opened 2024-09-17
fix(build): run web-dist tarball step for tags only

fabricionaweb:
After merge the #1721 we notice it that some steps doesnt run on merge event and pipeline failed because was missing "dist" folder (created by those skipped steps).

This is to fix it but also ensure to run on tags creation only.


zze0s (approved):
Lets hope this works :rocket:

PR #1738, opened 2024-09-22
feat(indexers): update PTP announce to new format

zze0s:
Update PTP to use new announce format.

Regex playground: https://regex101.com/r/8DvvNK/1


internal/indexer/definitions/ptp.yaml
```diff
@@ -93,31 +64,53 @@ irc:
...
+            imdb: "tt24249072"
+            torrentSize: "1899875175"
+            releaseTags: H.264/MKV/WEB/720p/FL
+            freeleech: "FL"
+            year: "2023"
+            title: "Last Straw"
+            torrentName: Last.Straw.2023.REPACK.720p.AMZN.WEB-DL.DDP5.1.H.264-FLUX
+            tags: thriller,horror
+        # <boolnew>:<category>:<gid>:<tid>:<key>:tt<imdbid]>:<sizebytes>|<format>/<container>/<source>/<resolution>(/S)?(/(FL/HL/NL)?|<year>|<title>|<releasename>|<taglist>
+        pattern: '(\d):(\d):(\d+):(\d+):(.+):(tt?\d+):(\d+)\|(.*?(?:\/(FL|HL|NL))*?)\|(\d+)\|(.+)\|(.+)\|(.+)'
```
martylukyy, line 98:
```suggestion
        pattern: '(\d):(\d):(\d*):(\d*):(.*):(tt\d*):(\d*)\|(.*?)\/?(FL|HL|NL)?\|(\d*)\|(.*)\|(.*)\|(.*)'
```
I would suggest to replace all the + quantifiers with * in case on of the var happens to be missing.
Furthermore i think there is no IMDB ID that doesn't begin with tt so making the second t optional is not needed here?
The non-capturing group isn't needed aswell here afaict so i got rid of the nesting/non-capturing group.
Feel free to point out in case i have a mistake in my logic!

nuxencs: those changes look good, but having the * quantifier for the IMDb ID probably doesn't make sense, because if there is no id it won't match either way

martylukyy: Good point! Thanks nuxencs!
Edited the group optional in my suggestion!

The new announce line is already being used in their channel and ze0s any myself are gathering announces.

nuxencs: Ahh it's actually announcing twice, once with the old and once with the new format

nuxencs: looking at a few announnces, we could probably restrict the `key` matching to alphanumeric with `\w*`

martylukyy: Since key *should* also be there i replaced it with `\w+`

nuxencs: I'm guessing the same could be said for `gid` and `tid`? 🤔

martylukyy: `tid` yes, `gid` i am actually not sure but probably yes...
However having * here makes it more failsafe imo.
Open to discussion here though in case i didn't think about something here!

Having + for the key is actually beneficial since it seems like we need it for URL building now.

internal/indexer/definitions/ptp.yaml
```diff
@@ -93,31 +64,113 @@ irc:
...
+          - title
           - torrentName
           - tags
 
+    mapCustomVars:
+      announceType:
+        "0":
+          announceType: NEW
+        "1":
+          announceType: PROMO
```
nuxencs, line 142: shouldn't his be the other way around? 🤔

internal/indexer/definitions/ptp.yaml
```diff
@@ -93,31 +64,113 @@ irc:
...
+            imdb: "tt0107426"
+            torrentSize: "63996413804"
+            releaseTags: BD66/m2ts/Blu-ray/2160p
+            editionTags: "Dolby Atmos/Dolby Vision"
+            year: "1993"
+            title: "Little Buddha"
+            torrentName: Little.Buddha.1993.2160p.FRA.UHD.Blu-ray.DV.HDR.HEVC.DTS-HD.MA.5.1
+            tags: drama,italian
+        # <boolnew>:<enum_fl>:<boolscene>:<category>:<gid>:<tid>:<key>:tt<imdbid>:<sizebytes>|<format>/<container>/<source>/<resolution>|<edition_tags>|<year>|<title>|<releasename>|<taglist>
+        pattern: '(\d):(\d):(\d):(\d):(\d+):(\d+):(.+):(tt?\d+):(\d+)\|(.*)\|(.*)\|(\d+)\|(.+)\|(.+)\|(.+)'
```
nuxencs, line 119:
```suggestion
        pattern: '(\d):(\d):(\d):(\d):(\d+):(\d+):(\w+):(tt\d+)?:(\d+)\|(.*)\|(.*)\|(\d+)\|(.+)\|(.+)\|(.+)'
```
Updated regex example with some change suggestions 😄 
https://regex101.com/r/8DvvNK/3

internal/indexer/definitions/ptp.yaml
zze0s: Should it? :thinking:

nuxencs: at least in the announce it's 1 for new stuff, haven't seen one that is not 1 to compare tho 😄

martylukyy: Agree with nuxencs here.

zze0s: It's about the be changed to something like this so the above is correct for now.

0 - new - only on fresh upload
1 - freeleeched
2 - reseeded
3 - marked golden popcorn
4 - checked ... ?

PR #1739, opened 2024-09-24
fix(filters): release download counts

cascandaliato:
When counting the number of recent downloads to enforce the `Max Downloads` rule on a filter, multiple actions attached to the same filter will be counted as multiple downloads.

This PR changes the logic to count multiple actions as a single download.

### DIFF
--- a/internal/database/filter.go
+++ b/internal/database/filter.go
@@ -1320,13 +1320,12 @@ func (r *FilterRepo) GetDownloadsByFilterId(ctx context.Context, filterID int) (
 
 func (r *FilterRepo) downloadsByFilterSqlite(ctx context.Context, filterID int) (*domain.FilterDownloads, error) {
 	query := `SELECT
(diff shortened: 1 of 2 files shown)


cascandaliato:
Currently discussing an alternative query on Discord:
```sql
SELECT
  COUNT(DISTINCT CASE WHEN CAST(strftime('%s', datetime(timestamp, 'localtime')) AS INTEGER) >= CAST(strftime('%s', datetime('now', 'localtime', 'start of day')) AS INTEGER) THEN release_id END) as "day_count"
FROM release_action_status where status in ('PUSH_APPROVED', 'PUSH_PENDING') and filter_id = ?;
```

cascandaliato:
SQL queries updated as discussed with @KyleSanderson on Discord ([link](https://discord.com/channels/881212911849209957/881216126980341780/1288124730573983776)).

zze0s (approved):
Thanks for the PR! This fixes a long standing bug with the filter counting.

The additional tests should make any refactoring in the future way more reliable :partying_face:

PR #1751, opened 2024-10-03
feat(filters): sort by created and updated

martylukyy:
This PR will implement sorting filters by date.
The 4 options will be:
- Recently created first
- Recently created last
- Recently updated first
- Recently updated last

The react query devtools show that the field param is getting submitted,
but checking with the debugger in GoLand in database/filter.go:75 shows that the param is missing...
Not sure what is going on here 🤔


martylukyy (commented):
Works like a charm!
Thanks for jumping in and implementing the missing part in the backend!
Appreciate you!

(I cannot approve my own PR but take that comment as an approval.)

zze0s (approved):

PR #1761, opened 2024-10-07
feat(download-client): add label to Deluge if it does not exist

KyokoMiki:
This feature was previously in a TODO state. I set out to complete it, but considering that labels generally only need to be added once, checking every time would consume extra time. Therefore, the processing logic I used is to set the label directly unless an exception with rpcErr.ExceptionMessage == 'Unknown Label' is returned. I don't have a Deluge 1.x client, but since there's no difference in the interface in go-libdeluge, I believe it should also work on Deluge 1.x. I would be really grateful if someone could help me test it.

### DIFF
--- a/internal/action/deluge.go
+++ b/internal/action/deluge.go
@@ -137,10 +137,17 @@ func (s *service) delugeV1(ctx context.Context, client *domain.DownloadClient, a
 			}
 
 			if labelPluginActive != nil {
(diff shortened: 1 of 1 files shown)


zze0s (approved):
Hey! Thanks for the PR! This is a great improvement!

I broke it out into a separate function since the same block was called 4 times. Makes it a bit cleaner.

I tested on Deluge v1.3.15 and it works the same :partying_face:

PR #1765, opened 2024-10-12
fix(wildcard): partial match ignoring consumed pattern

KyleSanderson:


zze0s (approved):
The wildcard package has crossed my threshold for easy reviewing so I can just look at the tests to see if this fixes something, and this appears to solve the new cases.

martylukyy:
At the risk of making myself unpopular with this comment, 
we should really add all the test cases that have resulted in wrongful matching behaviour.

PR #1771, opened 2024-10-19
fix(indexers): PTP API test function

s0up4200:
Fix for PTP api returning ID as int on browse.php endpoint and string on Get by ID.

### DIFF
--- a/pkg/ptp/ptp.go
+++ b/pkg/ptp/ptp.go
@@ -104,6 +104,33 @@ type Torrent struct {
 	RemasterYear  *string `json:"RemasterYear,omitempty"`
 }
 
+// custom unmarshal method for Torrent
+func (t *Torrent) UnmarshalJSON(data []byte) error {
+	type Alias Torrent
+
+	aux := &struct {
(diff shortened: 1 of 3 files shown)


zze0s (approved):
:rocket:

pkg/ptp/ptp.go
```diff
@@ -104,6 +104,33 @@ type Torrent struct {
...
+		Alias: (*Alias)(t),
+	}
+
+	if err := json.Unmarshal(data, &aux); err != nil {
+		return err
+	}
+
+	switch id := aux.Id.(type) {
+	case float64:
+		t.Id = fmt.Sprintf("%.0f", id)
```
zze0s, line 124: Shouldn't this test as int64 instead of float64?

s0up4200: Oops. Yeah definitely. Will fix!

s0up4200: Correction: The numbers are parsed as floats when marshalled so it should stay float

zze0s (approved):

PR #1774, opened 2024-10-19
fix(web): remove unnecessary link to Logs

s0up4200:
The link linked to the page you're already viewing, which seemed unnecessary.

Old:
image

New:
image


martylukyy (changes requested):
Since there are no JSX components used in the description anymore  
you can get rid of the curly brackets and the empty JSX tag/fragment aswell
and replace it all with two double quotes. ( "..." )

martylukyy (approved):

zze0s (approved):

PR #1777, opened 2024-10-20
feat(http): set ResponseHeaderTimeout

KyleSanderson:


zze0s (commented):
Can this potentially mess with external filters running webhooks such as cross-seed?

zze0s (approved):

PR #1812, opened 2024-11-06
build(ci): implement PGO

KyleSanderson:
default.pgo needs to be along side the unit being compiled at the time of this writing, which isn't great for general profiles being created. The file is intentionally named cpu.pprof, to denote that it needs to be statically added on the command as defualt.pgo isn't read from the current (present working) directory.

Committing binary profiling data to a git repository is just crazy (which is the present recommendation from the go team), hence the run-time generation as part of the pipeline. Tests are used as a kludge now to try and get more inlining, but this is the wrong approach in the future once more e2e comes into the tree.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -157,7 +157,7 @@ jobs:
         uses: test-summary/action@v2
         with:
           paths: "unit-tests.xml"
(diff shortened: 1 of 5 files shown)


zze0s (approved):
Not a fan of the increased build times but if this improves performance then there's no reason to hold it back.

Well done :+1:

PR #1819, opened 2024-11-16
feat(http): add error logging for API responses

s0up4200:
Adds proper error logging with stack traces for API error responses.

**Before:** API errors (like 500 responses) were not logged
**After:** All API errors are logged with stack traces and context

Testing:
Verified by attempting to create a second user when onboarding is unavailable:

```curl
curl -i -X POST http://localhost:7474/api/auth/onboard -H "Content-Type: application/json" \
  -d '{"username": "testuser2", "password": "test123"}'
```

This produced the expected error log with stack trace:

```shell
2024-11-16T12:38:27+01:00 ERR server error error="onboarding unavailable" stack=[...] status=503
```

### DIFF
--- a/internal/http/encoder.go
+++ b/internal/http/encoder.go
@@ -6,9 +6,19 @@ package http
 import (
 	"encoding/json"
 	"net/http"
+
+	"github.com/autobrr/autobrr/internal/logger"
 )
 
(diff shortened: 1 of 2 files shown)


zze0s (approved):
:+1:

PR #1822, opened 2024-11-17
feat(cache): implement TTLCache and TimeCache

KyleSanderson:
This is a tricky scenario, and one I find rather ridiculous coming to the end of 2024.

Inventing a cache doesn't make sense, however there doesn't appear to be an equivalent out there for the features desired.

The differences here is there's no fixed item limit - this uses a regular golang map taking a global RW lock. Others use a lock free implementation, which is fine, but because the objects are commonly stored in a doubly linked list the elements are not stored together, which is problematic for cache coherency. The lookups are tricky on the non-map implementations, because all of the memory is allocated upfront certain types (unlike comparable) are just not permitted to be used as keys. I had concerns about the GC, however this doesn't appear to even be a factor compared to the other implementations.

Other implementations update the object on every single read (commonly, this is a unique feature, others have no option for this), which was the default for many filesystems up until RHEL6~. xfs as an example uses a relative atime (relatime) updating scheme, which is similar to what has been implemented here with timecache. This means that our insertion latency is only hit on the worst case, which can be hundreds or thousands of cache hits later, before the next time duration is hit (which would mean 1 object replacement, on 1000 reads, as opposed to tit for tat).

Right now, from this minor mindset shift, this is roughly 10x faster on our real wildcard workload for looking up cached items compared to https://github.com/jellydator/ttlcache . Other speedups are possible, but more risky, as they would replace the map implementation with something more fine grained. For thousands of items this is fast enough, but this is the beginning, not the end.

Unification of expiration of items is nice, and we can do more work here around database caching and similar to help insulate from hits as installations continue to grow in size. When the database query comes in, we can hit our cache if there hasn't been an invalidation event (an update or delete) on an item. This should let us get tighter on lookups and reduce the parsing cost of items.


zze0s (approved):
This is a big change. Lets see how it goes! :+1:

PR #1830, opened 2024-11-23
feat(indexers): REDACTED update urls

zze0s:
Update urls for RED.


fa1rid:
I think the regex pattern is wrong. `/` isn't escaped
Should be like this: `((.*) (?:\[(.*)\] \[(.*)\] - (.*))?) - .*id=(.*) \/ (https?:\/\/.+\/).+id=(\d+)[ -]*(.*)`

PR #1831, opened 2024-11-23
fix(build): Dockerfile windows entrypoint

zze0s:
Fix lint warning for entrypoint.


zze0s:
Just tested on Windows with Docker Desktop set to Windows container mode and it all works as intended.

I'm sure the few users will report issues if there are any.

PR #1863, opened 2024-12-06
feat(macros): add Description

nsenica:
Some trackers add information about the release in the description. This change would allow the description to be available for Exec action and take decisions based on that.


zze0s (approved):
Hey! Thanks for the PR!

PR #1895, opened 2024-12-26
fix(database): convert NULL hybrid values to FALSE in existing releases

s0up4200:
Can likely be closed.

https://discord.com/channels/881212911849209957/882310361943011399/1321736192475271209


s0up4200:
Closed as it would only affect some users who ran develop at the time of merging.

PR #1898, opened 2024-12-26
feat(lists): ui enhancements

fabricionaweb:
- Make the page look more similar to the others, if you switch between pages you can notice it.
- I could not 100% fix the overflow when big filters exists, but it looks better.
- Changed the types to a map labels.
- That `browserlist` is not used anymore (since webpack)

image

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -13,18 +13,6 @@
     "preview": "vite preview",
     "lint:watch": "pnpm run lint -- --watch"
   },
-  "browserslist": {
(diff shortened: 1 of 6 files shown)


martylukyy:
How do we feel about changing it to show the number of filters it is applied to if it is applied to more than 1 filter
and then display a popup/tooltip on click/hover/mouseover?

zze0s (approved):

PR #1906, opened 2024-12-27
fix(database): migrations to update PTP IRC channel

zze0s:
Add conditional check to PTP irc channel update to not cause fatal `UNIQUE CONSTRAINT` error during migrations.


s0up4200:
For the web changes in https://github.com/autobrr/autobrr/pull/1906/commits/0783dbac3e5a8933287756b1f1bf063636e8b84c
CleanShot 2024-12-27 at 20 40 01@2x

PR #1908, opened 2024-12-28
feat(auth): add option to disable built-in login when using OIDC

s0up4200:
Re https://github.com/autobrr/autobrr/discussions/1907

Adds the option to disable the built-in login form when a user exists in the db and OIDC is enabled.
Defaults to false.

```toml
oidc_disable_built_in_login = true
```

```env
AUTOBRR__OIDC_DISABLE_BUILT_IN_LOGIN=true
```


s0up4200:
A cleaner solution would probably be to simply hide the login form if OIDC is enabled. Since simply setting `AUTOBRR__OIDC_ENABLED` to `false` will show it anyway.

internal/domain/config.go
```diff
@@ -4,37 +4,38 @@
...
+	ProfilingEnabled        bool   `toml:"profilingEnabled"`
+	ProfilingHost           string `toml:"profilingHost"`
+	ProfilingPort           int    `toml:"profilingPort"`
+	OIDCEnabled             bool   `mapstructure:"oidc_enabled"`
+	OIDCIssuer              string `mapstructure:"oidc_issuer"`
+	OIDCClientID            string `mapstructure:"oidc_client_id"`
+	OIDCClientSecret        string `mapstructure:"oidc_client_secret"`
+	OIDCRedirectURL         string `mapstructure:"oidc_redirect_url"`
+	OIDCScopes              string `mapstructure:"oidc_scopes"`
+	OIDCDisableBuiltInLogin bool   `mapstructure:"oidc_disable_built_in_login"`
```
zze0s, line 38: I just saw that these uses `mapstructure` and snake case instead of toml and CamelCase like all the other ones. That's unfortunate that we already have users with configs for it.

You can add the correct variant and keep the existing as well.

```golang
OIDCEnabled             bool   `toml:"oidcEnabled" mapstructure:"oidc_enabled"`
OIDCIssuer              string `toml:"oidcIssuer" mapstructure:"oidc_issuer"`
OIDCClientID            string `toml:"oidcClientId" mapstructure:"oidc_client_id"`
OIDCClientSecret        string `toml:"oidcClientSecret" mapstructure:"oidc_client_secret"`
OIDCRedirectURL         string `toml:"oidcRedirectUrl" mapstructure:"oidc_redirect_url"`
OIDCScopes              string `toml:"oidcScopes" mapstructure:"oidc_scopes"`
OIDCDisableBuiltInLogin string `toml:"oidcdisableBuiltInLogin" mapstructure:"oidc_disable_built_in_login"`
```

zze0s (approved):

PR #1909, opened 2024-12-28
feat(releases): save early to improve skip duplicates matching

zze0s:
This is an experimental pr to improve the matching for skip duplicates with releases that are announced within seconds of each other.

* Store the release as early as possible and change the checks for duplicates to not look at Release Action Statuses.


kenstir:
This PR will help.  Is the only thing missing the startup check that @KyleSanderson suggests?  I dug deep into why my filter did not skip dups or download only 1 episode per day, and it was due to the delay caused by setting status=Approved only after [ReannounceTorrentWithRetry](https://github.com/autobrr/go-qbittorrent/blob/b355f45903eac9ae5b2a63cf2968d2347f1c888a/methods.go#L1318) returns.

zze0s:
Coming back to this after some time and thinking:

We could add a new option to client actions with something like "Wait for re-announce" and make it opt-in, or opt-out, and let re-announce run in the background without blocking/waiting.

That should improve things when it doesn't have to wait. And the goal is still to find a better solution for in-flight items but that will take a lot longer still.

zze0s:
Will be replaced by something else in the future.

PR #1914, opened 2024-12-31
feat(lists): scheduler to be configurable

luckylittle:
#### What is the relevant ticket?
* feature(lists): scheduler to be configurable

##### Add
* new configuration `schedule`
##### Change
* none
##### Remove
* none

* `cmd/autobrr/main.go`
* `config.toml`
* `internal/config/config.go`
* `internal/domain/config.go`
* `internal/list/service.go`

* `go build -o bin/autobrr cmd/autobrr/main.go && bin/autobrr`

#### Risk involved?
* no

#### Screenshots (if appropriate):
* n/a

* Yes

### DIFF
--- a/cmd/autobrr/main.go
+++ b/cmd/autobrr/main.go
@@ -142,7 +142,7 @@ func main() {
 		releaseService        = release.NewService(log, releaseRepo, actionService, filterService, indexerService)
 		ircService            = irc.NewService(log, serverEvents, ircRepo, releaseService, indexerService, notificationService, proxyService)
 		feedService           = feed.NewService(log, feedRepo, feedCacheRepo, releaseService, proxyService, schedulingService)
-		listService           = list.NewService(log, listRepo, downloadClientService, filterService, schedulingService)
+		listService           = list.NewService(log, listRepo, downloadClientService, filterService, schedulingService, cfg.Config)
 	)
 
 	// register event subscribers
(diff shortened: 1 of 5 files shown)


zze0s:
Hi! Thanks for the PR! Happy new years!

When I built the lists integration it was an active decision to not include a user configurable schedule because our own lists for example are refreshed every 24h, so having a user refresh every 1 hour would really benefit much.

The ARRs have webhooks you can setup for to tell autobrr when something is added or removed. https://autobrr.com/filters/lists#automatic-updates-via-webhook

Other lists might be updated more frequently than 6 hours. You could setup a cronjob to have it update more often. Here's the available methods.

* `https://autobrr.mydomain.com/api/webhook/lists/trigger` - Trigger all
* `https://autobrr.mydomain.com/api/webhook/lists/trigger/arr` - Trigger arr only
* `https://autobrr.mydomain.com/api/webhook/lists/trigger/lists` - Trigger non-arr lists only
* `https://autobrr.mydomain.com/api/webhook/lists/trigger/{listID}` - Trigger by listID

The Feeds for example also runs on a schedule but they have the schedule config as part of the config in the db, per feed. Not a global one in the config.

You obviously see the need for this since you implemented it, but what is the primary reason to have lists update more often than every 6 hour?

luckylittle:
Hi,

I personally don't have the need for <6hrs, I just don't feel like this value should be hard coded. Perhaps some people are using external sources that gets updated more frequently and would find it helpful?

I will look into the webhooks method, thanks

PR #1930, opened 2025-01-05
feat(metrics): add metrics server

Intuinewin:
Hi, this is a PR to add a metrics server to autobrr.

I wanted this in order to track autobrr stats in a grafana dashboard and to be notified with my existing alerting stack when an irc server is down or when a feed could not be fetched.

Currently it exposes basic metrics and can be improved (for example by returning accepted/rejected releases **per indexer**).

Let me know what you think of this PR and if some parts should be refactored. I'm open to suggestions

---
Implementation details

The metrics server can be enabled and configured with 3 new config parameters (`metricsEnabled`, `metricsHost` and `metricsPort`). It only serves a /metrics route that returns the metrics in prometheus format.

The metrics server relies on a metrics manager that stores a prometheus registry in charge of collecting metrics from multiple collectors.
I put the collectors in their own package to group them together but maybe they should be moved elsewhere?

---

Quick question: I found out that the release stats endpoint returns the number of releases rejected by filters(`filter_rejected_count`) but they are actually not saved in db.
I created a metric for this value but should this value be removed or the releases rejected by filters saved in the db?


zze0s:
Hi! Thanks for the PR! Nicely done!

Just yesterday us maintainers talked about that not a single person had asked about metrics yet even tho there's lots of users with homelab setups that monitor everything :smile: 

Looks like you cover most useful metrics!

The `filter_rejected_count` is not used at this time because we don't save all the rejected releases in the db, due to it would quickly grow to a huge number for users with a few general indexers setup etc. It could of be mitigated with scheduled cleanup jobs but we're not there just yet.

Would a counter/histogram/summary for announces maybe be interesting? Could be global and/or per channel? Could make for some interesting views of when there's more or less releases.

The code looks good!

thezak48:
> Just yesterday us maintainers talked about that not a single person had asked about metrics yet


Lies 🤣

zze0s:
Skip that extra stuff and leave it as is, we can extend it later on. Will make a proper review in coming days :+1:

zze0s:
The benefit of this is you can bind it to a separate ip and port only available to your monitoring stack etc. Could add the existing auth middleware on the handler.

onedr0p:
> Does this actually have to sit on another IP and Port? Can't we protect this with the existing API key mechanism (perhaps add scoping to keys) and give clients a nice URL to use from the WebUI that they can copy into their Prometheus config. The & and other characters might have to be encoded, but that's where the nice WebUI copy string can come in.

A different IP? Not necessarily but a different port is a pretty standard thing to do with apps that choose to implement prometheus metrics. I don't think that having metrics protected by the API key is the way to go as it complicates the scrapeConfigs to now contain a secret so I am all for the metrics being served up on another port as well.

zze0s (approved):
@Intuinewin Finally had some time to properly review and add optional Basic Auth to the handler.

* I also changed the port to `9074` to be in the regular prometheus range of 9000-9200
* Added a new command `htpasswd` to `autobrrctl` to easily generate a bcrypt hashed password to be used for basic auth

Thanks for the PR!! If you have any Grafana dashboards etc please share if you can :smile:

zze0s (approved):
Realized the Go runtime metrics etc was missing so added those so now it should cover both runtime metrics and everything else :+1:

Intuinewin:
Thanks ! Here is the grafana dashboard I made to test this PR:
[Autobrr-1737823851445.json](https://github.com/user-attachments/files/18547001/Autobrr-1737823851445.json)

PR #1940, opened 2025-01-14
docs(cli): add manpage

s0up4200:
### Description
Adds a manpage for both `autobrr` and `autobrrctl` commands, following standard Unix manpage conventions.

#### Installation
The manpage will be installed to `/usr/share/man/man1/` (or `/usr/local/share/man/man1/` with Make) via:
- Package managers (deb, rpm, apk, archlinux)
- Manual installation using `make install` or `make install-man`

#### Testing
- Verified with `mandoc -T lint` for proper formatting


zze0s (commented):
Can the man dir be moved into docs or does that break anything?

s0up4200:
> Can the man dir be moved into docs or does that break anything?

Shouldn't be a problem. Moved the man dir into a new docs dir now 👍

zze0s (approved):

Issue #1938, opened 2025-01-12
When RSS Generic Feed contains symbol "&" (without quotes) it is encoded with it's unicode codepoint \u0026


RioSif:
**Version**
v1.57.0

**Describe the bug**
When RSS Feed contains symbol "&" (without quotes) it is encoded with it's unicode codepoint \u0026

**To Reproduce**
1. Add a generic RSS Feed containing this symbol
2. For example signup and add the RSS of this https://oldgreektracker.xyz/signup.php
3. Make a filter for this feed
4. Check the result, you won't be able to download any torrent.

**Expected behavior**
The URL keeps having the symbol & and the download work as normal.

I run the below script to make it as expected to get the screenshot bellow.
`sqlite> last_run_data = replace (last_run_data, '\u0026', '&');`

**Screenshots**
https://imgur.com/a/OXx50f3

<img width="680" alt="Image" src="https://github.com/user-attachments/assets/ffabde60-5b93-42fb-acb2-ca9b3b76527b" />
<br/><br/>
<img width="690" alt="Image" src="https://github.com/user-attachments/assets/237906a2-42c5-4f48-9e31-567036583b61" />

PR #1942, opened 2025-01-17
fix(feeds): handle unicode escaped url characters

s0up4200:
Should solve #1938


pkg/sanitize/sanitize.go
```diff
@@ -12,6 +12,30 @@ func String(str string) string {
 	return str
 }
 
+func SanitizeURLEncoding(str string) string {
```
nuxencs, line 15:
```suggestion
func URLEncoding(str string) string {
```
I would avoid adding the package name to the function

internal/feed/rss.go
```diff
@@ -158,8 +159,8 @@ func (j *RSSJob) processItem(item *gofeed.Item) *domain.Release {
 			if parentURL, _ := url.Parse(j.URL); parentURL != nil {
 				parentURL.Path, parentURL.RawPath = "", ""
 
-				// unescape the query params for max compatibility
-				escapedUrl, _ := url.QueryUnescape(parentURL.JoinPath(rls.DownloadURL).String())
+				downloadURL := sanitize.SanitizeURLEncoding(rls.DownloadURL)
```
nuxencs, line 162:
```suggestion
				downloadURL := sanitize.URLEncoding(rls.DownloadURL)
```
I would avoid adding the package name to the function

nuxencs (changes requested):
just some nitpicking 😄

s0up4200:
> just some nitpicking 😄

good call! 👍

internal/feed/rss.go
```diff
@@ -158,8 +159,8 @@ func (j *RSSJob) processItem(item *gofeed.Item) *domain.Release {
 			if parentURL, _ := url.Parse(j.URL); parentURL != nil {
 				parentURL.Path, parentURL.RawPath = "", ""
 
-				// unescape the query params for max compatibility
-				escapedUrl, _ := url.QueryUnescape(parentURL.JoinPath(rls.DownloadURL).String())
+				downloadURL := sanitize.URLEncoding(rls.DownloadURL)
+				escapedUrl, _ := url.QueryUnescape(parentURL.JoinPath(downloadURL).String())
 				rls.DownloadURL = escapedUrl
 			}
 		}
```
zze0s, line 166: This is only used for relative links like `/details.php?=10`. The if above handles all links so it needs to be there as well.

I did some tests locally and now it does what is expected. The latest added test does not do anything. I'll probably commit my local changes to add an integration test for this.

zze0s (approved):

PR #1946, opened 2025-01-23
feat(web): upgrade to tailwind v4

martylukyy:
This PR updates tailwind to v4 and gets rid of PostCSS since there is now a vite plugin for tailwind.

Most of the upgrade has been done by the automated upgrade tool but it produced some errors.

Some of these error were missing colors in RSMC or react-select components which i fixed.

There are a couple of borders which have been removed or changed because of tailwind's new border behavior.
ref: https://tailwindcss.com/docs/upgrade-guide#new-default-border-color

Please check for double or misaligned borders during review and also for color mismatches in various selects 
in case i missed some of them.

PS: I also snuck in the removal of the table header hover animations since they are not sortable anymore.


zze0s (approved):
Thanks for putting in the work marty!!

s0up4200 (approved):
LGTM

Issue #1899, opened 2024-12-26
TorrentSeeds: IRC announce is shown two times in log


DigitGram:
**Version**
v1.56.0

**Describe the bug**
Just added "TorrentSeeds" **AND** "TorrentSeeds Music" in autobrr.

1. **Same IRC-Nickname is not possible**. You will get an error-message: `HTTP request to 'api/irc/network/6' failed with code 500 (error executing query: constraint failed: UNIQUE constraint failed: irc_network.server, irc_network.port, irc_network.nick (2067))`
2. if you activate in Settings-IRC "TorrentSeeds" **AND** "TorrentSeeds Music" **you will get each announce two times**. See log ([autobrr.log](https://github.com/user-attachments/files/18253630/autobrr.log)). If you deactivate just one of them, behavior is normal again (each announce one time)

**To Reproduce**
1. add "TorrentSeeds" and "TorrentSeeds Music" as Indexer
2. same IRC-nickname not possible
3. if both IRC are active, you will get each announce two times. Therefore matching releases are getting added two times.

**Expected behavior**
Same nickname on IRC-channels, single announce in log, therefore single add of a matching release.

**Screenshots**
![Image](https://github.com/user-attachments/assets/c7ee517f-42bb-4299-8b10-f859341d90cc)

**Additional context**
possible error: if you join irc.torrentseeds.org:6697: as soon as you identify yourself, you are forced to join several channels.
```
03:45:47 -irc.torrentseeds.org- *** You were forced to join #announce
03:45:47 -irc.torrentseeds.org- *** You were forced to join #announce-music
03:45:47 -irc.torrentseeds.org- *** You were forced to join #request
03:45:47 -irc.torrentseeds.org- *** You were forced to join #torrentseeds
03:45:47 -NickServ- You were forced to identify by Cerberus to account xxxxxxxxxxx
```

DigitGram:
I play a little with regex-pattern and try to merge "TorrentSeeds" and "TorrentSeeds Music" into one .yaml-file. :-)

![Image](https://github.com/user-attachments/assets/a30b8702-d915-45ba-b86b-198914dda95c)

DigitGram:
Merged "TorrentSeeds" and "TorrentSeeds Music". Working fine so far. 
Please test!

DigitGram:
Category: "E-Books" was not recognized by my regex-pattern. Fixed!
[...]

PR #1948, opened 2025-01-23
feat(indexers): merged torrentseeds.yaml and torrentseeds-music-yaml

DigitGram:
fixed bug, see https://github.com/autobrr/autobrr/issues/1899.


zze0s:
Hi! Removing the music one is probably not ideal. It will break for users currently using it. No idea how many there are but we try our hardest to not break backwards compatibility.

autobrr does not properly handle multi channel irc parsing like that in the combined config. It might work ok because they have their own irc network but there's no proper handling for it AFAIK.

Will get back to this.

zze0s:
I'll close this as this isn't a properly supported feature, even if it sort of works. 

We'll implement a proper solution for this in the future.

PR #1949, opened 2025-01-24
feat(lists): add anilist support

fabricionaweb:
This adds integration to AniList using our internal API.
Since AniList only have GraphQL I didnt add personal/custom urls to it.
For now just our internal apis (that refresh once a day)

- Current Season
- Next Season
- Trending


internal/list/process_list_anilist.go
```diff
@@ -0,0 +1,96 @@
...
+	l.Debug().Msgf("fetching titles from %s", list.URL)
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, list.URL, nil)
+	if err != nil {
+		return errors.Wrapf(err, "could not make new request for URL: %s", list.URL)
+	}
+
+	list.SetRequestHeaders(req)
+
+	resp, err := s.httpClient.Do(req)
```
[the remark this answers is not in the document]
martylukyy: Yes, CodeQL warned about this on all the other lists too.
So feel free to dismiss and ignore.

martylukyy (commented):
Code wise it looks fine to me but the actual output of the list isn't good.

Some examples:
`classic★stars`
`solo?leveling?2ª`
`bogus?skill?<<fruitmaster>>`
`iku?iku?♡?succubus`
`witchy?precure*~mirai?days~`

These will never match because of the ambiguous characters 
and ideally should be cleaned up...

In my opinion adding these characters to the ones to be filtered out shouldn't be a problem but i would love to get another opinion on this.

martylukyy (commented):
Okay there are still some gotchas here...

On the one hand we have this now, which is good:
`colorful?stage*the?movie*a?miku?who?can?t?sing`
`colorful?stage*the?movie*a?miku?who?cant?sing`
The Unicode Character `'` (U+0027) Apostrophe gets removed and due to that 
we have `can?t` and `cant` so regardless of how it's announced it will be matched.

But we also have this case:
`this?is?him*can?t?be?anyone?else`
`this?is?him*can?t?be?anyone?else?`
So you might ask yourself, why is this happening? Why is there `can?t` two times?
Don't worry, i have an answer.

It seems we only check for the good ol' Unicode Character `'` (U+0027) Apostrophe in title.go with the regexp `\p{P}`
https://github.com/autobrr/autobrr/blob/5e2769639fecea528951d478c0aa94a484f9b90c/internal/list/title.go#L16
but not for:
- `´` (U+00B4) Acute Accent 
- ``` ` ``` (U+0060) Grave Accent
- `‘` (U+2018) Left Single Quotation Mark
- `’` (U+2019) Right Single Quotation Mark
and probably some more.

I am not sure about the last 2 ones i specifically mentioned since those should be included by the regex
but that seems to be the character that AniList uses and it does not get removed or replaced...

This isn't good because if we get it announced with one of those it will not match...
I think because of that we should replace/remove Unicode categories Pf Pi and Sk - thoughts?

martylukyy:
It should be covered yeah... but i just checked again and it's not which is very weird to me too...
Check the synonyms of: https://anilist.co/anime/182158/Senpai-wa-Otokonoko-Ame-Nochi-Hare
and copy the P class char of the last synonym.

Compart says it's a https://www.compart.com/en/unicode/U+2019
but it doesn't get removed by the regex so something's up here and i don't know what...

What's even funnier to me:
If you copy the P class char of the synonym into the search field of your browser for this PR you get a couple of hits.
But those hits are for 3 (!!!) different unicode characters!
Can someone enlighten me what the fudge is going on here because i don't what is wrong or right anymore 😂

zze0s:
It's reasonable to split out the anime specific handling into a separate title processor, since there is a lot of special handling, if it makes sense.

nuxencs:
> Compart says it's a https://www.compart.com/en/unicode/U+2019
but it doesn't get removed by the regex so something's up here and i don't know what...

this test case shows that it should match 🤔 
https://regex101.com/r/uph6ZZ/2

martylukyy:
i think i found the culprit and we and discussing the fix

martylukyy:
So, i added
- `´` (U+00B4) Acute Accent 
- ``` ` ``` (U+0060) Grave Accent
- `‘` (U+2018) Left Single Quotation Mark
- `’` (U+2019) Right Single Quotation Mark
to the apostrophe replacers which fixes our problem.

Furthermore i made some changes to the `regionCodeRegexp` since it was too lax in some cases.
It matched everything `The Matrix*` while we were looking for `The Matrix Reloaded`
and it would have matched `this is him` while we want `this is him (can’t be anyone else)`

nuxencs:
good job finding that! 😄
that whole logic looks like it could need a refactor 😅

martylukyy:
All power to you, but you want to open the box of pandora even further than i just did? 😂

zze0s:
Lets keep the scope of this pr limited, there's enough changes as is ;)

nuxencs:
> Lets keep the scope of this pr limited, there's enough changes as is ;)

I was more so talking about an additional PR to tackle this 😄

nuxencs:
> Furthermore i made some changes to the regionCodeRegexp since it was too lax in some cases.

could maybe even restrict it to only 2 chars, as region codes should only be 2 chars long? 🤔

martylukyy:
hmm.. yeah need to run the tests with [a-z]{2]

nuxencs:
> hmm.. yeah need to run the tests with [a-z]{2] but i think it also get's rid of (GER) 🤔

according to [this](https://www.ibm.com/support/pages/country-or-region-codes-0) `[a-Z]{2,3}` should be enough to cover everything, but it also doesn't have `GER` 🤔

martylukyy:
> > hmm.. yeah need to run the tests with [a-z]{2] but i think it also get's rid of (GER) 🤔
> 
> according to [this](https://www.ibm.com/support/pages/country-or-region-codes-0) `[a-zA-Z]{2,3}` should be enough to cover everything, but it also doesn't have `GER` 🤔

I just realized the way it was before and is now it would also get rid of (2025) for example...
So we need to account for that too so we don't break matching...

martylukyy:
I noticed that we only need to exclude Unicode Block “Latin-1 Supplement” at this point of this PR
and we don't need to make changes to the API anymore.
I went ahead and closed the respective API PR. Feel free to reopen in case you disagree.

All the additional processing is now done in either title.go or process_list_anilist.go.
Unicode Block “Latin-1 Supplement” chars that previously caused problems 
will now be replaced by line 74 in process_list_anilist.go.

nuxencs:
I'd argue we should still expand the Latin-1 Supplement a bit further with the ones I suggested [here](https://github.com/autobrr/autobrr/pull/1949#issuecomment-2619431664), otherwise this looks good 👍🏼

web/src/components/inputs/select_wide.tsx
```diff
@@ -78,12 +78,8 @@ export function SelectFieldCreatable<T>({ name, label, help, placeholder, toolti
...
               value={field?.value ? { value: field.value, label: field.value  } : field.value}
               onChange={(newValue: unknown) => {
-                if (newValue) {
-                  setFieldValue(field.name, (newValue as { value: string }).value);
-                }
-                else {
-                  setFieldValue(field.name, "")
-                }
+                const option = newValue as { value: string };
+                setFieldValue(field.name, option?.value ?? "");
```
zze0s, line 82: Won't all of these error if the newValue is undefined for example?

martylukyy: I asked myself the same question but i couldn't find a select that triggered that error...

PR #1961, opened 2025-02-03
fix(web): border opacities

martylukyy:
This PR fixes the border opacities for 2 HeadlessUI listboxes
that got messed up during the tailwind v4 migration.

Left is before the fix with tailwind v4.
Right is after the fix with tailwind v4 and how i was with tailwind v3.

borderold borderfixed

darkborderold darkborderfixed


fabricionaweb:
LGTM

fabricionaweb (approved):

PR #1993, opened 2025-03-11
feat(filters): add DSD formats to music

fabricionaweb:
RED already supported and announces it since OCT 2024. Im not on OPUS (or other music tracker) so I dont know about them


zze0s (approved):
LGTM!

PR #2006, opened 2025-03-18
feat(oidc): show profile pic if present

s0up4200:
Use the user's profile pic from the OIDC provider. Fall back to the FontAwesome icon if it is empty. Link to the issuer in Account Settings.

CleanShot 2025-03-18 at 15 49 33@2x

CleanShot 2025-03-18 at 15 49 51@2x

No icon:
CleanShot 2025-03-18 at 22 16 35@2x


internal/auth/oidc.go
```diff
@@ -180,25 +180,25 @@ func (h *OIDCHandler) HandleLogin(w http.ResponseWriter, r *http.Request) {
 	http.Redirect(w, r, authURL, http.StatusFound)
 }
 
-func (h *OIDCHandler) HandleCallback(w http.ResponseWriter, r *http.Request) (string, error) {
+func (h *OIDCHandler) HandleCallback(w http.ResponseWriter, r *http.Request) (string, string, error) {
```
zze0s, line 183: Returning the whole claims struct as a pointer might be better here because then the caller can decide to grab username and extra variables.

internal/http/auth.go
```diff
@@ -213,8 +213,9 @@ func (h authHandler) validate(w http.ResponseWriter, r *http.Request) {
 		h.log.Debug().Msgf("found user session: %+v", session)
 		// Return username if available in session
 		response := map[string]interface{}{
-			"username":    session.Values["username"],
-			"auth_method": session.Values["auth_method"],
+			"username":        session.Values["username"],
+			"auth_method":     session.Values["auth_method"],
+			"profile_picture": session.Values["profile_picture"],
```
zze0s, line 218: `profile_picture` probably needs to be checked if it exists first here since it's might or might not exist on the session.

web/src/components/header/RightNav.tsx
```diff
@@ -19,7 +19,7 @@ import { AuthContext, SettingsContext } from "@utils/Context";
 
 export const RightNav = (props: RightNavProps) => {
   const [settings, setSettings] = SettingsContext.use();
-
+  const [imageError, setImageError] = useState(false);
```
zze0s, line 22: Is this state needed? Either we have a `profile_picture` string or we don't?

web/src/routes.tsx
```diff
@@ -304,10 +304,19 @@ export const AuthRoute = createRoute({
     if (!AuthContext.get().isLoggedIn) {
       try {
         const response = await APIClient.auth.validate();
+        // Also get OIDC config if needed
+        let issuerUrl;
+        if (response.auth_method === 'oidc') {
+          const oidcConfig = await APIClient.auth.getOIDCConfig();
+          issuerUrl = oidcConfig.issuerUrl;
+        }
```
zze0s, line 312: If we have this, then we could grab profile_picture here instead of storing it in the session token/response, which might be better.

web/src/screens/settings/Account.tsx
```diff
@@ -160,6 +161,22 @@ function Credentials() {
 }
 
 function OIDCAccount() {
+  const auth = AuthContext.get();
+  const [imageError, setImageError] = useState(false);
```
zze0s, line 165: Same state comment here, not sure what this adds versus checking if the variable exists and is not empty.

zze0s (approved):

PR #2021, opened 2025-04-03
chore(indexers): remove ESLint workflow

martylukyy:
Since this GH workflow now errors due to our unmaintained ESLint setup,
we should remove this workflow since we don't currently use ESLint anyway.

If we can decide on a ruleset for a ESLint setup i can put together a new working setup.
Be aware - since we did not have a working ESLint setup for a while now there will be a lot of stuff to lint.


martylukyy:
I mean we are still running a workflow for something we don't use currently.
If we still wanna keep it feel free to close 😄

PR #2034, opened 2025-04-14
feat(filters): show disabled indexer warning

s0up4200:
Shows a red warning on `/filters` if a filter is using an indexer in disabled state.
On hover it will list all indexers in disabled state.

CleanShot 2025-04-15 at 00 15 50@2x

### DIFF
--- a/web/src/screens/filters/List.tsx
+++ b/web/src/screens/filters/List.tsx
@@ -38,7 +38,7 @@ import { FilterAddForm } from "@forms";
 import { useToggle } from "@hooks/hooks";
 import { APIClient } from "@api/APIClient";
 import { FilterKeys } from "@api/query_keys";
-import { FiltersQueryOptions, IndexersOptionsQueryOptions, IndexersQueryOptions } from "@api/queries";
+import { FiltersQueryOptions, IndexersOptionsQueryOptions } from "@api/queries";
 import { toast } from "@components/hot-toast";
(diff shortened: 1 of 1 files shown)


web/src/screens/filters/List.tsx
```diff
@@ -195,17 +196,25 @@ function FilterList({ toggleCreateFilter }: any) {
...
 
-  const { isLoading, data, error } = useQuery(FiltersQueryOptions(indexerFilter, sortOrder));
+  const { isLoading: isLoadingFilters, data: filtersData, error: filtersError } = useQuery(FiltersQueryOptions(indexerFilter, sortOrder));
+  const { data: indexersData, isLoading: isLoadingIndexers } = useQuery(IndexersQueryOptions());
+
+  // create a set of disabled indexer IDs for quick lookup
+  const disabledIndexerIds = useMemo(() => {
+    if (!indexersData) return new Set<number>();
+    return new Set(indexersData.filter(indexer => !indexer.enabled).map(indexer => indexer.id));
+  }, [indexersData]);
```
zze0s, line 206: Great addition!

There is however no need to fetch the indexer list again because the filters call include the indexers per filter, with the enabled property.

This can however keep the disabledIndexerIds for convenience, or just keep the current logic with `disabledIndexersInfo` in the component.

zze0s (approved):
LGTM! 🥳

PR #2036, opened 2025-04-17
feat(lists): add option to skip cleaning of Plaintext data

luckylittle:
At the moment, `internal/list/title.go` has `processTitle`, which correctly cleans up the unwanted characters, such as:
`What If…?` ➡️  `What?If*`
`*****… (los asteriscos…)` ➡️  `*los?asteriscos*`

This is great and expected for regular titles, typically coming from *arrs. However, this is causing undesirable behavior, when you are trying to populate **Match Release**, that contains wildcards - for example:
The string `daredevil?born?again?s01*720p` in the plaintext is changed to `*daredevil?born?again?s01?720p*` and therefore not working - the asterisk (`*`) is changed to question mark (`?`).

I am proposing a new list type called **Plaintext Untouched**, that bypasses the `processTitle` and simply concatenates (using the `,`) the titles in whatever format they are.


luckylittle:
Can someone please comment/review?

luckylittle:
> If we're being given garbage (questionable?) titles from sonarr, we should handle those cases. The example provided doesn't make sense, because `dare*` and `*dare*` should result in the later matching the former.
> 
> (so I think at face value, the description is absolutely not what autobrr should inflict on users - let's fix this in general)

@KyleSanderson 

If there is garbage coming from *arrs, it is already being handled quite well. Maybe I was not clear what the intention is - I believe there is need for both types of plaintext lists. The main purpose of this "RAW" list is to synchronize lists "as-is" and not filter or escape any characters. I need the ability to push filters to multiple Autobrrs instances ideally via Git - and also I believe it would be a good practice to have filters-as-code.

zze0s:
I replied on discord when you posted the pr initially, but maybe you missed it. 

I wrote

> Maybe it can be simplified to just check if the line contains any wildcard instead of a separate type and implementation?
> so if match releases is toggled and it contains any wildcard it just joins them?
> and check if it's surrounded by stars, if not add them

And about autobrr-filters-as-code, sure, new acronym invented AFAC. The lists can read from file so you have half of it. Everything else have an api endpoint, and I think someone already built a terraform provider for autobrr to do something like it.

But do you really have to use multiple instances, when one of the core features is to support local and remote clients?

Could probably be easier to use one of the many SQLite tools to sync databases?

luckylittle:
I think we are on the same page. You're absolutely right that the existing behavior helps reduce support issues by ensuring more forgiving input handling, and I completely agree that’s something worth preserving. My intention with introducing the `Plaintext Untouched` or `Plaintext (Raw)` - whatever you want to call it - is to offer a path forward that maintains those benefits while also giving power users an explicit way to work with exact formats like `daredevil?born?again?s01*720p`, which currently get unintentionally altered due to the `processTitle` logic.

To be clear, this change is fully backward-compatible - the existing list type and its cleanup behavior remain untouched, so existing setups will continue to work exactly as before. `Plaintext Untouched` is opt-in and designed for those edge cases where precision matters, like advanced wildcard-based matching. We'll also make sure to clearly document both list types and their behaviors so users understand which to use and when.

I’m fully on board with finding that middle ground you mentioned - one that supports more advanced use cases without increasing the support burden. Happy to iterate further if you think there’s a better way to position or name this. Also feel free to add commits to this branch.

zze0s:
I like that this is backwards compatible and all, but what about adding a new toggle like `Skip Clean/Sanitize/other` which is opt-in, on the existing Plaintext list instead?

Could make it a json column in the db that takes a Settings object so we can easily add more options in the future without further migrations needed. We do that already in a couple of places and it works pretty well.

Thoughts? I can help out with the data layer if you don't want to touch that.

luckylittle:
I've reverted old & pushed a new commit that should implement the changes you proposed, including the toggle on the Plaintext list and the settings stored as JSON. When you get a chance, could you take a look and let me know if it's in line with what you had in mind? And if I missed anything, I’d really appreciate your help on the data layer side. Thanks again!

zze0s (commented):
Looks good! Will test it out 👍

zze0s (approved):

PR #2039, opened 2025-04-19
fix(metrics): irc channel enabled gauge

Intuinewin:
Plz see https://github.com/autobrr/autobrr/discussions/2035

Currently a channel is only considered enabled if it's enabled in the db.

However when you disable an IRC server, its channel remains enabled in the db but should be considered disabled in the metrics.


zze0s (approved):
Thanks for the fix!

onedr0p:
Yes, thanks @Intuinewin ❤️

PR #2042, opened 2025-04-23
feat(web): persist releases incognito state

s0up4200:
Persists incognito state using localStorage via `IncognitoContext`

Updates the list of Linux ISOs used in `RandomLinuxIsos`. (Taken from https://distrowatch.com/dwres.php?resource=bittorrent&sortorder=date)

Added a new `RandomIsoTracker` utility to randomize tracker names. (Added fosstorrents to the list)

### DIFF
--- a/web/src/utils/Context.ts
+++ b/web/src/utils/Context.ts
@@ -12,6 +12,7 @@ interface SettingsType {
   scrollOnNewLog: boolean;
   indentLogLines: boolean;
   hideWrappedText: boolean;
+  incognitoMode: boolean;
 }
 
 export type FilterListState = {
@@ -42,7 +43,8 @@ const SettingsContextDefaults: SettingsType = {
   darkTheme: window.matchMedia('(prefers-color-scheme: dark)').matches,
   scrollOnNewLog: false,
   indentLogLines: false,
-  hideWrappedText: false
+  hideWrappedText: false,
+  incognitoMode: false
 };
 
 const FilterListContextDefaults: FilterListState = {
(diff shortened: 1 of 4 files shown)


web/src/utils/Context.ts
```diff
@@ -79,6 +79,17 @@ function ContextMerger<T extends {}>(
...
+
+const getInitialIncognitoState = (): boolean => {
+  try {
+    const stored = localStorage.getItem(IncognitoKey);
+    return stored ? JSON.parse(stored) : false;
+  } catch (e) {
+    console.error(`Failed to read ${IncognitoKey} from localStorage`, e);
+    return false;
+  }
+};
```
zze0s, line 92: Could we merge this into the `SettingsContext` to have a single place for such settings/states?

s0up4200: Should be good now I think

web/src/utils/Context.ts
```diff
@@ -55,7 +57,8 @@ const FilterListContextDefaults: FilterListState = {
 function ContextMerger<T extends {}>(
   key: string,
   defaults: T,
-  ctxState: StateWithValue<T>
+  ctxState: StateWithValue<T>,
+  incognitoKey?: string
```
zze0s, line 61: This extra thing shouldn't be needed I think? We just keep the single SettingsKey so there's no separate handling, all done via a single context for settings.

s0up4200: now its good I think, no separate key for local storage now

zze0s (approved):
LGTM!

PR #2044, opened 2025-04-24
feat(indexers): add T66y

nitrix:
## What

This adds supports for t66y.com (Caoliu Community) which is a forum aggregating mostly Chinese/Japanese adult content.

## Changes

* Adds a new `T66y` indexer using IRC (Rizon.net on #t66y).
* Adds the ability for IRC indexer definitions to templatize `torrenthash` and `magneturi`.
* Adds tests.

## Note

The categories on the forum don't map to anything common/generic on the releases so they are kept to their original numeric form.

### DIFF
--- a/internal/domain/indexer.go
+++ b/internal/domain/indexer.go
@@ -433,11 +433,6 @@ func (p *IndexerIRCParse) Parse(def *IndexerDefinition, vars map[string]string, 
 		return errors.Wrap(err, "could not parse release name")
 	}
 
-	// parse torrent hash
-	if err := def.IRC.Parse.Match.ParseTorrentHash(mergedVars, rls); err != nil {
-		return errors.Wrap(err, "could not parse release hash")
-	}
-
 	var parser IRCParser
 
 	switch def.Identifier {
--- a/internal/domain/indexer.go
+++ b/internal/domain/indexer.go
@@ -250,6 +250,8 @@ type IndexerIRCParseLine struct {
 type IndexerIRCParseMatch struct {
 	TorrentURL  string   `json:"torrenturl"`
 	TorrentName string   `json:"torrentname"`
+	TorrentHash string   `json:"torrenthash"`
+	MagnetURI   string   `json:"magneturi"`
 	InfoURL     string   `json:"infourl"`
 	Encode      []string `json:"encode"`
 }
@@ -331,6 +333,15 @@ func (p *IndexerIRCParseMatch) ParseURLs(baseURL string, vars map[string]string,
 		rls.DownloadURL = downloadURL.String()
 	}
 
+	if p.MagnetURI != "" {
+		magnetURI, err := parseTemplateURL("magnet:", p.MagnetURI, vars, "magneturi")
+		if err != nil {
+			return err
+		}
+
+		rls.MagnetURI = magnetURI.String()
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


internal/domain/indexer.go
```diff
@@ -353,6 +364,24 @@ func (p *IndexerIRCParseMatch) ParseTorrentName(vars map[string]string, rls *Rel
...
+		var torrentHashBytes bytes.Buffer
+		if err := tmplTorrentHash.Execute(&torrentHashBytes, &vars); err != nil {
+			return errors.New("could not write torrent hash template output")
+		}
+
+		rls.TorrentHash = torrentHashBytes.String()
+	}
+
+	return nil
+}
```
zze0s, line 383: Isn't the hash in this case just a string? It doesn't look like it really needs to run with templates to execute

internal/indexer/definitions/t66y.yaml
```diff
@@ -0,0 +1,76 @@
...
+
+        pattern: '(\d+)\ (\d+)\ ([a-z0-9]{40})\ (.*)'
+        vars:
+          - postId
+          - category
+          - torrentHash
+          - torrentName
+
+    match:
+      magneturi: 'magnet:?xt=urn:btih:{{ .torrentHash }}&dn={{ urlquery .torrentName }}&tr=http://sukebei.tracker.wf:8888/announce&tr=udp://tracker.opentrackr.org:1337/announce&tr=http://tracker.bt4g.com:2095/announce&tr=http://1337.abcvg.info/announce&tr=http://tracker.opentrackr.org:1337/announce&tr=http://93.158.213.92:1337/announce&tr=udp://open.stealth.si:80/announce&tr=udp://exodus.desync.com:6969/announce&tr=udp://tracker.torrent.eu.org:451/announce'
```
zze0s, line 76: Do we have to have all of these trackers hardcoded?

zze0s (commented):
Hi! Thanks for the PR! Left some comments.

internal/indexer/definitions/t66y.yaml
nitrix: It mimicks what their rmdown site was adding, though I agree with you that DHT alone should be sufficient to find peers. Most people running autobrr probably have their bittorrent client configured on a server with a reasonably not too restricted internet.

I'm just going to remove them then so it doesn't smell as much.

internal/domain/indexer.go
nitrix: Agreed. Saw too big there, updating.

nitrix:
Updated the tests too, PR should be in a better shape now.

internal/domain/indexer.go
zze0s: Because it's just a string and we don't have to do anything special with it, and it's part of the `vars`:

```yaml
vars:
  - postId
  - category
  - torrentHash
  - torrentName
```

it doesn't need any special handling as a parse function because these vars are available when we build the variables under `match`, like your new `magneturi` and `torrenturl` etc.

We do however have to explicitly set it like this [here in MapVars](https://github.com/autobrr/autobrr/blob/039162986261b64dac94fc372152432217509457/internal/domain/release.go#L967-L969) if we want to be able to use it later for anything.

```go
if torrentHash, err := getStringMapValue(varMap, "torrentHash"); err == nil {
	r.TorrentHash = torrentHash
}
```

nitrix: I see. Updated, I moved the torrentHash to the MapVars.

internal/domain/indexer.go
```diff
@@ -250,6 +250,8 @@ type IndexerIRCParseLine struct {
 type IndexerIRCParseMatch struct {
 	TorrentURL  string   `json:"torrenturl"`
 	TorrentName string   `json:"torrentname"`
+	TorrentHash string   `json:"torrenthash"`
```
zze0s, line 253: This can be removed since it's no longer used

internal/domain/irc.go
```diff
@@ -161,7 +161,7 @@ type IRCParser interface {
 
 type IRCParserDefault struct{}
 
-func (p IRCParserDefault) Parse(rls *Release, _ map[string]string) error {
+func (p IRCParserDefault) Parse(rls *Release, vars map[string]string) error {
```
zze0s, line 164: This can be reverted to `_` since we do not use the vars here anymore.

zze0s (changes requested):
Just minor changes! Then it should be ready!

Are you able to test it before we merge so we see if the magnet stuff works as expected etc?

internal/domain/irc.go
nitrix: Done in `5e95dc6`.

internal/domain/indexer.go
nitrix: Done in `5e95dc6`.

zze0s:
@nitrix Looks good now! I added a test for the magneturi parsing as well to make sure it works as intended.

Were you able to test it out to make sure releases are able to be grabbed without the extra trackers in the magneturi?

nitrix:
It sets up an IRC network correctly:
image

My filter picks up new releases:
image

Sent to the action (qBittorrent in my case) correctly:
image

zze0s (approved):
Perfect! Thanks for your contribution!

PR #2048, opened 2025-04-25
fix(http): change onboarding unavailable log level

martylukyy:
This PR changes the log level for the `onboarding unavailable` log message,
since we experienced an upsurge in support workload due to people asking about this warning.

### DIFF
--- a/internal/http/auth.go
+++ b/internal/http/auth.go
@@ -199,8 +199,8 @@ func (h authHandler) onboardEligible(ctx context.Context) (int, error) {
 	}
 
 	if userCount > 0 {
-		h.log.Trace().Msg("onboarding unavailable")
-		return http.StatusServiceUnavailable, errors.New("onboarding unavailable")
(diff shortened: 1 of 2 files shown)


s0up4200 (dismissed):

zze0s (commented):
Instead of checking the message `StatusWarning` encoder maybe we can just change the used function in `internal/http/auth.go` on line 182 to StatusResponse and still put in the same `status, err.Error())

Keep the log on 202, that's still good.

Thoughts?

martylukyy:
> Instead of checking the message `StatusWarning` encoder maybe we can just change the used function in `internal/http/auth.go` on line 182 to StatusResponse and still put in the same `status, err.Error())
> 
> Keep the log on 202, that's still good.
> 
> Thoughts?

Good point!
Way cleaner and we can always change warnings to responses instread of fiddling with a string slice.

zze0s (approved):

PR #2054, opened 2025-04-29
fix(releases): handle multiple episodes in same release

### LINKED ISSUE #972: Multi-episode releases are not detected correctly
**Version**
v1.26.2

**Describe the bug**
Multi-episode releases (e.g. S01E23E24) are detected as season packs (season 0 episode 0).  
This is not correct.

Log:
```
DEBUG
filter.Service.CheckFilter: (Filter) for release: Ghost Force S01E23E24 1080p HULU WEB-DL DDP 5.1 H.264-LAZY rejections: (season not matching. got: 0 want: 1-99, episodes not matching. got: 0 want: 1-99)
```

**To Reproduce**
Wait for a multi-episode release to be announced.

**Expected behavior**
Season should at least be detected as Season 1 in the example.
Regarding episodes it should be supported but I guess it'll require more development as it only support a single episode value.

komuw:
- Updates: https://github.com/autobrr/autobrr/issues/972
- The actual issue is that the release parsing library, https://github.com/moistari/rls did not have support for multiple episodes. I have opened a PR upstream to fix that; https://github.com/moistari/rls/pull/12 
- This PR would have to await the resolution of the [upstream one](https://github.com/moistari/rls/pull/12) because that will determine what sort of fix we can implement here.
  But I'm opening this PR for awareness and early feedback. 
- In this PR, if a release has multiple episodes we pick one of them when creating a release. This is better than the current behavior on develop branch where it fails with an error; https://github.com/autobrr/autobrr/issues/972      
  It is however not a complete fix since we ignore the second episode. There are a couple of ways we could handle this;    
    (a) We could change `Release.ParseString()` to be a stand-alone function `func ParseString(title string) []Release` that returns a list of releases. In the case of multiple episodes it would return two releases(one for each episode)
    (b) We could also change `Release.Episode` to be a `[]int` instead of `int`.  That's what I've done upstream; https://github.com/moistari/rls/pull/12 
    (c) Something else.    
  I'm looking for feedback on which option to go with. Obviously this is highly dependent on how the upstream `rls` library decides to handle it. So we should probably wait the resolution of https://github.com/moistari/rls/pull/12


zze0s:
Hi! Thanks for starting to look into this.

I'll try and poke the dev. There is allegedly a full rewrite of the library that is yet to be released, and maybe it fixes this. But it might never see the light of day and the dev might have gone dark so we might have to fork the lib, but we'll see.

moistari:
@zze0s i will merge the commit here, or i'll do it myself. the rewrite is progressing, i just don't want to release anything that's not working.

zze0s:
This should already be working since https://github.com/autobrr/rls/releases/tag/v0.6.0

PR #2067, opened 2025-05-09
Add music channel to SeedPool's autobrr definition.

frank2:
Credit goes to user An7hrax for the actual regex work, the only thing I did was give it a multi parser home in the original file. Let me know if this is acceptable for the autobrr project!


frank2:
Closing PR since this suffers from a fundamental misunderstanding of tracker definition files: I am attempting to map multiple channels to multiple patterns, when that doesn't have an example currently. Will hang around the Discord for further dev assistance.

PR #2068, opened 2025-05-09
fix(web): ensure release table auto-update

s0up4200:
Fixes an issue where the Activity and Release tables would not auto-update with new data unless the page was manually refreshed. The regression was introduced in commit 24648e45. The root cause is related to how the table detects data changes and how `useEffect` dependencies were structured after the incognito mode feature was added.


s0up4200:
Confirmed to auto-update on my instance. Both on the front page and the Releases tab.

PR #2077, opened 2025-05-22
feat(indexers): CapybaraBR parse freeleech and internal

a1Thiago:
This PR updates the CapybaraBR tracker parsing, adding FreeleechPercent and internal fields.

### DIFF
--- a/internal/indexer/definitions/capybarabr.yaml
+++ b/internal/indexer/definitions/capybarabr.yaml
@@ -50,7 +50,7 @@ irc:
     type: single
     lines:
       - tests:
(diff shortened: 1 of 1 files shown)


internal/indexer/definitions/capybarabr.yaml
```diff
@@ -59,8 +59,10 @@ irc:
               baseUrl: https://capybarabr.com/
               torrentId: "15123"
               torrentSize: 25.36 GiB
+              FreeleechPercent: "100%"
```
zze0s, line 62: Change all the occurrences to camelCase `freeleechPercent`.

zze0s (changes requested):
Hey! Thanks for the PR! Looks good, but the variable for `FreeleechPercent` needs to become `freeleechPercent` (camelCase).

internal/indexer/definitions/capybarabr.yaml
a1Thiago: Hey! Thanks for the feedback. I've made the changes you requested. Let me know if there's anything else

zze0s (approved):

PR #2085, opened 2025-06-05
feat(indexers): new IRC auth mechanism for RocketHD

martylukyy:
RocketHD will make a change to their IRC server and its announce channel this weekend. 
After this change, the IRC server requires authentication with a network/server password. 
Users will be able to obtain an IRC key on site which needs to be used as their IRC server password. 
A channel password or NickServ auth will no longer be required.


martylukyy:
Their IRC server will no longer be running NickServ so i removed the fields from the form.
Additionally i added database migrations for a smoother transition so that users just have to provider their
IRC server password instead of changing removing all the stuff too.

martylukyy:
Tested SQLite and PostgreSQL migrations successfully.

PR #2086, opened 2025-06-06
feat(ui): filters grouping

luckylittle:
#### What is the relevant ticket?
* Discussed in Discord #dev-general

##### Add
* option `groupFiltersInListPage`
* function `getCommonPrefix`
* function `createGroupedItems`
* const `toggleGroupingSetting`
##### Change
* None
##### Remove
* None

* `web/src/utils/Context.ts`
* `web/src/screens/filters/List.tsx`

* `make build && go run cmd/autobrr/main.go`

#### Risk involved?
* Unsure, needs comprehensive test?

#### Screenshots (if appropriate):
* New grouping option in the console:
Screenshot from 2025-06-06 10-54-30

* Expanded groups - example:
Screenshot from 2025-06-06 10-54-48

* Collapsed groups - example:
Screenshot from 2025-06-06 10-55-09

* No

### DIFF
--- a/web/src/screens/filters/List.tsx
+++ b/web/src/screens/filters/List.tsx
@@ -191,6 +191,8 @@ function filteredData(data: Filter[], status: string) {
 }
 
 // Helper function to find the common prefix of two strings
+const MIN_PREFIX_LENGTH_FOR_GROUPING = 3;
+const MIN_ITEMS_FOR_GROUP = 1;
 function getCommonPrefix(s1: string, s2: string): string {
   let i = 0;
   while (i < s1.length && i < s2.length && s1[i] === s2[i]) {
@@ -206,9 +208,6 @@ interface GroupedItem {
   filters?: Filter[]; // Filters in this group
   filter?: Filter;   // Single filter if not in a group
 }
-
-const MIN_PREFIX_LENGTH_FOR_GROUPING = 3; // Minimum length of a common prefix to be considered for grouping
-const MIN_ITEMS_FOR_GROUP = 2;          // Minimum number of items to form a group
 
 function createGroupedItems(allFilters: Filter[]): GroupedItem[] {
   if (!allFilters || allFilters.length === 0) return [];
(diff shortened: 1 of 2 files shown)


martylukyy (commented):
Okay, so there a few things i want to point out...

---

First, if you have a filter with `ABC` and another with `ABC and D` these filters will not be grouped
even though they start with the same 3 letters.
While i understand that `ABC` is not really a prefix in this case,
there are people who use just the abbreviation for their indexers as a base filter
and then expand on that name based on certain criteria.

---

Second, the prefixes should be sanitized of whitespaces.
Currently if you have `TL 1` and `TL 2` you will get a `TL ` group with a trailing whitespace.
Not ideal since some users will be wondering why certain filters are being grouped even though 
`MIN_PREFIX_LENGTH_FOR_GROUPING` is 3.

---

Third, this is a thing:
image
The Actions counter should not have a higher z index than the heading.
 
---

Fourth, people with **a lot** of filters will get a **significantly** longer list since single filters will still get a group heading,
even if you set `MIN_ITEMS_FOR_GROUP` to 2 single filters do still get a group with a heading.

I understand that you can collapse them but they start expanded each time you disable and enable grouping.
Collapsing (or expanding) all those groups feel like a chore to me.

I don't know what the best approach here would be since not grouping them at all feels suboptimal aswell..
In my head i would do it like a file explorer where you have folders with grouped filter at the top
and ungrouped filters at the bottom but i understand that this is way more work than the current solution.

---

Last but not least:

The import for `ChevronRightIcon` does not have a conflict and does not need to be renamed.
Adding another flavour of icons to the imports is also not neccessary.
Adding the `ChevronRightIcon` was all that was needed! 😄 

---


We thought about a grouping solution that accommodates everyone's needs but this solution would include
adding a new field to the database where you could choose one or more groups for a filters within the filter itself.
The prefix stuff is a valid approach but will lack options for people that would like their filters grouped by more factors.
An example of this would be indexer and resolution or media type.

While we appreciate your continued efforts and contributions this is not something we can see us adding to autobrr.
Even with the things pointed out fixed.

PR #2090, opened 2025-06-09
fix(database): db convert SQLite to PG issues

### LINKED ISSUE #1622: Error when converting from SQLite to PostgreSQL
**Version**
v1.44.0

**Describe the bug**
I tried to run 
`autobrrctl db:convert --sqlite-db /path/to/autobrr.db --postgres-url postgres://username:password@127.0.0.1:5432/autobrr`
to migrate an existing database, but it returns: 
`2024/08/20 14:22:37 Failed to query SQLite table 'action': SQL logic error: near "?": syntax error (1)`

**Additional context**

I suspect the problem might be here, but I don't know how to fix it.
https://github.com/autobrr/autobrr/blob/cb2c62b353fc29aa674f0acd134ae3a011653370/internal/database/tools/convert.go#L96

katiekloss:
- SQLite doesn't support parameterized table names, so now the tool does a `Sprintf` instead.

- Tables aren't migrated in the right order to maintain foreign keys, and some newer tables are missing from the list as well.

- FKs with `ON DELETE SET NULL` will contain old stale values if the rows they reference are deleted, so those values need to be nulled out in order to successfully migrate the rows (most commonly `release` and `release_action_status`).

- Postgres sequences need to be updated with the real most-recent IDs from their tables (instead of the default 1), so that new rows don't collide with the newly-migrated ones.

Mostly fixes #1622.

### DIFF
--- a/cmd/autobrrctl/main.go
+++ b/cmd/autobrrctl/main.go
@@ -25,6 +25,7 @@ import (
 	"github.com/autobrr/autobrr/internal/logger"
 	"github.com/autobrr/autobrr/internal/user"
 	"github.com/autobrr/autobrr/pkg/errors"
+
 	_ "github.com/lib/pq"
 	"golang.org/x/crypto/bcrypt"
 	"golang.org/x/term"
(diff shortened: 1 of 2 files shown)


katiekloss:
@zze0s Finally came back to this, it's good to review. I think I tested all of the possible migration failures.

PR #2100, opened 2025-06-16
feat(indexers): revert UHDBits regex pattern

### LINKED ISSUE #2098: UHDBits: line not matching expected regex
**Version**
1.63.1

**Describe the bug**
Fails to parse the release name from the announce channel

```
New Torrent: Final Destination: Bloodlines [2025] - ###  Type: Movie / 1080p / WEB-DL / Freeleech: 100 Size: 4.28GB - https://uhdbits.org/torrents.php?id=##### / https://uhdbits.org/torrents.php?action=download&id=#####
```

**To Reproduce**

1. Add UHDBits as an indexer.

**Expected behavior**
Release matches the regex pattern and gets parsed.

**Additional context**
It appears as if the announcements have changed back to their previous format, as it was prior to the change here https://github.com/autobrr/autobrr/pull/561

I've done some testing in Regex101 and it seems to match as expected.

Image

martylukyy:
We have received a report of UHDBits reverting their announces to their old format.
This PR reverts our regex pattern to our previous one, with some minor adaptations.
It would be nice if someone could organize some more test examples,
since the one we received was edited and i am unsure if we covered all the cases.

resolves #2098

### DIFF
--- a/internal/indexer/definitions/uhdbits.yaml
+++ b/internal/indexer/definitions/uhdbits.yaml
@@ -66,21 +66,21 @@ irc:
     type: single
     lines:
       - tests:
-        - line: Category [Movie] A movie [1988] A.Movie.1988.Extended.720p.BluRay.x264-GROUP.mkv [720p, Encode, Freeleech, Scene, mkv] Size [4.99 GB] Url [https://uhdbits.org/torrents.php?action=download&id=00000]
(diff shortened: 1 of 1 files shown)


zze0s (approved):
Looks like their uhdAnnounceBot with the old format was z-lined on the 9th of June. Prior to that they had both old and "new/current" format from two different bots.

Wonder if this is intentional. If they want this old inferior format then sure :+1:

PR #2104, opened 2025-06-17
feat(releasedownload): clean up old temp files

### LINKED ISSUE #2103: Garbage collection in /tmp
**Version**
1.63.1

**Describe the bug**
I've noticed that there are thousands of temporary autobrr files in `/tmp`:

```
-rw-------.  1 user user    7479 Jun 16 17:22 autobrr-941106339
-rw-------.  1 user user   16128 Jun 15 03:34 autobrr-944336511
-rw-------.  1 user user   26416 Jun 11 18:57 autobrr-947476959
-rw-------.  1 user user   13277 Jun 13 11:05 autobrr-94814340
-rw-------.  1 user user   11044 Jun 12 00:46 autobrr-950669956
-rw-------.  1 user user   41579 Jun  8 12:48 autobrr-951126355
-rw-------.  1 user user   18143 Jun 13 13:59 autobrr-953622486
-rw-------.  1 user user  190643 Jun  8 05:49 autobrr-955763458
-rw-------.  1 user user   41603 Jun  7 12:20 autobrr-95634370
-rw-------.  1 user user   46391 Jun  8 12:47 autobrr-958252384
-rw-------.  1 user user   18139 Jun 15 05:22 autobrr-958901074
-rw-------.  1 user user   37607 Jun 16 00:49 autobrr-95971992
-rw-------.  1 user user  337725 Jun 11 06:33 autobrr-96035351
-rw-------.  1 user user   17238 Jun 15 19:11 autobrr-96242459
-rw-------.  1 user user  119942 Jun  9 19:08 autobrr-9657066
-rw-------.  1 user user   23937 Jun 13 15:13 autobrr-972211505
-rw-------.  1 user user  111799 Jun 11 07:11 autobrr-97293874
-rw-------.  1 user user  122961 Jun 10 12:23 autobrr-976339858
-rw-------.  1 user user   19783 Jun 16 05:46 autobrr-97838329
-rw-------.  1 user user   13054 Jun 11 14:24 autobrr-978833184
-rw-------.  1 user user   18759 Jun 14 20:03 autobrr-980769833
-rw-------.  1 user user   13795 Jun 13 10:01 autobrr-982724120
-rw-------.  1 user user    8075 Jun 17 00:46 autobrr-982769725
-rw-------.  1 user user   47399 Jun  7 12:18 autobrr-983341462
...
```

```
$ ls -la /tmp/autobrr* | wc -l
2097
```

**To Reproduce**
1. I have only discovered this now - unsure if it's normal? They aren't errored pushes as I don't have that many.

**Expected behavior**
1. I would expect the files to be gradually removed, when they are not used after some time? Or is it expected to be cleaned by `/usr/lib/systemd/system/systemd-tmpfiles-clean.timer`?

**Screenshots**
N/A

**Additional context**
N/A

martylukyy:
resolves #2103 

This PR adds code to remove old temporary files with the prefix `autobrr-`

There already is some code for this flying around in the codebase but 
i couldn't make sense on when it gets called and my debugging skills weren't enough to figure it out.

https://github.com/autobrr/autobrr/blob/6073480bc42bf77383638507c647cd66d9abc606/internal/domain/release.go#L940-L947
https://github.com/autobrr/autobrr/blob/46f6fbe5cc4d95b8d299eb4bbc222c3e12a2e566/internal/release/service.go#L206

This PR has likely a lot of room for improvement so feel free to suggest changes!

### DIFF
--- a/internal/releasedownload/download_service.go
+++ b/internal/releasedownload/download_service.go
@@ -14,6 +14,7 @@ import (
 	"net/http/cookiejar"
 	"net/url"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
@@ -146,6 +147,16 @@ func (s *DownloadService) downloadTorrentFile(ctx context.Context, indexer *doma
 
 	tmpFilePattern := "autobrr-"
 	tmpDir := os.TempDir()
+
+	// Clean up old tmp files
+	files, err := os.ReadDir(tmpDir)
+	if err == nil {
+		for _, file := range files {
+			if strings.HasPrefix(file.Name(), tmpFilePattern) {
+				os.Remove(filepath.Join(tmpDir, file.Name()))
+			}
+		}
+	}
 
 	// Create tmp file
 	// TODO check if tmp file is wanted


zze0s:
This is better suited to run as a daily cleanup job rather than trigger per each download.

Check the implementation for scheduler and `CheckUpdatesJob`. You won't need any special service dependencies for the job, just have it run the cleanup code and add the job with an interval/schedule once daily.

martylukyy:
Done.

Still thinking about yeeting all the code mentioned in https://github.com/autobrr/autobrr/pull/2104#issuecomment-2981158376 ...

internal/scheduler/jobs.go
```diff
@@ -48,3 +51,30 @@ func (j *CheckUpdatesJob) Run() {
...
+				tempFiles := filepath.Join(tmpDir, file.Name())
+				fileInfo, err := os.Stat(tempFiles)
+				if err == nil {
+					if fileInfo.ModTime().Before(time.Now().Add(-time.Hour)) {
+						os.Remove(tempFiles)
+					}
+				}
+			}
+		}
+	}
```
zze0s, line 79: I think it's wise to add error handling to both os.Readdir, os.Stat and os.Remove functions and log errors.

Do we possibly have to check for the owner of the files in case of shared systems?

And maybe have to age check only delete files older than 24h? In case someone is doing backup or something else with the files.

Could also add a counter and log the amount of deleted files at the end. If you want to make it even fancier sum the size of the files as well and log that on the same line.

zze0s (approved):

PR #2124, opened 2025-07-21
feat(filters): add per-filter notification support

s0up4200:
## Summary

This PR implements per-filter notification support, allowing users to configure specific notifications for individual filters that override the global notification settings. This enables fine-grained control over which notifications are sent for specific filters.

## How It Works

The implementation uses a two-tier notification system:

1. **Global Notifications** - The existing system, applies to all filters by default
2. **Filter Notifications** - New per-filter overrides

When a filter has notifications configured:
- Only the filter-specific notifications are used
- Global notifications are completely ignored for that filter
- This is an all-or-nothing override

### Muting Filters

Filters can be effectively muted by adding a notification with no events selected:
- The "Mute filter" button quickly adds a notification with no events
- Visual indicator shows when a filter is muted
- No notifications will be sent for muted filters (not even global ones)

### System Events

System-level events always use global notifications and are never affected by filter-specific settings:
- IRC Disconnected/Reconnected events
- App Update Available events
- These cannot be overridden or muted at the filter level

## Database Changes

```sql
CREATE TABLE filter_notification (
    filter_id        INTEGER NOT NULL,
    notification_id  INTEGER NOT NULL,
    events          TEXT[] NOT NULL DEFAULT '{}',
    created_at      TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
    PRIMARY KEY (filter_id, notification_id),
    FOREIGN KEY (filter_id) REFERENCES filter(id) ON DELETE CASCADE,
    FOREIGN KEY (notification_id) REFERENCES notification(id) ON DELETE CASCADE
);
```

## API Changes

New endpoints:
- `GET /api/filters/{id}/notifications` - Returns array of filter notifications
- `PUT /api/filters/{id}/notifications` - Updates filter notifications (replaces all)

## Migration

- Existing setups continue to work unchanged
- Filter notifications are opt-in per filter

## Performance Impact

- One additional database query when loading filters (to fetch notifications)

## Breaking Changes

None. This is a backwards-compatible addition.


luckylittle:
I have implemented a temporary workaround for this using the secondary "Exec" action:
Action: `Exec`
Name: `Pushover Notification`
Path to executable: `/usr/bin/curl`
Arguments: `-s --form-string "token=<API_TOKEN>" --form-string "user=<API_USER>" --form-string "message=TRACKER: {{.IndexerName}} NEW TORRENT: {{.TorrentName}}" https://api.pushover.net/1/messages.json`

PR #2146, opened 2025-08-13
build(make): add build/dockerx for cross platform docker builds

kinghrothgar:
This is needed to build docker images for other platforms (I'm on Apple ARM for instance). This will build the image for both arm64 and amd64. Docker and container registries seamlessly support multiple archs for the same image tag, so if you build both, it will automatically select the correct one to run. Similarly, if you push the image tag, it will push up both and when you pull, it will select the correct arch for the target system.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -172,6 +172,14 @@ make build/docker
 
 The image will be tagged as `autobrr:dev`
 
+To build a cross platform Docker image (for instance if you're running on arm64 or Apple ARM):
+
+```shell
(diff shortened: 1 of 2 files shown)


kinghrothgar:
@KyleSanderson I thought the purpose of the build command in the Makefile was for local dev. I am aware images are published for many archs, just thought it would be nice to make cross arch building easier locally for what I believe are the two most common arches people dev on. I build on ARM but need to be able to test it also in my amd64 Kube cluster.

zze0s (approved):

PR #2158, opened 2025-08-24
feat(http): implement session storage in database

zze0s:
Implement session storage in the database.


ppfeister (approved):
LGTM
Didn't do a full manual CR, but it was a successful upgrade and it passes validation.

_(..or, well, invalidation....)_

PR #2163, opened 2025-08-29
feat(http): mask secrets in api responses

zze0s:
Mask secrets and sensitive strings in api responses such as indexer keys, download client password etc.


ppfeister (approved):
LGTM.
Appears to upgrade cleanly, and checked values are redacted in scrape.

In an **ideal** world, the redaction would be a fixed-length string to not show the length, but this is more than likely sufficient for the task at hand.

ppfeister:
Non-blocking caveat: IRC auth commands not using SASL/ns are unredacted. However, this would require a bit more work to do properly, since you kinda need to see the command to edit it

PR #2165, opened 2025-09-01
feat(indexers): DarkPeers parse freeleech

jschavey:
adds support for parsing freeleech tags to darkpeers

### DIFF
--- a/internal/indexer/definitions/darkpeers.yaml
+++ b/internal/indexer/definitions/darkpeers.yaml
@@ -51,7 +51,7 @@ irc:
     type: single
     lines:
       - tests:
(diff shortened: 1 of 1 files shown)


internal/indexer/definitions/darkpeers.yaml
```diff
@@ -60,8 +60,9 @@ irc:
               torrentSize: 4.79 GiB
               baseUrl: https://darkpeers.org/
               torrentId: "1704"
+              freeLeechPercent: "100"
```
zze0s, line 63: This and the ones in `vars` on line 74 needs to be `freeleechPercent` instead. Minor difference with the lowercase l but it make a difference.

zze0s (changes requested):
Hi! Thanks for the PR! It needs one change before we can merge 😃

internal/indexer/definitions/darkpeers.yaml
jschavey: thank you I see you already made the changes for me so resolving this

zze0s (approved):

PR #2171, opened 2025-09-03
feat(filters): add Toggle All button

luckylittle:

* Request in Discord `#suggestions` channel

##### Add

* Option to enable/disable all filters

* `web/src/screens/filters/List.tsx`

1. `go run cmd/autobrr/main.go`
2. `pnpm dev`

#### Risk involved?

* Unsure

Screenshot_20260902_224143
Screenshot_20260902_224135
Screenshot_20260902_224123

* No


zze0s:
Not opposed to the idea and the implementation is decent for being frontend only, but It looks a bit off :thinking: Needs some more work before merge.

Not sure exactly what, maybe @s0up4200 or @martylukyy have some feedback?

martylukyy:
A checkbox in this way should be indicative of what is currently selected and **not** what is currently enabled or disabled.
I am not a fan of mix and matching checkboxes and switches that way.

If we decide to implement something like this i would rather like to see a button where we can enter a multi select mode that inserts checkboxes left to the switches to enable or disable a subset of filters.
That would add a lot more complexity though of course.

As an alternative i would propose 2 buttons to enable and disable all filters.
Here is a quick sketch of what i have in my mind:
image

martylukyy:
The two new buttons occupy approximately the same vertical screen estate as the new line in the filter list. Depending on the margins, there is a point to be made that this will even consume less vertical screen estate than the new line in the filter list. Narrowing down a single line in the filter list will look awkward imo.

Personally, i cannot think of another global action that we could add for the filter list but if you have proposals, please feel free to name them.

While i agree with your last point I would still like to see a solution where we can control a subset of filters like proposed in my previous comment if we decide on doing a checkbox solution.

With the newly added changes shown in the last draft i have the apprehension that people will think that they can change a "global state" and not the state of each filter which could lead to complaints.

I hope i covered all of your points and you see where we are coming from.

luckylittle:
FYI Windows 11 toggles behave exactly the same way as I've shown above - screenshot:

Microphone toggle

luckylittle:
> > FYI Windows 11 toggles behave exactly the same way as I've shown above - screenshot:
> > Microphone toggle
> 
> For indexers and IRC where the damage is limited, sure (which aligns to your screenshot). For filters though (true user workloads)? It's going be impossible to remember all 97 filters states. There is a place for this, I just don't immediately see the correct way to implement this for where you're asking for.
> 
> The only correct way I can see is some sort of template stacking which has been discussed for years. I'm not sure that will be implemented anytime soon.

Respectfully, I think you are over-thinking it. This is actually pretty simple and straightforward and does not have to be over-engineered. The global toggles work the same way everywhere - at least what I have seen in different products so far. It is intuitive and I would absolutely expect that all 97 filters will be turned on when I click on it (assuming the default global toggle is "off") - that's the whole point.

zze0s:
Don't worry, I have a solution in mind. Can hopefully spends some time next week to implement it.

zze0s (approved):

PR #2172, opened 2025-09-04
feat(list): include year for mdblist

luckylittle:

Problem statement: When you use MDBList to populate **MOVIES / SHOWS** or **MATCH RELEASES**, it only contains title at the moment. It leads to undesirable results, when e.g. movie(s) with the same title were released on different years. The MDBList json does contain the `release_year` already, so we can use that to make the matching more precise. See the schema of MDBList item:

```json
  {
    "id": 1061474,
    "rank": 2,
    "adult": 0,
    "title": "Superman",
    "tvdbid": null,
    "imdb_id": "tt5950044",
    "mediatype": "movie",
    "release_year": 2025
  }
```

##### Add

* New toggle option "Include year", that will turn the above example to `Superman*2025*`, instead of just `Superman`

* `internal/database/list.go`
* `internal/database/postgres_migrate.go`
* `internal/database/sqlite_migrate.go`
* `internal/domain/list.go`
* `internal/list/process_list_mdblist.go`
* `web/src/forms/settings/ListForms.tsx`
* `web/src/types/List.d.ts`

* `go run cmd/autobrr/main.go` & `pnpm dev` - go to `http://<AUTOBRR>/settings/lists` and add a new MDBList, enable the Include Year option + Match Release and look at the results.

#### Risk involved?

* Yes, if the Match Release is not ticked, this will not work. 

* TBA

* Yes

### DIFF
--- a/internal/database/list.go
+++ b/internal/database/list.go
@@ -45,6 +45,7 @@ func (r *ListRepo) List(ctx context.Context) ([]*domain.List, error) {
 		"include_unmonitored",
 		"include_alternate_titles",
 		"skip_clean_sanitize",
+		"include_year",
 		"last_refresh_time",
 		"last_refresh_status",
 		"last_refresh_data",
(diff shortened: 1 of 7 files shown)


luckylittle:
I've completed & tested backend against multiple lists (see below) and it now works as expected. It can only work with MDBList as [Trakt lists don't have the year data in them](https://api.autobrr.com/lists/trakt/upcoming-movies). Another thing I realized is it would not match against TV Shows as they usually don't have the year information as a tag so I am doing another condition to do that only for movies.

1. Test [movie](https://mdblist.com/lists/garycrawfordgc/netflix-movies/json) list 🆗 (`*The?Thursday?Murder?Club?2025*`)
2. Test [show](https://mdblist.com/lists/hdlists/latest-tv-shows/json) list 🆗 (`*The?Terminal?List*Dark?Wolf*`)
3. Test [mixed](https://mdblist.com/lists/luckylittle/external/30636/json) list (movie/show) 🆗 (`*The?Bikeriders?2024*` but `*Mammals*`)
4. Test list with [null](https://mdblist.com/lists/luckylittle/external/30596/json) year 🆗 (`*Coyote?vs*Acme*`)

**The only thing that is missing is minor UI improvement. What I would like to do is automatically enable the "Match Release" when you enable "Include Year" as seen here:**
Screenshot from 2025-09-05 12-48-33

Can someone help, please?

internal/database/list.go
```diff
@@ -45,6 +45,7 @@ func (r *ListRepo) List(ctx context.Context) ([]*domain.List, error) {
 		"include_unmonitored",
 		"include_alternate_titles",
 		"skip_clean_sanitize",
+		"include_year",
```
zze0s, line 48: Minor nickpick but can we move this up above skip so the includes are next to each others? 😄

zze0s (commented):
This is a real nitpick but I'd like to see the ordering updated of the new field to be before skip_clean_sanitize where possible 😄 

I only commented on one, but it's basically on all changes.

I also tried to figure out if we could grab season from mdblist but it didn't look like that was possible without having it pull metadata from somewhere else per item and process it. Maybe we can do that in the future.

zze0s:
Updated the ui to auto set Match Release when Include Year is enabled.

[Screencast_20250908_164907.webm](https://github.com/user-attachments/assets/8f1ce841-6750-4860-8601-8b0d5ba331a5)

zze0s (approved):
Now it's perfect! Thanks @luckylittle!

PR #2178, opened 2025-09-05
feat(http): make CORS settings configurable

zze0s:
Make cors settings configurable.


zze0s:
It defaults to wildcard now as it did before, but overwrites that when it's set in the config/env vars.

PR #2191, opened 2025-09-18
fix(irc): Soju bouncer nick check

winterberryteastories:
Instead we should probably also accept those messages when they include the wildcard character.


winterberryteastories (commented):
That makes a lot of sense!

I just tested it and it worked :)

PR #2225, opened 2025-10-15
fix(filters): improve filter limits handling

### LINKED ISSUE #1845: Rapid-fire announcements in IRC can cause a filter's maximum download limits to be exceeded
**Version**
1.51.0

**Describe the bug**
When multiple announcements occur almost-simultaneously on an IRC channel, a filter's maximum download limits can be exceeded.

**To Reproduce**
There seem to be a few trackers that reproduce this problem most easily. The one I've noticed it happening on the most is Anthelion (ANT).
1. In Settings, set up Anthelion as an indexer, linking it to a valid Anthelion IRC user account.
2. Create a filter that listens to the indexer that was just set up. Limit it to 1 download per day. In the 'Advanced' tab, set it to only snatch freeleech torrents (by enabling the freeleech toggle rather than setting the freeleech percentage).
3. Enable the filter.
4. Wait for the filter to grab a download. More often than not, when it fires, it will attempt to grab more than one.

**Expected behavior**
The expected behavior would be for the download limits set in the filter to be respected.

**Screenshots**
Image
Image

 - OS: macOS Sonoma 14.4.1
 - Browser: Firefox 132.0.2

**Smartphone (please complete the following information):**
Not applicable.

**Additional context**
I believe this may be happening as a result of a race condition, and consequently it may not always reproduce reliably. I have some additional thoughts, which I'll add separately.

zze0s:

Fixes #1845 

#### What's this PR do?

Implements a custom rate limit implementation with a token bucket per filter, where a filter acquires a token at the start of filtering and if it rejects it releases it and if it matches it grabs it.

On start we initialize all the filter counters if enabled in the filter.

##### Add

* Token bucket ratelimiter implementationfor filters

##### Change

* Skip/removes the CheckMaxDownloads in CheckFilter in favor for ratelimiter check before

* CheckFilter method of filter service

* TBD

#### Risk involved?

* Moderate?

* Should clarify how it works on the docs


zze0s:
Will be replaced by something else in the future.

PR #2349, opened 2026-02-19
feat(downloadclients): add new host parsing for Transmission

karaolidis:

No ticket open, but I noticed that autobrr was unable to communicate if transmission was behind a subpath.

##### Add

* New parsing method for Transmission host, similar to QBit.

##### Change

* UI to hide Port by default

* This is a generally small change, but I do wonder if we should completely remove the TLS and port fields in the UI?

* Apply patch on existing and new installation, and test that transmission via a subpath works as expected

#### Risk involved?

* Existing installations will need to update host to point to /transmission

* Later.

* No

### DIFF
--- a/web/src/forms/settings/DownloadClientForms.tsx
+++ b/web/src/forms/settings/DownloadClientForms.tsx
@@ -304,7 +304,7 @@ function FormFieldsRTorrent() {
 
 function FormFieldsTransmission() {
   const {
-    values: { tls }
+    values: { port, tls }
   } = useFormikContext<InitialValues>();
 
   return (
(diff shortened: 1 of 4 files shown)


web/src/forms/settings/DownloadClientForms.tsx
```diff
@@ -313,7 +313,7 @@ function FormFieldsTransmission() {
         required
         name="host"
         label="Host"
-        help="Eg. client.domain.ltd, domain.ltd/client, domain.ltd"
+        help="Eg. http(s)://client.domain.ltd, http(s)://domain.ltd/transmission, http://domain.ltd:9091/manage/torrents"
```
zze0s: Whats `/manage/torrents` here? Is that a built in method? Think it's best if it's not included if it's not something users expect.

Issue #1611, opened 2024-08-03
[FEATURE REQ] Allow disabling authentication (where external auth methods like Authelia/Authentik are used)


pentago:
**Is your feature request related to a problem? Please describe.**
Using external/proxy-level authentication methods like Authelia/Authentik currently results in a double login issue.

**Describe the solution you'd like**
Ability to turn off authentication via login form in scenarios where a single user exists and where external/proxy-level authentication methods are used.

**Describe alternatives you've considered**
The alternative is publicly exposing the autobrr login form, which is a privacy/security risk considering it only relies on password authentication. 

**Additional context**
N/A

zze0s:
Dupe of https://github.com/autobrr/autobrr/issues/483

Some more context here https://github.com/autobrr/autobrr/issues/979

pentago:
@zze0s as an alternative, would something like Calibre Web do would be possible?

They allow setting a HTTP header name in app config that allows authentication and Authelia ships a number of headers by default.

For example, upon successful login to Authelia form, they pass on User-Name header downstream which opens an app without login form.

Issue #483, opened 2022-10-01
Authentification proxy support


KptCheeseWhiz:
**Is your feature request related to a problem? Please describe.**
All my microservices are already behind Authelia, an authentification proxy, that already provides a way to authenticate users. Since there is no way to disable authentification in autobrr, I sometimes have to login twice, once through Authelia and another one in autobrr. Yes, this is not really an issue if I have to click on one more button if I haven't been on the page after one month, but it would still be a nice feature to have.

**Describe the solution you'd like**
I'd like the application to use the authentification proxy's headers to authentificate the user or at least, have an option to disable authentification.

**Describe alternatives you've considered**
There are no alternatives

**Additional context**
I implemented the authentification proxy on a [fork](https://github.com/KptCheeseWhiz/autobrr), but I am not sure if it fits the code requirements here (my first time coding in Golang). It works on my setup, up to you guys to see if you want this feature implemented on the main repo.

zze0s:
Hi

It's been discussed before and my stance on allowing auth proxies like Authelia has been that it needs to be more than just a "disable password" option, because people will see that and use it without thinking much, which then leads to unsecured and exposed instances of autobrr which isn't great.

This however, looks like it does a bit more than that, with some extra settings. Well done! And if it's your first time in Golang I'm highly impressed 😄

I need to check it out thoroughly and test it myself. I see one TODO about validating a header etc.

Some concerns etc:

* Will this only work with Authelia, or will it work with eg Authentik as well?
* How well does it work behind a reverse proxy? More headers etc needed?

So in conclusion - not opposed but it needs testing! 👍

davidnewhall:
In notifiarr, the operator switches from the "built in user" to auth proxy support by setting the username to `webauth`. The operator also has to set the `upstreams` list variable so it contains the CIDR or IP of the trusted upstream [auth] proxy. This makes sure the app only accepts the `x-webauth-user` header from designated IPs or networks. Then, for a user to be logged in, the upstream proxy has to pass the header `x-webauth-user: <username>`. This configuration requires the operator to know a little bit about their network, and properly configure the header in their proxy. It supports any auth proxy that can pass the username in a header (which is pretty much all of them).

I believe all the most relevant code can be found in this file: https://github.com/Notifiarr/notifiarr/blob/main/pkg/client/handlers_gui.go

Good luck!

TMD20:
I was hoping that all the arr clients moving to force auth would support oauth.

PR #2351, opened 2026-02-19
feat(auth): add option to disable local auth

karaolidis:

#483, #1611, #1664, partial #2292

#### What's this PR do?

I am well aware that this PR is probably getting denied, and rightfully so. It is a dangerous option to provide the users with. However, hear me out :)

The other -arr services provide a way to disable authentication, and provide a way to declare an API key without needing to authenticate first. Autobrr does neither. This becomes a problem for declarative setups such as Nix (my own homelab setup can be found [here](https://git.karaolidis.com/karaolidis/nix/src/branch/main/hosts/jupiter/users/storm/configs/console/podman/media/autobrr/post-start.sh), for example).

I am of the opinion that we should have the option, even if undocumented. If somebody is savvy enough to dig into the code and discover it, they most likely know what they are doing.

##### Add

* No Auth Mode

* Let's discuss first whether this PR has any chance of getting merged, and if the change should be documented. If so, the changes themselves are fairly straightforward.

* Disable auth and test that actions are available normally.

#### Risk involved?

* The obvious risks associated with allowing people to disable auth.

* Nope.

* See above.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -119,6 +119,11 @@ sessionSecret = "secret-session-key"
 # Disable Built In Login Form (only works when using external auth)
 #oidcDisableBuiltInLogin = false
 
+# Disable authentication entirely
+#
+# Do not use unless your instance is behind a secure proxy/firewall!
+#disableAuth = false
+
 # Metrics
 #
 # Enable metrics endpoint
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -131,6 +131,11 @@ sessionSecret = "{{ .sessionSecret }}"
 # Disable Built In Login Form (only works when using external auth)
 #oidcDisableBuiltInLogin = false
 
+# Disable authentication entirely
+#
+# Do not use unless your instance is behind a secure proxy/firewall!
+#disableAuth = false
+
 # Metrics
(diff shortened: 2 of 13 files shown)


KyleSanderson:
This already exists, just set the cookie secret to something static and issue your own.

onedr0p:
@karaolidis you might want to model this after the feature that landed in qui, you might have some success with doing it that way.

https://github.com/autobrr/qui/pull/1464

zze0s (changes requested):
I've been vocal about this before - I personally hate this.

But as many times before I have caved and implemented features I don't care for, or about.

All those extra none auth checks everywhere makes things a mess as well.

If this is going to be implemented I want multiple config flags that the end user has to acknowledge and answer to really drive home that they think this is a good idea.

CIDR ranges should be required and I'd go as far as saying to not allow 0.0.0.0/0 because that's just lazy.

I'd almost want an encrypted value users have to ask a key for :)

karaolidis:
I'm closing this PR as I currently don't have enough time to implement it. To anyone that comes across this, feel free to pick it up :)

PR #2356, opened 2026-02-24
fix(indexers): update HappyFappy urls

nixielectra:
##### Change

* Update HappyFappy indexer to support new domain

* urls/baseUrl

*

#### Risk involved?

* Low

image

first is the new .net
second is the old .org


zze0s:
FYI you can update the url in indexer settings if you start to type in the field.

Use that as a bandaid solution until there's a new release out later this week.

zze0s (approved):

PR #2416, opened 2026-04-05
feat(web): add Russian language

NoeRi9n:
##### Add

* Russian language support

* First json file

* Make sure it builds and that the language can be changed in settings.

#### Risk involved?

* Low

* No

#### Footnote

* **Note on language names:** 
In my `ru/common.json` file, I used the native endonym for Russian (Русский) as well as native names for the other languages, matching the pattern established in the `zh-CN` file. However, I noticed that `en/common.json` currently translates language names into English (e.g., "Simplified Chinese" instead of "简体中文"). Using native names everywhere is usually better for UI clarity so users can always find their language. I kept the existing pattern in the `en` file for this PR to avoid scope creep, but let me know if you'd like me to update the ru file  according to `en/common.json` file.


zze0s (approved):
Thanks for the PR! Much appreciated!

PR #2421, opened 2026-04-06
feat(auth): add PKCE support to OIDC implementation

oynqr:

* Requested once at https://github.com/autobrr/autobrr/pull/1853#issuecomment-2563507905

#### What's this PR do?

This PR adds PKCE support to the OIDC implementation, which is by default required by some IDPs like kanidm. The used oauth2 library already has support for PKCE, it just had to be wired up.

* Reading relevant Oauth2 specs and go library docs, verifying modified auth flows are correct

* Test OIDC login flows with PKCE disabled and enabled on the IDP side

#### Risk involved?

* High (potential login failures for existing users)

* No

### DIFF
--- a/internal/auth/oidc.go
+++ b/internal/auth/oidc.go
@@ -219,13 +219,27 @@ func (s *OIDCService) OAuthExchange(ctx context.Context, code string, opts ...oa
 	return s.oauthConfig.Exchange(ctx, code, opts...)
 }
 
-func (s *OIDCService) OauthAuthCodeURL(state string) string {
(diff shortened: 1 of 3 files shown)


zze0s (approved):
Thanks for the PR!

I tested it with pocketid to make sure that still works, and it does :+1:

oynqr:
Thanks for the quick review/merge! Sorry for the mess with the force pushes, I had wrongfully assumed that the Nix package I used for testing would have run the test suite.

I'll also look into adding this feature to the other applications.

PR #2441, opened 2026-04-27
feat(macros): expose release RawVars

sudo2k16:
# Description

Indexer definitions capture vars beyond what autobrr maps to known release fields (e.g. `tmdb`, `releaseName`, `doubleUp`). These were previously silently dropped and couldn't be used in webhook payloads or exec args.

- `Release.RawVars` - stores the full raw `varMap` from the indexer definition after `MapVars`, not persisted (`json:"-"`)
- `Macro.RawVars` - exposes them in webhook/exec/notification templates, e.g. `{{ .RawVars.releaseName }}` or `{{ toRawJson .RawVars }}`

## Type of change

## How has this been tested?

* `go build ./...` and `go test ./internal/domain/...` pass, including new `test_raw_vars` macro parse case
* Manual check: add a webhook action with `{{ toRawJson .RawVars }}` on an indexer that captures extra vars and verify the payload contains them

## Checklist


zze0s (commented):
We've already discussed this one a bit, but I do have some reservations about this.

I think it's one of those features that will be hard to change in the future if we don't want to break anything.

I feel like `Vars` is a bit too generic, but `AnnounceVars` is too limited and specific. 

As it sits atm in `internal/domain/release.go:1272` there's quite a bit of parsing that happens on each step before that could be useful even if those are available as macros.

I assume users will want to filter on these raw values at some point and to support that properly we'd have to revamp the filters, which is a long term goal.

So atm I'll leave this open, and merge your other PR for RocketHD.

zze0s (approved):
Changed it from `Vars` to `RawVars` and that's good enough 👍

PR #2454, opened 2026-05-11
build(ci): use web package manager for pnpm setup

s0up4200:

* N/A

##### Change

* Uses `pnpm/action-setup` with `package_json_file: web/package.json` in the web dependency workflows.

This is necessary because the workflows were resolving pnpm from the root/Corepack default before installing dependencies in `web/`, which let CI pick a newer pnpm store path than the version pinned by `web/package.json`; using the web package manifest keeps pnpm setup and cache behavior on the same version.

* `.github/workflows/eslint.yml`
* `.github/workflows/codeql.yml`
* `.github/workflows/release.yml`

* Re-run the affected GitHub Actions workflows.

#### Risk involved?

* Low; workflow-only change that aligns pnpm setup with the existing web package manager declaration.

* N/A

* No

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -56,10 +56,12 @@ jobs:
       with:
         node-version: ${{ env.NODE_VERSION }}
 
-    - name: Set up corepack
-      run: npm install -g corepack@latest && corepack enable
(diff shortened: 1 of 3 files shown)


zze0s (approved):
🚀

PR #2480, opened 2026-05-31
fix(indexers): include group id in infourl for PTP

mynameisbogdan:
##### Change

* Adding group id to info URLs for PTP indexer, it's needed to reflect the same info URL across Radarr/Prowlarr/Jackett to get history grabs for pushed releases by autobrr when using interactive search.

#### Risk involved?

* Low

### DIFF
--- a/internal/indexer/definitions/ptp.yaml
+++ b/internal/indexer/definitions/ptp.yaml
@@ -190,5 +190,5 @@ irc:
           uploadVolumeFactor: 0
 
     match:
-      infourl: "/torrents.php?torrentid={{ .torrentId }}"
+      infourl: "/torrents.php?id={{ .groupId }}&torrentid={{ .torrentId }}"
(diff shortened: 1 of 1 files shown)


zze0s:
Do releases not get recognized as the same/existing without group id? :eyes: 

For the detail links it does a 301 redirect if you only supply the torrent id, so that's fine for browsers, but if it's needed for matching then I get it.

zze0s (approved):

mynameisbogdan:
It's a simple [info URL equals check](https://github.com/Sonarr/Sonarr/blob/8150aa22daee320faa643d586e7b768fdad77398/src/Sonarr.Api.V5/Release/ReleaseController.cs#L289) that fails due to `/torrents.php?id=1&torrentid=1` being different than what autobrr is pushing: `/torrents.php?torrentid=1`

And it's an improvement for links in general, as in case the torrent is deleted you could remove the torrentid and still get the torrents for that group.

zze0s:
I see! Looks like PTP is the only one using groupid that we didn't include it in the details url for, so all the other ones should be fine, but if you find any then please open a PR.

Thanks for the PR b!

PR #2494, opened 2026-06-29
feat(actions): improve webhooks and exec handling

zze0s:

* none

##### Add

* Add ordering to actions like external filters
* Add expected status to exec cmd action
* Add expected status, more options and err handling to webhook action

* db migrations

* Add a filter, add multiple actions, click the position buttons, save and try reorder again.

#### Risk involved?

* Low

*

* yes


seeARMS:
Concrete integration case for the expected-status/error handling in this PR: https://github.com/autobrr/qui/issues/680.

I reproduced on develop `33bacd80e156ece313e0299c8ce61c735d6ade32` that the current `runWebhook` returns nil for HTTP 200 + JSON `success:false/no_match`, HTTP 422, and HTTP 500. A tiny httptest server provides the response; no tracker or qui server is required. The HTTP-status part of this PR addresses a real gap.

For the qui integration, HTTP validation alone still leaves HTTP 200/no-match looking successful. It would help to coordinate a response contract or opt-in outcome predicate, then test the full filter chain: no-match and HTTP error should reach an eligible lower-priority acquisition action; added/already-present should stop there. A candidate/preflight match should not be treated as an addition.

I prepared and exercised a local v1.86.0 compatibility patch that checks HTTP status and scopes qui outcome parsing to its apply endpoint. The release-service regression uses the real native webhook action and a synthetic following SONARR action. This is evidence for the integration case, not a proposal to impose qui's response format on generic webhooks. The corresponding details are in the linked qui issue.

PR #2495, opened 2026-06-29
fix(irc): surface channel JOIN errors as connection errors

arjunrn:
## Summary

Register callbacks for IRC error numerics that indicate a failed channel join, and surface them as connection errors so the failure is visible to the user in the UI.

Handled numerics:
- `471` — `ERR_CHANNELISFULL`
- `473` — `ERR_INVITEONLYCHAN`
- `474` — `ERR_BANNEDFROMCHAN`
- `475` — `ERR_BADCHANNELKEY` (wrong or missing channel key)
- `477` — `ERR_NEEDREGGEDNICK` (must be identified with NickServ to join)

Previously these errors were silently dropped; the bot would appear to be running normally but not monitoring the channel.

## Test plan

- Unit tests covering `handleJoinError` for all five numerics, with and without a reason string, and for short/malformed messages.
- Manual: configure a channel with `+k` or `+i` and verify the connection error appears in the network status UI.

### DIFF
--- a/internal/irc/handler.go
+++ b/internal/irc/handler.go
@@ -301,6 +301,11 @@ func (h *Handler) Run() (err error) {
 	}
 	client.AddCallback("INVITE", h.handleInvite)
 	client.AddCallback("366", h.handleJoined)
+	client.AddCallback("471", h.handleJoinError) // ERR_CHANNELISFULL
+	client.AddCallback("473", h.handleJoinError) // ERR_INVITEONLYCHAN
+	client.AddCallback("474", h.handleJoinError) // ERR_BANNEDFROMCHAN
+	client.AddCallback("475", h.handleJoinError) // ERR_BADCHANNELKEY
+	client.AddCallback("477", h.handleJoinError) // ERR_NEEDREGGEDNICK
 	client.AddCallback("PART", h.handlePart)
 	client.AddCallback("PRIVMSG", h.onMessage)
 	client.AddCallback("NOTICE", h.onNotice)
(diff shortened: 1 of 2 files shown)


internal/irc/handler.go
```diff
@@ -577,6 +582,17 @@ func (h *Handler) handleNickServ(msg ircmsg.Message) {
...
+			// Alternate nick isn't registered; retry with account name so the "password accepted"
+			// notice fires and setAuthenticated() is called from there.
+			h.log.Debug().Msg("NickServ: alternate nick not registered, retrying with account name")
+			if err := h.Send("PRIVMSG", "NickServ", fmt.Sprintf("IDENTIFY %s %s", h.network.Auth.Account, h.network.Auth.Password)); err != nil {
+				h.log.Error().Err(err).Msg("NickServ: failed to send account-based IDENTIFY")
+			}
+		} else {
+			// No account configured; join channels without NickServ auth.
+			h.log.Debug().Msg("NickServ: alternate nick not registered, no account configured, proceeding unauthenticated")
+			h.setAuthenticated()
```
zze0s: Can you give a good rationale for this else branch that sets authenticated?

zze0s (commented):
Hi! Thanks for the PR!

What llm model/harness was used to debug and implement this, so I can gauge the level of hallucination to consider?

arjunrn:
@zze0s thanks for the quick review. You're right this is generated with LLMs. But I have taken some time to understand what's happening and why there were issues. Here's what I have been able to piece together so far:

1. The indexer setup creates an IRC connection configuration which includes a nickname. 
2. If there's an existing connection with the nickname the identify step fails.
3. The connection then attempts to use alternate nicknames which fail.

None of this information is surface except in the logs at the trace level. But this was not the primary problem in this case.

The actual issue was the the channel needs a channel key. This channel key is not present in the presets and needs to be entered manually. When the channel key is not present there are messages which indicate that in the response but are not displayed in the interface.

So I've updated the PR to only display the connection errors. This should be sufficient to help troubleshoot the same issue in the future. All changes related to registration have been reverted.

I'm also open to changing the implementation as needed. Let me know. And in case you still need to know the LLM used is Opus 4.8.

internal/irc/handler.go
arjunrn: This entire change has been reverted.

zze0s:
Gotta be careful these days :smile: Opus 4.8 does a good job overall :+1: 

Which indexer is this related to? We have a way to set channel password from indexer setup, so if there's an indexer that might have changed to such a setup we should fix it. The two existing ones that does it are BTFiles and AnimeWorld which set it like:

```yaml
    - name: channels.password
      type: secret
      required: true
      label: Channel Password
      help: Channel password
```

The IRC implementation sure still has some holes and debugging issues aren't super easy. Your change to just surface the errors is much cleaner :+1: 

I started on a big IRC refactor back in 2024 that I still haven't been confident enough to merge because it changes a bunch of things. https://github.com/autobrr/autobrr/pull/1709

arjunrn:
I've made a follow-up [PR](https://github.com/autobrr/autobrr/pull/2498) to update preset for the indexer.

zze0s:
Closing this in favour of #1709

It's a big refactor and now handles these events and more.

PR #2498, opened 2026-06-30
feat(indexers): make NickServ optional and channel password required for BIT-HDTV

arjunrn:
## Summary

- Makes `auth.account` (NickServ Account) optional for the BIT-HDTV indexer
- Makes `auth.password` (NickServ Password) optional for the BIT-HDTV indexer
- Adds mandatory `channels.password` (Channel Password) field for the `#bit-hdtv` channel


zze0s (approved):
🥳

PR #2359, opened 2026-02-28
feat(filters): add rolling-window support for download limits

thmro:
# Description

Download limits can now be enforced over a custom window: `150 per 3 days` instead of only per calendar hour/day/week/month.

- new `max_downloads_period` field multiplies the window size, and `max_downloads_window_type` selects `FIXED` (resets at calendar boundaries, the existing behavior and default) or `ROLLING` (sliding window of the last N units, enforced by parameterizing the window in SQL: `datetime('now', 'localtime', '-3 days')` / `CURRENT_TIMESTAMP - INTERVAL '3 days'`)
- existing filters are unaffected: defaults are `FIXED` with period `1`
- new `MINUTE` unit for `max_downloads_unit`
- the limit fields are now loaded on the announce path (`findByIndexerIdentifier`) and list view (`find`), not just `FindByID` - previously the new columns would have been a no-op where enforcement actually happens
- `release_action_status.timestamp` converted to `TIMESTAMPTZ` on postgres, so window comparisons are instant-correct when the app and database run in different timezones - previously limits could stop counting entirely (e.g. `TZ=America/Los_Angeles` app with a UTC postgres never matched the current hour)
- enforcement now counts only the window instead of also computing an unbounded all-time total, and postgres gets a composite `(filter_id, status, timestamp)` index enabling index range scans - fresh postgres installs were missing the `filter_id` index entirely (only upgraders got it via migration 23)
- `FilterDownloads` JSON reduced to a single `period_count` (the per-unit count zoo and `total_count` had no remaining consumers); reject reasons read `reached 150 per 3 DAY`
- filter list shows `Downloads: 127/150 per 3 DAY`; the filter form gains a window type select and an `Every` field shown for rolling windows, translated in all locales
- `autobrrctl` filter export includes the new fields
- migrations: sqlite `97`, postgres `87`. Postgres note: existing naive timestamps are reinterpreted in the session timezone during conversion; on setups where the app wrote a different zone, historical window counts can be off by that offset until they age out of the enforcement window (they were unusable before)

Supersedes #2500 - thanks @thewillwilson: the window-parameterized query approach and several tests are adapted from it. Built on @thmro's original PR.

Fixes #2089

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## How has this been tested?

* `go test ./...` and the integration suite (`-tags integration`) against sqlite and dockerized postgres, both green
[...]

PR #2500, opened 2026-06-30
feat(filters): add max_downloads_period for rolling N-day/week/month download limits

thewillwilson:
## Summary

- Adds a new `max_downloads_period` field (integer, default 1) to filters, acting as a multiplier for the existing `max_downloads_unit`
- Enables limits like "150 downloads per 3 days" instead of being restricted to per-day/week/month calendar windows
- Switches download count queries from fixed calendar windows to rolling time windows, which is more correct for rate limiting

## Motivation

Some trackers enforce unsatisfied torrent limits on a rolling N-day basis rather than a per-calendar-day basis. For example, MaM allows 150 unsatisfied torrents every 3 days. Previously the only way to approximate this in autobrr was to divide — 50/day — which wastes headroom and doesn't match the tracker's actual rule. This change lets you set exactly 150 per 3 days.

## How it works

`max_downloads`, `max_downloads_unit`, and `max_downloads_period` work together:
- `max_downloads = 150`, `unit = DAY`, `period = 3` → allow 150 downloads in any rolling 3-day window
- `period = 1` (default) behaves identically to the previous per-day/week/month behaviour, just with a rolling window instead of a calendar reset

## Changes

- **Migration**: adds `max_downloads_period INTEGER NOT NULL DEFAULT 1` column (SQLite migration 93, Postgres migration 83)
- **Domain**: new `MaxDownloadsPeriod` field on `Filter` and `FilterUpdate`; `FilterDownloads` simplified to `PeriodCount`/`TotalCount`; helper methods for generating SQLite modifier and PG interval strings
- **Database**: `GetFilterDownloadCount` rewritten with rolling-window queries parameterised by period; all CRUD methods updated
- **Frontend**: new "every" number field between max downloads count and unit; list view shows e.g. "127/150 per 3 days"

## Relationship to #2359

This PR solves the same user need as #2359 but with a different implementation. The key difference is in how the N-period multiplier is applied:

**#2359** keeps the existing 1-unit SQL window and scales the *limit* by the interval — so "150 per 3 days" is enforced as "450 per 1 day". That produces different (and incorrect) behaviour for tracker rules like MaM's.

**This PR** expands the SQL window itself — `datetime('now', 'localtime', '-3 days')` for SQLite, `NOW() - $2::interval` for Postgres — so "150 per 3 days" genuinely checks whether fewer than 150 releases occurred in the last 3 days.

#2359 also has unresolved migration number conflicts with `develop` (its PG migration 82 and SQLite migration 92 are taken by the RoTorrent rename), and includes a stray `config.toml` change. This PR slots cleanly onto `develop` with migration numbers 83/93.


zze0s:
This is a dupe of https://github.com/autobrr/autobrr/pull/2359.

Have your agent compare the two implementations.

thewillwilson:
I did — they're not the same implementation. The key difference is in how the N-period multiplier is applied.

**#2359** has two hardcoded SQL queries (FIXED and ROLLING). The rolling query uses fixed intervals baked into the SQL — `-1 day`, `-1 hour`, `-7 days` etc. — regardless of what interval the user sets. The interval value is never passed to SQL at all. Instead it's used purely in Go to scale the *limit*. From [lines 80–81 of #2359's `filter.go`](https://github.com/thmro/autobrr/blob/ab57e60c5de5a797b741304bc350dcc96d26f1b8/internal/domain/filter.go#L80-L81):

```go
// Proportional calculation: allow maxDownloads * interval downloads
return count < (maxDownloads * interval)
```

So with `max_downloads = 150`, `unit = DAY`, `interval = 3` and ROLLING selected:
- SQL counts releases in the last **1 day** (hardcoded in the query)
- Go checks `day_count < (150 * 3)` → `day_count < 450`

You could download 449 things today and not be blocked. That's not 150 per 3 days — it's 450 per day.

**This PR** parameterises the SQL window itself. `max_downloads_period = 3` with `unit = DAY` produces `datetime('now', 'localtime', '-3 days')` for SQLite and `NOW() - $2::interval` with `'3 days'` for Postgres. The query genuinely counts releases across the last 3 days and checks `count < 150`. After 150 downloads in any 3-day span, autobrr blocks further downloads — which is exactly how MaM's rule works.

#2359 also has unresolved migration number conflicts with `develop` (its PG migration 82 and SQLite migration 92 are already taken by the RoTorrent rename) and includes a stray `config.toml` change. This PR slots cleanly onto `develop` with migration numbers 83/93.

Happy to close if you'd prefer to fix #2359 instead, but wanted to flag the semantic difference first.

zze0s:
The intention of #2359 was to have the window dynamic, but it wasn't implemented correctly yet, so I think these will be the same.

I'll do a review of both :+1:

zze0s:
Superseded by #2359 that now has the same functionality.

Issue #2417, opened 2026-04-05
Notifications are sent for filters for which the notification is disabled


nothing2obvi:
**Version**
1.75.1

**Describe the bug**
Notifications are sent for filters for which the notification is disabled.

**To Reproduce**
1. Turn off global events
2. Set notifications only for filters A, B, C, D
3. Notification for filters A, B, C, D, E, F are sent

**Expected behavior**
Notifications are only sent for filters for which the notification is enabled.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Ultra.cc Seedbox
 - Browser: Chrome
 - Version Version 146.0.7680.178 (Official Build) (arm64)

zze0s:
Add your info here so we only have to track one issue for this https://github.com/autobrr/autobrr/issues/2235

PR #2511, opened 2026-07-10
fix(notifications): per-filter and muted filters

### LINKED ISSUE #2235: Per-filter notification mute does not always work
**Version**
`1.68.0 18d452cda028847eaf66ee5e49cc30a246707150` 

**Describe the bug**
Per-filter mute does not always work. I am still getting notifications.

**To Reproduce**
I am not 100% sure what causes this. One filter (`cross-seed`) is set to override global notification settings and mute the notification, however it is still sending notifications.

**Expected behavior**
When the filter is muted, it should not send any notifications.

**Screenshots**

* Global notifications:
Image

* Discord notifications setting - you can see two filters are muted:
Image

* Notification setting under the filter:
Image

* Example of notification in Discord that still goes through:
Image

 - OS: Linux 5.14.0-570.37.1.el9_6.x86_64
 - Browser Chromium
 - Version 135.0.6850.0 (Official Build) custom (64-bit)

**Additional context**
❗  Interestingly, this was working when Pushover was used. I have three notification services, but only one is enabled - it might be the culprit?❗

zze0s:

* [issue #2235](https://github.com/autobrr/autobrr/issues/2235) — Per-filter notification mute does not always work                                                                                 
* [issue #2417](https://github.com/autobrr/autobrr/issues/2417) — Notifications are sent for filters for which the notification is disabled                                                         

#### What's this PR do?                                                                                                                                                                             

Fixes per-filter notifications so mutes/overrides are respected reliably, and closes a latent crash in the notification service. Three root causes are addressed:                                   

1. **Editing a notification wiped its per-filter mutes/overrides** (primary cause of #2235). `Notification.filters` is an unexported map, so a notification decoded from an HTTP `PUT` always has a `nil` filters map. `Update`/`Store` rebuilt the live sender from that raw object, silently dropping every per-filter mute/override until restart. Senders are now always rebuilt from a canonical object whose per-filter events are reloaded from the DB.                                                                                                                                            
2. **Muting/overriding one notification silenced the others for the same filter** (#2235 + #2417). `Send()` used a single set-wide `hasFilterSpecific` flag: if any notification had a per-filter row for the filter, the global fallback was skipped for *every* notification. Each notification is now evaluated independently.                                                                     
3. **Unsynchronized maps could crash the daemon.** `Send()` runs synchronously on the release/action/IRC goroutine and ranged over `s.senders` (and read the shared `filters` map) with no lock, while HTTP handlers mutated them — a `concurrent map iteration and map write` fatal. Added an `RWMutex` + copy-on-write rebuilds.                                                                   

##### Add                                                                                                                                                                                           

* `notification.Service` `sync.RWMutex` guarding `senders`/`notifications`; `Send()` selects interested senders under `RLock` and releases it before any network I/O.                               
* `hydrateAndRegister` helper that rebuilds a fresh canonical `*domain.Notification` (copy-on-write) with per-filter events reloaded from the DB, keeping `s.notifications` and `s.senders` in sync on every write path.                                                                                                                                                                                
* `domain.Notification.Clone()` for the copy-on-write rebuild.                                                                                                                                      
* Regression + race tests: mute survives a notification edit, per-sender independent routing, freshly created notification is mutable, and a `-race` concurrency test.                              

##### Change                                                                                                                                                                                        

* `Send()` now evaluates every sender uniformly via `CanSendPayload` for both filter-scoped and global (`FilterID == 0`) events — a notification with a per-filter row uses its rule (incl. mute), and every other notification independently falls back to its own global events.                                                                                                                     
* `Store`/`Update`/`StoreFilterNotifications`/`DeleteFilterNotifications`/`registerSenders` rebuild affected senders from DB-hydrated objects; `Delete` now also clears `s.notifications`.          

##### Remove                                                                                                                                                                                        

* The two-phase set-wide `hasFilterSpecific` gate in `Send()`.                                                                                                                                      

* `internal/notification/service.go` — `Send()` (gate removed + locking) and the new `hydrateAndRegister` helper, then the `Store`/`Update`/`StoreFilterNotifications`/`DeleteFilterNotifications` write paths.                                                                                                                                                                                        
* `internal/domain/notification.go` — `Clone()`.                                                                                                                                                    

**Mute survives an edit (#2235):**                                                                                                                                                                  
1. Create a Discord notification with *Push Approved* enabled globally.                                                                                                                             
2. On a filter, toggle *Mute filter* for that Discord notification; confirm matches for the filter no longer notify Discord.                                                                        
3. Edit the Discord notification (e.g. rename it or change the webhook) and save.                                                                                                                   
4. Trigger another match for that filter → Discord stays muted (previously it started firing again).                                                                                                

**Muting one notification doesn't silence the others (#2417):**                                                                                                                                     
1. Enable two notifications (e.g. Discord + Telegram), both with *Push Approved* globally.                                                                                                          
2. Add a per-filter mute for **Telegram only** on a filter (leave Discord with no per-filter config).                                                                                               
3. Trigger a match → Discord still fires; Telegram is muted (previously Discord was suppressed too).                                                                                                

**Stability under load:** save notifications/filters repeatedly while releases are being processed — no `concurrent map iteration and map write` panic. Covered automatically by `go test -race -run TestService ./internal/notification/...`.                                                                                                                                                           

#### Risk involved?                                                                                                                                                                                 

* **Behavioral change (intended):** per-filter scoping is now per-`(filter, notification)` rather than filter-wide. If a user previously relied — likely unknowingly — on one notification's per-filter row suppressing global fallback for *other* notifications, those other notifications will now fire for that filter. This matches the documented two-tier intent and the expectation in both issues.                                                                                                                                                                                        
* **Locking:** `Send()` holds only an `RLock` for the selection loop and releases it before network I/O, so there's no added latency on the send path. DB reads during a config change happen under the write lock, briefly blocking `Send` selection; config changes are infrequent user actions.                                                                                                      
* Backend only; no schema/migration changes.                                                                                                                                                        

* No.


zze0s:
Superseded by #2629

PR #2530, opened 2026-07-18
feat(auth): support disabling authentication behind a trusted reverse proxy

arjunrn:
# Description

qui introduced disabled authentication for trusted subnets [here](https://github.com/autobrr/qui/pull/1464). I have followed the same pattern to allow disabling authentication with the same extra environment variable to emphasize this is dangerous. But my implementation varies a little bit. The frontend does not display the user page when authentication is disabled. Also there's no logout button. 

I'm open to changing the implementation to meet your preferences. I had also started a [discussion](https://github.com/autobrr/autobrr/discussions/2505) to get the maintainers input first.

## Type of change

## How has this been tested?

- Unit tests for configs:
  - Ensure that when the authentication is disabled user also acknowledges the risk.
  - Ensure CIDRs are always specified and a valid
- Unit test for middleware.
- Manual testing:
  - Verified for failure when auth disabled and CIDR not specified
  - Verified that auth only disabled when both danger acknowledged and disable explicitly set.
 
image
image
image

## Checklist

## AI disclosure

Claude(Opus 4.8) was used to examine the existing functionality in qui and replicate it. But several edits/improvements have been made after. Also the code has been reviewed repeatedly.


internal/config/config.go
```diff
@@ -447,6 +451,36 @@ func (c *AppConfig) loadFromEnv() {
 	if v := GetEnvStr("METRICS_BASIC_AUTH_USERS"); v != "" {
 		c.Config.MetricsBasicAuthUsers = v
 	}
+
+	// Auth disabled configuration
+	if v := GetEnvStr("AUTH_DISABLED"); v != "" {
+		c.Config.AuthDisabled = strings.EqualFold(strings.ToLower(v), "true")
+	}
+
+	if v := GetEnvStr("I_ACKNOWLEDGE_THIS_IS_A_BAD_IDEA"); v != "" {
```
zze0s: This should probably have a the `AUTH_` prefix, and rename/change the semantics to `AUTH_DISABLED_ACKNOWLEDGEMENT` or something similar, and then the value can be the full `I_ACKNOWLEDGE_THIS_IS_A_BAD_IDEA`.

internal/config/config.go
```diff
@@ -289,6 +289,10 @@ func (c *AppConfig) defaults() {
 		MetricsHost:           "127.0.0.1",
 		MetricsPort:           9074,
 		MetricsBasicAuthUsers: "",
+
+		AuthDisabled:               false,
+		IAcknowledgeThisIsABadIdea: false,
```
zze0s: Prefix with `Auth` and change to `AuthDisabledAcknowledgement` acknowledgement and change type to `string`.

internal/domain/config.go
```diff
@@ -43,6 +50,85 @@ type Config struct {
 	MetricsHost             string `toml:"metricsHost"`
 	MetricsPort             int    `toml:"metricsPort"`
 	MetricsBasicAuthUsers   string `toml:"metricsBasicAuthUsers"`
+
+	// AuthDisabled disables all authentication when both AuthDisabled and
+	// IAcknowledgeThisIsABadIdea are set. Intended only for deployments behind
+	// a reverse proxy that handles authentication itself. Use IsAuthDisabled()
+	// to check whether auth is actually disabled.
+	AuthDisabled               bool     `toml:"authDisabled"`
+	IAcknowledgeThisIsABadIdea bool     `toml:"I_ACKNOWLEDGE_THIS_IS_A_BAD_IDEA"`
```
zze0s: Same as `internal/config/config.go`

internal/config/config.go
```diff
@@ -289,6 +289,10 @@ func (c *AppConfig) defaults() {
...
+		AuthDisabledAllowedCIDRs:   []string{},
```
zze0s: Thoughts on making this just `AuthAllowedCIDRs`? Then it could be used to also restrict access in general? Need to think about this a bit more.

zze0s (commented):
I will once again say I'm not a fan of being able to fully disable auth, but alas.

internal/config/config.go
arjunrn: Actually this setting is the one I don't see much value in. I have it here just to make it match with other projects. The implementation in this PR and I suspect in other projects is to simply validate that the IP of the caller is in a particular CIDR, which in most cases is the IP address of reverse proxy. But if something can reach your app from outside your "trusted" network you have larger problems. The app should be listening on loopback only and there should be firewalling to prevent outside connections. I guess it's just there as a last ditch protection measure.

Does it add any value when there is proper authentication with OIDC setup? IMHO not really. Maybe there's some more value when it's basic auth.

internal/config/config.go
arjunrn: I agree, this is better and matches the other variables. Updated.

internal/config/config.go
zze0s: The problem is that it's so simple to just expose things to the internet for inexperienced users, as well as experienced ones that slip.

I'm not sure yet. Will give it some thinking :+1:

arjunrn:
@zze0s Just checking in to see if you're still interested in this PR. Happy to make any changes that you think would get this change in shape to get merged.

zze0s:
I started on a decent refactor of things but got busy with other stuff. Not abandoned :+1:

PR #2549, opened 2026-07-28
fix(indexers): TBy remove invite command

luckylittle:
# Description

It is not required to provide the invite command. In fact, it has not been working for a long time and the response is always `<erica> irc key or username is wrong!`. You just need to be identified with NickServ.

I believe due to this, AutoBrr still thinks it is not in the channel. This should fix the issue `join not confirmed`, even though it clearly is in the channel and can see the announces and it actually works.

## Type of change

## How has this been tested?

- Not tested

image

## Checklist

## AI disclosure

No AI was used.

### DIFF
--- a/internal/indexer/definitions/torrentbytes.yaml
+++ b/internal/indexer/definitions/torrentbytes.yaml
@@ -47,7 +47,7 @@ irc:
     - name: invite_command
       type: secret
       default: "erica letmeinannounce USERNAME IRCKEY"
-      required: true
+      required: false
       label: Invite command
       help: Invite auth with Erica, please replace USERNAME and IRCKEY.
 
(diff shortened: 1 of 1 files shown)


internal/indexer/definitions/torrentbytes.yaml
```diff
@@ -47,7 +47,7 @@ irc:
     - name: invite_command
       type: secret
       default: "erica letmeinannounce USERNAME IRCKEY"
-      required: true
+      required: false
```
zze0s: The required here is really only for the UI. Nothing in the irc handler checks for required, it's simply based on if invite_command is empty or not.

Does it work as expected if you remove the invite command?

luckylittle: My invite command is empty and it is still showing "join not confirmed". I have not tested it with `required: false`.

zze0s: Okay, then it's probably better to fully delete the invite command setting from the definition.

zze0s (commented):
Pull the image and try with this. It should start working.

PR #2554, opened 2026-07-29
fix(indexers): update TorrentHR pattern

senseiriksha:
# Description

Updated the matching pattern for the TorrentHR UNIT3D platform and adjusted the related test coverage accordingly.

## Type of change

## How has this been tested?

Created a custom definition and successfully received a success pattern on the first successful upload.

## Checklist

- [ ] I ran `go fmt` and the test suite passes locally

## AI disclosure

No AI was used.

### DIFF
--- a/internal/indexer/definitions/torrenthr.yaml
+++ b/internal/indexer/definitions/torrenthr.yaml
@@ -58,14 +58,23 @@ irc:
       parse:
         type: single
         lines:
-          - pattern: '^\[.*?\]\s+~THR Novi torrent -> Ime: (?P<releaseName>.*?) \.:\. Uploader: (?P<uploader>.*?) \.:\. Kategorija: (?P<category>.*?) \.:\. Velicina: (?P<torrentSize>.*?) \.:\. URL: (?P<baseUrl>https?:\/\/.+\/)torrents\/(?P<torrentId>\d+)'
+          - pattern: '^\[Novi Torrent\] Ime: (?P<releaseName>.*?) \| KatID: (?P<categoryId>\d+) \((?P<category>.*?)\) \| TMDb: (?P<tmdbId>\d+) \| IMDb: (?P<imdbId>\d+) \| Veličina: (?P<torrentSize>.*?) \((?P<torrentSizeBytes>\d+)\)(?: \| HNR-Free: (?P<hnrFree>DA|NE))?(?: \| Freeleech: (?P<freeleech>\d+%))?(?: \| DoubleUpload: (?P<doubleUpload>DA|NE))?(?: \| Internal: (?P<internal>DA|NE))?(?: \| Personal: (?P<personal>DA|NE))? \| Uploader: (?P<uploader>.*?) \| URL: (?P<baseUrl>https?://.+?/)torrents/(?P<torrentId>\d+)$'
             tests:
-              - line: '[2026-07-29 16:48:14] ~THR Novi torrent -> Ime: A.torrent.name.here.WEBRip.x264-tina .:. Uploader: uploaderName .:. Kategorija: Filmovi/HD .:. Velicina: 6.70 GB .:. URL: https://www.torrenthr.org/torrents/000000'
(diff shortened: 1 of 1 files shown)


zze0s:
If you have any influence and can nudge staff to make changes, I'd highly recommend moving to a ultra dense format like this: https://github.com/autobrr/autobrr/blob/develop/internal/indexer/definitions/midnightscene.yaml#L55-L57

There's multiple UNIT3D indexers using my announce bot tool [UNIT3D-Announcer](https://codeberg.org/ze0s/UNIT3D-Announcer) for this, and is really easy to setup and improve the announce format with more information that benefits users.

senseiriksha:
I reached out to the tracker staff, and together we created a new announce pattern containing significantly more information, including optional fields that were not previously available.

The staff mentioned that their primary goal is to remain unique, so they decided to use their own format and solution rather than adopt the suggested format directly.

I hope this version is a better fit. I’ve pushed the changes in a new commit. If you have any further suggestions or questions, I’ll be happy to help.

zze0s:
Good to get more info in there but it will 100% run into irc message limits when you have a long release name.

Wanting to be unique for this vs using something that's proven to work, not the best of ideas.

Happy to help improve it, but my recommendation is like I posted earlier. You get the maxium info in there without wasting a lot of characters that can be used for a long release title.

Edit: we'll btw have to add extra parsing for DA and the opposite value whatever that is?

senseiriksha:
The tracker staff is aware of the IRC message length limit. Since HNR-Free, Freeleech, DoubleUpload, Internal, and Personal are optional and only appear when applicable, they believe announcements are unlikely to reach that limit under normal circumstances.
I should clarify that the example containing all optional fields represents a rare worst-case scenario. The announcement will normally look like this:
[Novi Torrent] Ime: Linux is the best 2026 1080p Platform WEB-DL DD+ 5.1 H.264-RLSGRP | KatID: 17 (Filmovi/HD) | TMDb: 000000 | IMDb: 000000 | Veličina: 3.44 GB (3694404116) | Uploader: THR | URL: https://www.torrenthr.org/torrents/000000

The announce channel is also actively followed by regular users and is not used exclusively for bot parsing. Therefore, the staff would prefer to retain a more descriptive, user-friendly format instead of switching to a fully positional numeric one.

Regarding the values mentioned in your edit, DA means YES, while the opposite value is NE, meaning NO. The optional fields may contain either DA or NE (except Freeleech), or the entire field may be omitted from the announcement.

Thank you very much for all the advice, recommendations, and warnings. They are greatly appreciated, and we will keep them in mind if the format needs further optimization.

If you have any additional recommendations or feedback you would like me to pass on to the tracker staff, please let me know and I’ll discuss it with them.

zze0s:
Alright, good. I'll add some field mappings so we don't have to touch any parsing code.

zze0s:
Added some additional announce tests, can you check them over if they are correct with the optional values?

The mappings should handle the translation to origin etc.

senseiriksha:
Thanks, the changes look good overall.

One small note regarding the additional tests and mappings for negative values: as far as I understand the THR announce format, optional fields are only shown when they are active/enabled. For example, Internal/Personal/DoubleUpload/HNR-Free should only appear when the value is “DA” / “Yes”, and Freeleech should only be relevant when it is above 0%.

Because of that, test cases such as `Internal: NE` are technically not expected to occur in real announces. They do not hurt, but they are probably not necessary since they cover a scenario that likely will not happen.

For the same reason, mappings for 'NE' are also not really needed. They can stay, but they should never be used in practice if THR only displays those optional fields when enabled.

One more semantic detail: 'Internal: DA' and 'Personal: DA' refer to two different things. Internal means an internal release group/release, while Personal means a personal release. Because of that, I do not think `Personal: DA` should be mapped to 'origin: INTERNAL' unless THR intentionally treats personal releases as internal-equivalent.

Everything else looks OK to me.

That said, even if it stays as-is, it should not break anything and should work fine.

zze0s:
Ah okay, good clarifications. We don't have a PERSONAL origin filter at this time and not sure we want to add it for this atm.

senseiriksha:
No problem at all. It should not cause any issues if it stays like this at this stage.

zze0s:
@senseiriksha can you check if the updated patterns are correct? More fields are optional now.

senseiriksha:
I’ve confirmed this with the tracker staff. When one of these optional flags has a negative value, the announcer omits the entire field. The fields are only included when their value is positive, in which case they are displayed as DA (or Freelech with 0% <). Therefore, your latest changes are correct. Thank you!

zze0s (approved):

PR #2583, opened 2026-08-07
fix(irc): gate NickServ auth on mechanism and stop flapping reconnects

zze0s:
# Description

TorrentLeech published an announcement (2026-08-07) after banning a wave of autobrr users whose bots were rapidly connecting and disconnecting, blaming the SASL/NickServ options in autobrr's IRC settings. Investigating the claim against our IRC handler turned up three real problems on our side, all fixed here:

1. **The NickServ IDENTIFY fallback was gated on password presence alone.** `Auth.Mechanism` was only consulted for SASL setup, so a network with Mechanism "None" and leftover credentials still sent `PRIVMSG NickServ IDENTIFY` on every connect - meaning TL's own remediation advice ("set the Mechanism back to None") did not actually stop the services traffic. The fallback is now gated through `IRCAuth.NickServEnabled()`: a password must be set **and** the mechanism must not be NONE. SASL_PLAIN keeps its long-standing NickServ fallback for when SASL does not complete (#333), and rows written via the API without a mechanism keep the historical behavior. On a network without services, a NONE network with stale credentials now sends nothing, and the state machine goes straight to joining channels instead of parking in Authenticating waiting for a NickServ reply that can never come. Stored credentials are deliberately kept when the mechanism is NONE so auth can be re-enabled later without re-entering them.

2. **The add-indexer wizard silently manufactured SASL configs.** Filling the (often optional) account/password fields a definition displays made the wizard hardcode `SASL_PLAIN`, even where the definition labels the fields "NickServ" - the user never saw or chose a mechanism. Indexer definitions can now declare the mechanism their network actually supports (`irc.auth.mechanism`, supported in both v1 and v2 definition formats), and the wizard uses the declared value, keeping `SASL_PLAIN` as the default only when a definition declares nothing. TorrentLeech, AnimeBytes, Aither and Zenith (all NickServ-only networks) now declare `NICKSERV`, and TorrentLeech's auth field help texts are marked OPTIONAL.

3. **A reconnect loop with no exit.** The irc-go client reconnects a dropped connection every 15 seconds forever, with no attempt cap or backoff growth. Recognized fatal failures already stop the network (SASL 904, 465 ban, TLS verification, bad NickServ credentials), but any unrecognized failure mode - e.g. a server that kills the bot shortly after registration each time - would cycle indefinitely at ~4 connects/minute until the user noticed. `onDisconnect` now runs a circuit breaker: 5 consecutive sessions shorter than 30 seconds stop the network, surface the reason in the UI, and send a notification. A session that lives longer resets the count, user-initiated stops/restarts never count, and a manual restart opens a fresh window. The breaker covers the post-registration path (irc-go only fires disconnect callbacks for registered connections); registration-time failures remain bounded by the existing 25-attempt connect backoff and the in-band fatal handlers.

For context on the original complaint: no autobrr version can cycle faster than the 15-second reconnect floor, and v1.82.0 already stops the network outright on SASL failure. The realistic pre-1.82 worst case was short-lived connections at a 15+ second cadence; this PR removes the remaining ways a misconfigured or unclassified failure can generate sustained connection churn at all.

## Type of change

## How has this been tested?

* New unit tests for the mechanism gate (`TestAuthenticateGatesNickServOnMechanism`, `TestOnConnectedSkipsAuthenticatingWhenMechanismNone`) and the circuit breaker (trip at threshold, streak reset by a long session, manual stops excluded). All new tests were verified to fail against the pre-fix code.
* New end-to-end test against the in-process ircd (`TestNoneAuthWithLeftoverPasswordSkipsNickServ`): a NONE network with leftover credentials joins, reaches Monitoring, and the server records zero messages to NickServ (new `NickServMessageCount` counter on the test ircd). Full `go test -tags=irc_integration_test ./test/irc/...` passes with `-race`, including the existing SASL/NickServ/None auth flows.
* Definition schema round-trip tests (`TestDefinitionIRCAuthMechanism`, `TestDefinitionIRCAuthAbsent`) covering both the v1 compatibility conversion and v2 decoding; `TestIndexerYamlExpectations` now also rejects invalid declared mechanisms in bundled definitions.
* `go build ./...`, `go vet`, `gofmt`, and the `internal/irc`, `internal/domain`, `internal/indexer` suites with `-race` all pass. Frontend `pnpm --dir web build` (tsc) passes; lint findings are pre-existing in files this PR does not touch.

## Checklist


zze0s:
Replaced by smaller individual PRs #2605 #2606 #2604

PR #2507, opened 2026-07-07
feat(indexers): handle deprecations

zze0s:
# Description

When a tracker shuts down and we delete its definition, the configured indexer becomes orphaned: it stays selectable in filter dropdowns, stays attached to existing filters, and its historical releases lose their friendly name. Retiring one also meant hand-writing a per-removal SQL migration.

This adds a first-class deprecation path: bundled YAML tombstones + a startup reconcile that soft-deletes (archives) orphaned indexers, plus UI to see and clean them up. Nothing is destroyed - archiving is reversible, releases keep their name.

Fixes #2481

**Tombstones**

* Retiring an indexer is now: delete the active definition, add `internal/indexer/definitions/deprecated/<identifier>.yaml` (identifier, name, reason, issue URL, optional `alias_of`, date). No SQL migration.
* Backfills tombstones for the 14 indexers already removed from `develop` (`fnp`, `torrentseeds`, `uhdbits`, `ptn`, …).
* Loaded from the existing `//go:embed definitions` FS with strict validation (`KnownFields`, filename must match identifier, reason/issue URL/date required, no duplicates).
* `scripts/check-indexer-deprecations.sh` runs on PRs in `release.yml`: a removed or renamed identifier with neither an active definition nor a tombstone fails CI.
* CONTRIBUTING.md gains a "Retiring an indexer" section; PR template gains an "Indexer definition update" type.

**Schema**

* `indexer.archived` / `indexer.archived_at` + `indexer_deprecation` table - sqlite `95`, postgres `85`, plus `current_schema` updates.

**Startup reconcile**

* `reconcileDeprecations()` runs before indexers are mapped: upserts tombstone metadata, archives rows with no live definition, and revives rows whose definition came back (re-added custom definition). Idempotent, single transaction.
* Logs a warning for any configured indexer with no definition *and* no tombstone - a tripwire for a removal that skipped the process.

**Enforcement**

* `StoreIndexerConnections` / `StoreIndexerConnection` resolve archived state from the DB and refuse to attach an archived indexer - the single choke-point covering save, duplicate and import.
* `filter.Store` and `filter.Duplicate` now validate indexers too (previously only `Update`), rejecting unknown and archived ids with `400`.
* Indexer `Update`, `Delete`, `TestApi` and `ToggleEnabled` reject archived rows; `findByIndexerIdentifier` skips them so an archived indexer can never match a release.

**API**

* `GET /api/indexer/deprecations` - tombstone metadata + `filter_count` per identifier.
[...]

PR #2585, opened 2026-08-08
chore(indexers): deprecate MoreThanTV

s0up4200:
# Description

MoreThanTV closed. The site is gone, and the IRC announce channel no longer has anything to announce, so the definition only adds a dead entry to the indexer list. Proof: https://reddit.com/r/trackers/comments/1vj7lpf/mtv_closing/

Users with a saved `morethantv` indexer are not broken by this. `mapIndexer` returns `ErrIndexerNotFound` and the caller skips the row, so the indexer disappears from the setup and the orphan database row stays. No migration is included.

## Type of change

- [x] Indexer deprecation

## AI disclosure

Claude Code (Opus 5) removed the definition file and wrote this description. The change is one file deletion.


fabricionaweb:
RIP 😭

zze0s:
Will merged after #2507 is done.

zze0s (approved):

PR #2587, opened 2026-08-09
fix(irc): accept join confirmation received before the join workflow starts

s0up4200:
# Description

Some servers force-join the bot into the announce channel immediately on connect (seen on Norbits, InspIRCd). The join confirmation (`RPL_ENDOFNAMES`) then arrives while the channel state machine is still `Idle`, and `Idle -> Monitoring` was not a valid transition. The confirmation was rejected, the state machine sent its own `JOIN` for a channel it was already in, and the server dropped the duplicate silently. No second confirmation came, so the 45s timer failed the channel into the error/backoff loop until it gave up. The UI showed "join not confirmed" while announces flowed the whole time.

The fix adds the missing `Idle -> Monitoring` edge, the same recovery edge that `Error`, `Kicked`, and `InviteFailed` already have. `Start()` now also returns early when the channel is already `Monitoring`. Before, that path was a no-op that logged an "invalid state transition" error on every reconnect where the force-join won the race.

## Type of change

## How has this been tested?

* Diagnosed on a live instance with TRACE logs: the raw IRC exchange shows the server-initiated `JOIN` and `366` before the join workflow ran, the rejected `Idle -> Monitoring` transition, and the silent drop of the duplicate `JOIN`.
* New test `TestForceJoinBeforeStart` covers the sequence: confirmation while `Idle`, then `Start()` must not disturb the channel.
* New row in `TestChannelStateTransitions_Recoverable` pins the `Idle -> Monitoring` edge.
* Full `internal/irc` test suite passes.

## Checklist

## AI disclosure

Was any AI used in this PR? Roughly how much of the code was AI-written, and what model/tool?

Yes. The diagnosis and all of the code were written with Claude Code (Fable 5), directed and reviewed by me.

### DIFF
--- a/internal/irc/channel_state_machine.go
+++ b/internal/irc/channel_state_machine.go
@@ -81,6 +81,7 @@ var validChannelTransitions = map[ChannelState][]ChannelState{
 	ChannelStateIdle: {
 		ChannelStateJoining,
 		ChannelStateAwaitingInvite,
+		ChannelStateMonitoring,
 		ChannelStateError,
 		ChannelStateKicked,
 		ChannelStateParted,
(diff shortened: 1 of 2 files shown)


s0up4200:
Field tested against Norbits (InspIRCd) on a live instance with the `pr-2587` image. The server force-join now confirms the channel: the state machine accepts `Idle -> Monitoring`, "monitoring since" is set, and the "join not confirmed" error loop is gone.

zze0s (approved):

PR #2594, opened 2026-08-12
fix(notifications): respect Discord embed limits

com6056:
# Description

Discord answers `400 {"embeds": ["0"]}` and drops the entire message when an embed field value goes over 1024 characters, and `buildEmbed` never capped anything.

It bites on PUSH_ERROR. A failed *arr push wraps the whole `http.Request` into its error string, that lands in the `Reasons` field, and the notification never arrives. So the one event you actually want to hear about is the one that reliably fails to send. It is not an edge case either: across 440 real PUSH_ERROR rows in my instance the rejection string has a median length of 1563 and a max of 1786, and 439 of the 440 are over the limit.

Checked against the real API by posting one of those 1787 character rejections through both versions of `buildEmbed`:

```
uncapped (develop)     longest field = 1795 runes -> HTTP 400 {"embeds": ["0"]}
capped (this PR)       longest field = 1024 runes -> HTTP 204
```

Long release names can do the same to the 256 character title limit, so title and description are capped too.

Truncation drops the middle rather than the tail. `errors.Wrap` prepends, so the cause of a failed push sits at the very end and tail truncation would throw away the only part worth reading. A useful side effect is that the middle is where the request dump keeps `X-Api-Key`, so the truncated field stops carrying it. The tracker passkey sits nearer the front and does survive, which is a separate bug at the source that I have opened a second PR for.

Only the four per-field limits are handled. The 6000 character combined limit and the 25 field cap are not: `buildEmbed` emits at most 10 fields, and I could not build a payload from any current call site that reaches 6000, so I would rather leave it than guess.

## Type of change

## How has this been tested?

* `go test ./...` passes. New tests cover the helper, multi-byte input, a PUSH_ERROR shaped payload asserting every limit holds and the code fence survives, and an over-long Subject/Message pair.
* Mutation checked rather than assumed. Removing the cap fails the suite, and swapping middle truncation for tail truncation fails the assertion that the cause survives.
* The live before and after against a real Discord webhook, shown above.

## Checklist

## AI disclosure

Yes. Written with Claude Code (Claude Opus 5). The implementation and the tests are AI-written and I reviewed them, and the verification above is mine: the 440 row measurement and the live webhook check both come from my own instance.

### DIFF
--- a/internal/notification/discord.go
+++ b/internal/notification/discord.go
@@ -47,6 +47,48 @@ const (
 	GRAY       EmbedColors = 10070709 // 99aab5
 )
 
+// Per-field limits. Exceeding any of them makes Discord reject the whole
+// message with 400 rather than truncating it.
+// https://docs.discord.com/developers/resources/message#embed-object-embed-limits
+const (
+	discordEmbedTitleLimit       = 256
+	discordEmbedDescriptionLimit = 4096
+	discordEmbedFieldNameLimit   = 256
+	discordEmbedFieldValueLimit  = 1024
+)
+
+const discordTruncationMarker = " [...] "
+
+// truncateForDiscord shortens s to at most limit characters by dropping the
+// middle. A failed *arr push wraps its cause last, so keeping only the head
+// would discard the one part worth reading. Discord counts characters rather
+// than bytes.
+func truncateForDiscord(s string, limit int) string {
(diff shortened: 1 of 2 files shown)


zze0s (approved):
Thanks for the PR! Great improvement!

PR #2595, opened 2026-08-12
fix(downloadclients): redact secrets from arr errors

com6056:
# Description

A failed *arr push builds an error string that carries three credentials.

The clients wrap transport errors with the whole `http.Request` via `%+v`. By that point `setHeaders` has run, so the rendered struct contains `X-Api-Key`, and `req.SetBasicAuth` has run too, so on an instance behind basic auth it also contains `Authorization: Basic …`. On a push the body is a `*bytes.Buffer`, which fmt renders through its `Stringer`, so the marshalled JSON goes in as well. The action layer then wraps that with the whole `ReleasePushRequest`, which carries `DownloadUrl` and therefore the tracker passkey.

That string does not just go to the log. It is attached to the release as a rejection, stored in `release_action_status.rejections`, rendered in the web UI, and handed to every configured notification provider. I found it because a Discord notification arrived with my tracker passkey in it.

Measured on a real transport failure, with placeholder secrets:

```
develop                 ~1.5 KB   apikey: yes   passkey: yes   basic auth: yes
clients fixed only         337    apikey: no    passkey: yes   basic auth: no
both fixed (this PR)       226    apikey: no    passkey: no    basic auth: no
```

The clients now drop the request from the wrap entirely rather than interpolating a URL. `http.Client.Do` already returns a `*url.Error` whose message carries the method and the URL with any password redacted by `stripPassword`, so adding `reqUrl` was both redundant and strictly less safe than the error it was prefixing: `u.String()` renders userinfo verbatim. That also matches what `lidarr` already did. Where a URL is genuinely useful, on `NewRequest` failures that produce no `*url.Error`, it now goes through `u.Redacted()`. The action layer wraps with `r.Title`.

Two things left alone on purpose. `errors.Wrap(err, "could not marshal data: %+v", data)` in the same files would render the same struct, but `json.Marshal` cannot fail for these all-scalar types and the only thing it could usefully print is the value itself. And the same `%+v`-on-request pattern is still live in `pkg/torznab`, `pkg/newznab`, `pkg/red` and `pkg/ops`. Those need a different fix, since torznab and newznab put the key in the query string where swapping the format verb does not help, so they belong in their own PR rather than being bolted on here.

Related: these strings are also usually over Discord's 1024 character embed limit, which is #2594. The two are independent and this one is the root cause.

## Type of change

## How has this been tested?

* New unit test in `pkg/arr/radarr` drives all three request paths (`get`, `getJSON`, `postBody`) against a closed listener, with a distinct placeholder for the API key, the passkey and the basic-auth password, so a failure says which one leaked.
* Mutation checked rather than assumed. Reintroducing `%+v", req` fails 4 subtests, and reintroducing `%v", reqUrl` fails 4 subtests via the basic-auth case.
* `go test ./...` passes.

## Checklist

## AI disclosure

Yes. Written with Claude Code (Claude Opus 5). The change and the tests are AI-written and I reviewed them. The leak itself I hit on my own instance, and the measurements come from there.


zze0s (approved):
Thanks for the PR!

I made some minor changes on the same theme in the same files.

This is only a partial fix so I'll open a PR that sanitizes before sending notifications to hopefully capture as much as possible.

PR #2607, opened 2026-08-14
feat(clients): add Chaptarr support

zze0s:
# Description

Adds [Chaptarr](https://github.com/Chaptarr/chaptarr) as a download client, filter action, and list source.

Chaptarr is a Readarr fork for audiobook and eBook libraries. Its API is `/api/v1` and `release/push` is inherited from Readarr unchanged, so this is a new client type rather than new plumbing.

* New `pkg/arr/chaptarr` client: `Test`, `Push`, `GetBooks`, `GetTags`
* New `CHAPTARR` type across `DownloadClientType`, `ActionType`, and `ListType`
* Action pushes to `POST /api/v1/release/push`; list source pulls `GET /api/v1/book`
* Frontend registration, plus i18n for all 8 locales

Why it is not just the Readarr client pointed at port 8789:

* `genres` has shipped both as a JSON string and as an array, which fails the whole `GET /book` decode - this is the exact error reported in the discussion. 0.9.911 sends an array, but `Book` omits the field entirely so neither shape can break the decode, and `Test_client_GetBooks` pins both.
* Chaptarr returns a reduced validation-error body. `ChaptarrErrorPipeline.cs` projects failures to `{propertyName, errorMessage}`, where Readarr serializes the full `ValidationFailure`. Reusing Readarr's 5-field `BadRequestResponse` renders every rejection as `[: ] Title: Unable to parse - got value:`.
* `appName` is `"Chaptarr"`, and the version is a 4-part `0.9.925.0`.
* Chaptarr keeps one `/book` row per media type, so a book held as both an audiobook and an eBook appears twice. `processChaptarr` dedupes on the generated title.

Related discussion: https://github.com/autobrr/autobrr/discussions/2413

## Type of change

## How has this been tested?

* `go test $(go list ./... | grep -v test/integration)` passes
* `go test -tags=integration ./pkg/arr/chaptarr/...` passes - covers push accepted, push rejected, the 400 validation body, `system/status`, bad credentials, and `GetBooks` with `genres` as an array, as a bare string, and empty
* `go build ./...`, `go vet`, and `go fmt` clean
* `pnpm --dir web lint` reports no errors on the changed files; `pnpm --dir web build` succeeds

Tested against a live `chaptarr/chaptarr:latest` container (0.9.911.0) with a real 497-book library. Every fixture under `pkg/arr/chaptarr/testdata/` is a verbatim capture from that instance, not hand-written.

| Path | Result |
| --- | --- |
| `Test` | `appName="Chaptarr"` `version="0.9.911.0"` |
| `Test` with a bad key | `unauthorized: bad credentials` (HTTP 401) |
| `Push` accepted | `approved=true`, no rejections |
| `Push` rejected | `["Unknown Author"]` |
| `Push` failing validation | HTTP 400 `[{"propertyName":"Title","errorMessage":"'Title' must not be empty."}]` -> `["Title: 'Title' must not be empty."]` |
| `GetBooks` | 497 books decoded |
| `GetBooks("audiobook")` | 287 books, so the `mediaType` filter works |
| `GetTags` | `[]` |

The duplicate-title problem is real at scale: those 497 rows are 287 audiobook + 210 ebook, and collapse to **281 unique titles**. Without the dedupe the list would write 216 redundant entries into `match_releases`.

Two things worth knowing for support:

* Chaptarr logs `[Warn] BookController: [API-PERFORMANCE] Taking general book path without an author/book filter.` on every unfiltered `GET /book`, which is what the list refresh does.
* `author` is never populated on `GET /book` - it is gated behind an `include` param - so `Book.Author` is always nil on this path.

No version gate: `release/push` was inherited from Readarr at fork time, so there is no minimum version to floor against, and the Readarr client has no gate either.

No tags support: Chaptarr's `BookResource` has no tags field - they live on `AuthorResource`, split into `tags`/`audiobookTags`/`ebookTags`. Chaptarr is therefore absent from `GetArrTags` and from the tag multiselect gate in `ListForms.tsx`, matching Readarr and Lidarr.

Chaptarr appears in the download client, action, and list type pickers. No new UI, only new entries in existing ones.

## Checklist

## AI disclosure

Yes. Written with Claude Code (Opus 5), driven and reviewed by me.

The model researched Chaptarr's API by reading its C# source to confirm `release/push` still exists and to find where it diverges from Readarr, then wrote the client, the wiring, and the tests, and finally validated all of it against a live container. Every divergence claimed above was verified against the upstream source and then confirmed on the wire, rather than assumed from Readarr parity. I reviewed the full diff.


zze0s:
Chaptarr is too immature and changes too much (aka proper slop) for it to be a stable integration at this time. Might revisit in the future.

PR #2612, opened 2026-08-15
chore(indexers): update Samaritano IRC port and TLS

vaaleyard:
# Description

Samaritano moved its announce server to TLS on port 6697.

* Definition: `port: 6667` -> `6697`, `tls: false` -> `true`.
* Definition: removed the `auth.account` / `auth.password` settings - NickServ auth is not needed to join the server, only the channel.
* Definition: added a parse test for announces without `tmdb`.
* Migrations: `sqlite/96_irc_update_samaritano_port_and_tls.sql` and `postgres/86_irc_update_samaritano_port_and_tls.sql` move existing `irc.samaritano.cc` networks to 6697 with TLS, so users do not have to edit the network by hand.
  * No-op for rows already on 6697 with TLS enabled.
  * Rows that would collide with an existing 6697 row for the same nick are skipped - `irc_network` has `UNIQUE (server, port, nick)`, and an unguarded update would abort the migration at startup for anyone who already added the TLS network manually.
* Existing `auth_account` / `auth_password` values are left untouched; the migration only changes `port` and `tls`.

## Type of change

## How has this been tested?

* Used the definition as a custom definition in a local autobrr instance.
* `go test ./internal/indexer/...` passes, covering the announce parse tests.
* `go test -tags=integration ./internal/database/migrations/` passes for both drivers, with new cases asserting the stale row is upgraded, a colliding row is left alone, and unrelated networks are untouched.

## Checklist

## AI disclosure

Was any AI used in this PR? Roughly how much of the code was AI-written, and what model/tool?
- Yes to check for any possible issues with the changes


vaaleyard:
The tests failed because there was one release without tmdb, should I add an empty expect for this field?

zze0s:
Hey! yeah it needs the empty tmdb there.

zze0s (approved):
I added db migrations so it auto migrates the irc network on startup if not already done manually 👍

vaaleyard:
> I added db migrations so it auto migrates the irc network on startup if not already done manually 👍

Thank you very much

PR #2635, opened 2026-08-22
feat(actions): support rejection via exec script output

wnielson:
# Description

The `Exec` action currently supports only two outcomes: approved (exit 0) or error (non-zero exit). This adds a third outcome — **rejected** — without breaking existing scripts.

Scripts can signal rejection by printing a trailing block of one or more lines prefixed with `REJECT:` to stdout/stderr and exiting 0. Detection walks backward from the end of the output and stops at the first non-`REJECT:` line, so a script can log freely and only its final lines control the rejection decision; a `REJECT:` line that isn't part of that trailing block is ignored. Each captured line (minus the prefix and surrounding whitespace) becomes a rejection reason, matching the `[]string` pattern used by other action types (arr clients, qBittorrent, etc.).

Non-zero exit codes remain errors, and scripts with no trailing `REJECT:` lines continue to be treated as approved. Existing scripts require no modification.

Example script:
```bash
#!/bin/bash
if some_condition; then
    echo "REJECT: release does not meet criteria"
    exit 0
fi
# approved — exit 0 with no REJECT: lines
```

**Why this format:** the goal was a rejection signal that's trivial for any script to emit and impossible to get wrong, without forcing script authors to adopt a structured output format. A convention that required scripts to emit JSON (or any other structured payload) on stdout would break every existing Exec action script overnight and raise the bar for writing a new one — you'd need a JSON encoder just to say "no thanks" from a five-line shell script. Plain `REJECT: <reason>` lines mean:

- **Backwards compatible** - scripts that don't know about this convention keep working exactly as before; their output is simply ignored and the release is approved, same as today.
- **Trivial to adopt** - `echo "REJECT: too old"` (or the equivalent in any language) is the whole integration. No schema, no encoding/escaping rules, no library.
- **Easy to reason about** - a human reading a script's output can immediately tell whether/why it rejected, without decoding anything.

> **Note:** The Exec action docs at autobrr.com should be updated to document the `REJECT:` prefix convention.

## Type of change

## How has this been tested?

Added/updated unit tests in `internal/action/exec_test.go` covering:

* no output (approved)
* a single trailing `REJECT:` line
* multiple trailing `REJECT:` lines
* a `REJECT:` line preceded by unrelated log output (still detected)
* a `REJECT:` line followed by non-`REJECT:` output (ignored, not trailing)
* non-`REJECT:`-prefixed output (approved)
* non-zero exit code (error)

`go test ./internal/action/...` passes.

## Checklist

## AI disclosure

Yes. This PR was developed with the help of Claude (Sonnet 5, Anthropic) via Claude Code. The implementation, tests, and this description were AI-assisted, but every design decision (the `REJECT:` convention itself, the trailing-block-only detection semantics, and the rationale for avoiding a structured/JSON output format) was discussed, iterated on, and reviewed by me. The resulting code and tests were read and verified line-by-line before pushing, not accepted as-is.

### DIFF
--- a/internal/action/exec.go
+++ b/internal/action/exec.go
@@ -4,8 +4,11 @@
 package action
 
 import (
+	"bufio"
+	"bytes"
 	"context"
 	"os/exec"
+	"strings"
 	"time"
 
 	"github.com/autobrr/autobrr/internal/domain"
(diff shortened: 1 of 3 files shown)


zze0s:
Hey! Thanks for the PR. In most cases you should use external filter if it needs to stop processing on reject.

Wouldn't it be better for this to trigger on some specific exit code instead of a text string of the script output?

wnielson:
I chose this route because I didn't want to break anyone's existing script workflow, but the cleaner solution to me is the return code route where 0=accepted, 1=rejected and >1 is error.

For my workflow a custom filter won't work because my exec script acts as a proxy to Sonarr--in some instances Sonarr will reject a release that I actually want and my script handles this extra logic (which can't be solved via Sonarr custom formats).

zze0s:
I see :+1: 

I started working on action improvements in this PR https://github.com/autobrr/autobrr/pull/2494 to have similar options as external filters and that includes expected status codes handling which makes this a dupe.

That PR isn't ready just yet but I'll close this one and make sure the other implements everything.

wnielson:
Sounds good and thanks for the great project!

PR #2640, opened 2026-08-27
feat: add *arr types to external filters

tYiZQMr93eYPp:
# Description

Allows `autobrr` users to use *arr type filters on external filters.
The context and motivation behind this is the following:
* Some indexers don't allow automations, but they announce through, for example, Telegram.
* I have an automation called [`telegramircrr`](https://github.com/tYiZQMr93eYPp/telegramircrr) that parses those announces and announces them on an IRC server.
* As some indexers don't allow automations, I want to still be notified of releases they publish. But, I want them to go through an *arr first to decide what to do with it.
* If not rejected, notify me through, for example, Apprise. And, optionally, execute additional actions.

The issue I was having is that all `Actions` execute no matter what. But that's not what I want.
I want a strict and ordered execution of steps, and that can be achieved with the `External` filters.

I hope this is clear. Otherwise, I could clarify.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] New indexer definition
- [ ] Indexer deprecation
- [ ] Indexer definition update
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor / maintenance (no functional change)
- [X] This change requires a documentation update

## How has this been tested?

* I'm, currently, testing it locally. So far, it's being doing great.
* Tests passed.
* The flow and setup look like this: Telegram -> telegramircrr -> IRC -> autobrr -> Filter -> External filter -> Action

external-filters-arrs

## Checklist

- [X] My PR title follows the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) format (it becomes the squashed commit message)
- [X] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [X] I ran `go fmt` and the test suite passes locally
- [X] If this changes the database schema, I have added migrations for both SQLite and PostgreSQL

## AI disclosure

* Q: Was any AI used in this PR? Roughly how much of the code was AI-written, and what model/tool?
  * A: Yes. 100% of the code was AI generated. The model was Claude Sonnet 5 (High thinking effort, 200k ctx)

## Additional comments

I'm no Go/Python developer. Excuse me for any mistakes.

I'm not sure if the Dockerfile changes are necessary or not. The agent said they were, otherwise it wouldn't build. I didn't want to dedicate much time to it.

Of course, feel free to squash the changes, request edits, or make them yourselves.  
I needed this feature and thought it may be useful for the rest of the people using your amazing project.

Thank you very much for your time and dedication!


zze0s:
Hey, thanks for the PR!

> As some indexers don't allow automations, I want to still be notified of releases they publish. But, I want them to go through an *arr first to decide what to do with it.
  If not rejected, notify me through, for example, Apprise. And, optionally, execute additional actions.
> The issue I was having is that all Actions execute no matter what. But that's not what I want.
I want a strict and ordered execution of steps, and that can be achieved with the External filters.

I do not understand why the regular flow doesn't work? Filter actions with arr clients already have the approve/reject/error states, so they are already handled like "regular filters" and will continue to the next filter if a release is rejected.

Running multiple arr actions per filter? That's generally not a good idea unless it's a HD and 4K sonarr or radarr clients.

The existing actions are very basic tho so I'm improving them here https://github.com/autobrr/autobrr/pull/2494 to behave a bit more like external filters with reject handling etc.

tYiZQMr93eYPp:
> I do not understand why the regular flow doesn't work? Filter actions with arr clients already have the approve/reject/error states, so they are already handled like "regular filters" and will continue to the next filter if a release is rejected.

@zze0s, You're right. I forgot to mention that I have custom notifications set up through a webhook instead of the using the filter/global notifications (just for these types of cases).

**Why? You may be asking**  
My notifications are set up in the `Actions` tab, and execute a webhook towards Apprise.  
Because those indexers don't allow automations (but do post the release's `infoURL`), I want to be able to easily go to the website just by clicking on the link of the custom notification body.  
The default notifications aren't enough for me for the following reasons:
1. They don't send the release's metadata (unless I've missed something).  
2. And, even if a release is rejected in the defined actions, it will keep executing the rest of the defined action filters, meaning that I will get notified even if the release got rejected by the arrs.

I hope to have clarified the problem I was facing.

---

> Running multiple arr actions per filter? That's generally not a good idea unless it's a HD and 4K sonarr or radarr clients.

I don't, I also think it's a bad idea.  
It's up to the user to do it that way or not, they will, probably, run into issues later on or make unnecessary requests to the set up actions.

---

> The existing actions are very basic tho so I'm improving them here #2494 to behave a bit more like external filters with reject handling etc.

Looks good! Would be happy to test that out and provide feedback.  
In any case, feel free to reuse whatever's in here.  
In the meantime, I'll keep testing this solution. It seems to be enough for me (for now...).

---

I'm down for suggestions.
#2394's idea LGTM.  
Maybe I should ask the model to implement ordering, rejections, and the rest in the `Actions` tab?
Or, maybe, that would cause confusion and a big mess. Not sure, I haven't explored the codebase much nor I know how to write proper Go code.

zze0s:
Ok that's some good extra context.

You can do custom webhook notifications instead of the built in ones, and in develop there's a some major bug fixes for filter notifications/global/filter muted etc that will help with it.

The external filters with script and webhook support should be enough to create custom scripts etc so we don't have to extend them with more specific features.

That other PR isn't done yet so it will get more features and I think it can be enough for your usecase so you can conditionally trigger a webhook action based on the previous action outcome.

Otherwise if I were you I'd build a tiny shim that sits between autobrr and the arr that have the same api for release/push that autobrr uses and you can intercept/check the return status in that and push your notification in that middleware instead.

I'm closing this PR in favour of #2494 and will keep your workflow in mind!

PR #2680, opened 2026-09-10
fix(dashboard): bucket release stats in UTC so charts show correct dates/times

### LINKED ISSUE #2673: Dashboard graphs showing wrong date/time
**Version**
v1.85.0

**Describe the bug**
I'm running in docker, with the environment variable `TZ=America/New_York`. The dashboard graphs seem to be all over the place in terms of wrong date showing for Activity/Download volume, and wrong hourly heatmap for Activity by hour.

I reset all of my stats earlier this morning (Sep 6), but the Activity and Download volume graphs only show Sep 5 and don't even display Sep 6 as of 14:00 local time here. Not sure how that happens even with a UTC vs UTC-4 error as recent activity from 1PM today should be Sep 6 regardless of errors in timezone settings.

The Activity by hour heatmap is also offset by several hours, but not in the direction I would expect if it was displaying UTC instead of my local timezone (UTC-4). For example the most recent activity was from 13:00-14:00 local (17:00-18:00 UTC) but the heatmap is showing that data at 09:00.

Tested in firefox/edge on desktop, and chrome on android and behavior is same in all browsers.

**To Reproduce**
1. Launch docker container with `TZ=America/New_York` (I assume other timezones will also have bugs)
2. Reset all statistics history
3. Wait for some activity to happen
4. Check graph display

**Expected behavior**
Dashboard graphs should show data mapped to correct system timezone, or at least indicate everything in UTC.

**Screenshots**
Screenshot taken at 14:30 UTC-4, Sep 6. System stats were reset morning of Sep 6, so no data exists for Sep 5

Image

 - OS: Windows 11
 - Browser: Firefox 155.0.1, Edge 152.0.4191.62

thejdubb02:
## Description

Fixes #2673.

The dashboard "Download volume" and "Activity by hour" charts showed the wrong dates and times. The root cause is a mismatch in how release timestamps are bucketed for stats:

- On Postgres the release.timestamp column is timestamptz (migration 44), so to_char/EXTRACT bucket in UTC.
- On SQLite the stats used substr() on the raw stored string, which keeps the local UTC offset written by time.Now().Format(time.RFC3339), so buckets were in the server's local time.

The charts assume the API grid is UTC and convert it to the viewer's local time. Because SQLite delivered a local-time grid, the heatmap hours were shifted a second time and the day labels rolled back a day for viewers west of UTC.

## Changes

- internal/database/release.go: statsDayExpr() and statsHourExpr() now normalize to UTC on both drivers. SQLite uses strftime('%Y-%m-%d' / '%H', timestamp), which normalizes the RFC3339-with-offset form and treats the legacy space-separated (CURRENT_TIMESTAMP, UTC) form as UTC. Postgres uses "... AT TIME ZONE 'UTC'" so bucketing is deterministic regardless of the session TimeZone. The API grid is now UTC on both databases.
- web/src/screens/dashboard/VolumeChart.tsx and ActivityChart.tsx: day keys are parsed as local midnight and formatted verbatim, so a UTC day bucket is labeled with that date in every timezone, instead of being parsed as UTC and then formatted in local time (which rolled the label back a day). HourHeatmap already converts the UTC grid to local hours, so it is correct now and is unchanged.
- Adds TestReleaseRepo_StatsBucketUTC, which stores a release with a non-UTC offset and asserts the stats bucket by UTC hour and day. It fails if the bucketing is reverted to local time.

## Testing

- go build ./... and go test ./internal/database/ pass on both SQLite and Postgres, run under TZ=America/New_York.
- The new test fails when the expressions are reverted to the old local-time bucketing.
- pnpm exec tsc --noEmit, pnpm lint, and pnpm test (52 tests) all pass. The volume-label test fails on the pre-fix code under a non-UTC timezone.

## Notes and tradeoffs

- strftime parses each row where substr did not. The stats queries filter by the selected range (WHERE timestamp >= cutoff, on the timestamp index), so this only touches the rows in range; the "all" range scans the table as it did before. Correctness needs the parse. Storing timestamps in UTC would let a later change restore the substring fast path.
- Day buckets are UTC days labeled verbatim, while the heatmap shows viewer-local hours. Showing day buckets in the viewer's own day would need the viewer timezone passed to the stats API, which is a larger change and is left out of scope here.

## AI disclosure

Most of this change was written with AI assistance (Claude, via Claude Code). It was human-reviewed and verified before submission: the SQLite and Postgres bucketing behavior was checked directly, the build and both-database tests were run under a non-UTC timezone, and the new test was confirmed to fail without the fix.

### DIFF
--- a/internal/database/release.go
+++ b/internal/database/release.go
@@ -1038,19 +1038,19 @@ func minDateKey[T any](m map[string]*T) string {
 
 func (repo *ReleaseRepo) statsDayExpr() string {
 	if repo.db.Driver == "sqlite" {
-		// substr instead of strftime: both timestamp formats found in old
-		// databases share the YYYY-MM-DD HH prefix, and skipping per-row date
-		// parsing roughly halves the scan cost on large tables
-		return "substr(timestamp, 1, 10)"
+		// strftime normalizes the stored offset to UTC for both the RFC3339 and the
+		// legacy space-separated forms, so buckets match Postgres (timestamptz).
+		// substr returned the raw local hour/day, which broke the dashboard charts.
+		return "strftime('%Y-%m-%d', timestamp)"
 	}
-	return "to_char(timestamp, 'YYYY-MM-DD')"
+	return "to_char(timestamp AT TIME ZONE 'UTC', 'YYYY-MM-DD')"
 }
 
 func (repo *ReleaseRepo) statsHourExpr() string {
 	if repo.db.Driver == "sqlite" {
-		return "CAST(substr(timestamp, 12, 2) AS INTEGER)"
+		return "CAST(strftime('%H', timestamp) AS INTEGER)"
 	}
-	return "EXTRACT(HOUR FROM timestamp)::int"
+	return "EXTRACT(HOUR FROM timestamp AT TIME ZONE 'UTC')::int"
 }
 
 func (repo *ReleaseRepo) statsQueryRows(ctx context.Context, qb sq.SelectBuilder, scan func(rows *sql.Rows) error) error {
--- a/internal/database/release_test.go
+++ b/internal/database/release_test.go
@@ -767,6 +767,66 @@ func TestReleaseRepo_StatsDashboard(t *testing.T) {
 	}
 }
 
+func TestReleaseRepo_StatsBucketUTC(t *testing.T) {
+	ctx := t.Context()
+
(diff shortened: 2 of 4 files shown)


zze0s:
Hey! Thanks for the PR! I will read some extensive tests and see if this is the correct fix :+1:

PR #2684, opened 2026-09-13
feat(indexers): enrich Hebits IRC releases from torrent API

Ddorda:
## Description

Hebits IRC announces do not include freeleech status, so a freeleech-only filter always rejected those releases. This adds a Hebits Gazelle API client that fetches torrent details (`freeTorrent`, seeders, leechers, size) with an optional session cookie and applies them **before** filter evaluation.

The cookie setting is optional and stored as an existing secret indexer setting. API failures fail open (log a warning, leave the release un-enriched), so a freeleech filter still rejects unknown values.

Also copies `pnpm-workspace.yaml` into the Docker web build so `pnpm install --frozen-lockfile` sees the same overrides as the lockfile.

## Type of change

## How has this been tested?

- `go test ./pkg/hebits/... ./internal/filter/...`
- Full Go test suite excluding integration tests
- `pnpm --dir web lint`
- Live Hebits API lookup against torrent `134205` using a local gitignored cookie file (cookie not committed)
- Docker image build after the `pnpm-workspace.yaml` copy fix

## Checklist

## AI disclosure

Yes. Most of this change was written with Cursor (Grok 4.6). I reviewed the hook point, tests, and secret handling. The live cookie used for verification was kept in a gitignored local file and was not committed.


.gitignore
```diff
@@ -51,6 +51,10 @@ package-lock.json
 # Docker
 config/
 
+# Local secrets — never commit tracker session cookies
+.hebits.cookie
+**/.hebits.cookie
+
```
zze0s, line 57: Revert change. This is specific to your setup.

.dockerignore
```diff
@@ -1,3 +1,5 @@
+.hebits.cookie
+
```
zze0s, line 2: Revert change. This is specific to your setup.

internal/domain/indexer.go
```diff
@@ -759,12 +759,16 @@ func (p *IndexerIRCV2Parse) Parse(def *IndexerDefinition, channelName string, va
...
-	Uploader    string `json:"Uploader"`
-	RecordLabel string `json:"RecordLabel"`
+	Id               string `json:"Id"`
+	TorrentId        string `json:"TorrentId,omitempty"`
+	InfoHash         string `json:"InfoHash"`
+	Size             string `json:"Size"`
+	Uploader         string `json:"Uploader"`
+	RecordLabel      string `json:"RecordLabel"`
+	Freeleech        bool   `json:"Freeleech,omitempty"`
+	FreeleechPercent int    `json:"FreeleechPercent,omitempty"`
```
zze0s, line 769: Remove `FreeleechPercent` here, unless they actually have it but doesn't look like that based on the usage in the api client.

Better to have the caller set all correct values if it comes back as freeleech.

internal/filter/service.go
```diff
@@ -721,6 +723,44 @@ func (s *Service) CheckFilter(ctx context.Context, f *domain.Filter, release *do
...
+	release.Seeders = torrentInfo.Seeders
+	release.Leechers = torrentInfo.Leechers
+
+	if release.Size == 0 {
+		if size := torrentInfo.ReleaseSizeBytes(); size > 0 {
+			release.Size = size
+		}
+	}
+}
+
```
zze0s, line 763: Not a fan of this implementation at a first glance.

zze0s (commented):
Hi! Do they really not have API keys and require Cookies?? If it's gazelle based it should support API keys.

Before this is merged I would suggest you try to influence them to add freeleech to the announce instead. They can add it to the first announce line at the end without breaking parsing for autobrr etc.

`^New: .* Category: (?P<category>.*) Size: (?P<torrentSize>.*) Seeders: .+ Leechers: .+`

Like:

`^New: .* Category: (?P<category>.*) Size: (?P<torrentSize>.*) Seeders: .+ Leechers: .+ Freeleech: \d`

Or optional value etc.

We don't really want more indexer api clients to maintain.

internal/domain/indexer.go
Ddorda: They actually have that!

Ddorda:
> Hi! Do they really not have API keys and require Cookies?? If it's gazelle based it should support API keys.
> 
> Before this is merged I would suggest you try to influence them to add freeleech to the announce instead. They can add it to the first announce line at the end without breaking parsing for autobrr etc.
> 
> `^New: .* Category: (?P<category>.*) Size: (?P<torrentSize>.*) Seeders: .+ Leechers: .+`
> 
> Like:
> 
> `^New: .* Category: (?P<category>.*) Size: (?P<torrentSize>.*) Seeders: .+ Leechers: .+ Freeleech: \d`
> 
> Or optional value etc.
> 
> We don't really want more indexer api clients to maintain.

Hey @zze0s! really appreciate you taking the time to review this.

Regarding the Hebits API - they don't expose api keys, and they heavily changed the original behavior of Gazelle.

Regarding asking them to change the way they announce the torrents, there's really zero chance for them to take this feeback, that's why I made this patch in the first place 😂

I really thought this could be useful for other use cases and users - The cookies login is something i've already seen on other *rr projects, and the api enrichement could fill the gaps when IRC message is not enough, just like in my case.

